PR #104, opened 2022-05-23
Fix non-square profile pictures being squished

sambartik:
#### Description
Currently the aspect ratio of profile pictures in dropdown is not being preserved, resulting in pictures looking "squished". Adding `object-fit: cover;` to the image element fixes this ([caniuse.](https://caniuse.com/object-fit)). Please, help me find me another components where there is a profile picture so I can change it there too.

Before:
Before

After:
After

#### To-Dos

- [x] Find all other places where the profile picture is displayed and fix it there too

#### Issues Fixed or Closed

No prior issue

### DIFF
--- a/src/components/IssueList/IssueItem/index.tsx
+++ b/src/components/IssueList/IssueItem/index.tsx
@@ -229,6 +229,7 @@ const IssueItem: React.FC<IssueItemProps> = ({ issue }) => {
                             src={issue.createdBy.avatar}
                             alt=""
                             className="avatar-sm ml-1.5"
+                            style={{ objectFit: 'cover' }}
                           />
                           <span className="truncate text-sm font-semibold group-hover:text-white group-hover:underline">
                             {issue.createdBy.displayName}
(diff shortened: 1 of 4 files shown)


Fallenbagel:
Oh I didnt notice this. Thanks for the pr! From as far as I remember (haven't gotten the time to check) I think you covered everywhere that references a profile picture 🤔

sambartik:
We could merge this and if any more instances come up, fix it up in next PR.

Fallenbagel:
> We could merge this and if any more instances come up, fix it up in next PR.

Btw I just checked on mine and I'm not sure why mine isn't squashed. I also have a square profile picture

sambartik:
The squishing happens only if the image is **not** a square. The effect is best visible with long and narrow images. This is the picture I used in demo:

long-narrow

Fallenbagel:
OH. Okay thanks for the clarification!

Fallenbagel:
> @Fallenbagel We should probably reverse this PR to make merging changes from the upstream repo sct/overseerr#2782 easier. Sorry for the mess. Should I open a new PR to do so?

Yes let's do that

PR #105, opened 2022-05-24
Fixed other squished avatars missed in #104

sambartik:
#104

#### To-Dos


Fallenbagel:
Hi could you pr the whole avatar fix thing to Overseerr?

sambartik:
I agree, this should be the way.

Fallenbagel:
> I agree, this should be the way.

Yes. When they accept, we can implement the rest of the fixes

PR #116, opened 2022-05-28
feat: Add Job to Remove old Requests

CyferShepard:
This Job Will run every minute to check for Requests marked as available that are older than 48 hours and then automatically remove them. This is kind of similar to the feature already present on Ombi.
 
#### To-Dos

### DIFF
--- a/server/job/schedule.ts
+++ b/server/job/schedule.ts
@@ -1,6 +1,7 @@
 import schedule from 'node-schedule';
 import { MediaServerType } from '../constants/server';
 import downloadTracker from '../lib/downloadtracker';
+import requestCleanup from '../lib/requestCleanup';
 import { plexFullScanner, plexRecentScanner } from '../lib/scanners/plex';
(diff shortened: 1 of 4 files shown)


Fallenbagel:
Hi. Could you PR this feature to Overseerr? As this is a core Overseerr feature, we should see what they think of this PR and then we add this feature to Jellyseerr (hopefully they will accept it as it is quite a requested feature). This is done to ensure to keep feature parity with them.

CyferShepard:
HI, sure, I added it here since i dont use overseerr but will look into this.

Fallenbagel:
> HI, sure, I added it here since i dont use overseerr but will look into this.

Neither do I, but since we are keeping feature parity with them, it would be easier to pull in any commits from upstream with no merging conflicts

Fallenbagel:
@CyferShepard Also could you ping me or sth here or on discord once you open the PR as I want to follow it closely

PR #121, opened 2022-06-02
Hide Overseerr settings when running in Jellyfin/Emby mode

notfakie:
#### Description
This will make things more clear until we support running multiple modes at the same time. 

#### To-Dos


Fallenbagel:
Nice. Love this

PR #126, opened 2022-06-05
 fix(jellyfin): ignore additional items with virtual location type

sambartik:
I have noticed that episodes with virtual location type break things as well, so I have added the filter to all API methods returning JellyfinLibraryItem to not worry in the future.


Fallenbagel:
Sorry I accidentally closed without merging

PR #137, opened 2022-06-11
prevent Email Update on Jellyfin Import

CyferShepard:
#### Description
With the implementation of the new Email Field, if an admin runs the Jellyfin Import and selects all users, the import overwrites the existing email fields for the users. This change removes the code to update the email field on any existing users. New users will still have the email field set to account.Name

#### To-Dos


Fallenbagel:
Hi. The import button has been overhauled by @NicolaiVdS so this PR is no longer needed. It will be added in the already open PR with some other fixes like when users are added, they are no longer shown

PR #217, opened 2022-09-07
fix(deps): do not list email-validator as a devDependency

sambartik:
Because it was listed as a devDependency it got removed during a container build process causing module not found error.

#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX


Fallenbagel:
Could you fix the lint issue?

PR #240, opened 2022-10-13
Fix jellyfin external url basepath being ignored 

### LINKED ISSUE #237: Jellyfin user import show no profile pictures
### Description

Jellyfin users import.

No profiles pictures are imported, they are shown as broken link

Thanks

### Version

1.2.0

### Steps to Reproduce

No need to explain

### Screenshots

image

### Logs

_No response_

### Platform

desktop

### Device

Windows 11

### Operating System

Linux

### Browser

Opera GX

### Additional Context

_No response_

### Code of Conduct

sambartik:
#### Description
The previous PR introduced a new syntax which ignores external jellyfin base path's. I. e. http://localhost:4124/jellyfin.

#### To-Dos

#### Issues Fixed or Closed

- Fixes #237

### DIFF
--- a/server/entity/Media.ts
+++ b/server/entity/Media.ts
@@ -200,10 +200,11 @@ class Media {
       const pageName =
         process.env.JELLYFIN_TYPE === 'emby' ? 'item' : 'details';
       const { serverId, hostname, externalHostname } = getSettings().jellyfin;
-      const jellyfinHost =
+      let jellyfinHost =
         externalHostname && externalHostname.length > 0
           ? externalHostname
           : hostname;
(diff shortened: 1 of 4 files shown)


sambartik:
A potential fix, probably having the trailing slashed removed just once after saving the settings would be the best solution so the code is kept DRY.

sambartik:
Note: The URL class which was used strips any paths from the base.

Fallenbagel:
> Note: The URL class which was used strips any paths from the base.

Ohh. How would we handle the trailing slash then 😏

Fallenbagel:
Also urls ending in a / are valid and different to ones without a / and proxies (can) treat them differently

Fallenbagel:
@sambartik Also could you fix the formatting issue so I could quickly release a hot fix?

PR #411, opened 2023-06-18
Update Linux installation instructions in README.md

elch01:
#### Description
Quick update of the linux install instructions and systemd service setup, please reply with feedback if needed.


README.md
```diff
@@ -91,7 +104,7 @@ Environment=NODE_ENV=production
 Type=exec
 Restart=on-failure
 WorkingDirectory=/opt/jellyseerr
-ExecStart=/root/.nvm/versions/node/v18.7.0/bin/node dist/index.js
+ExecStart=node dist/index.js
```
fallenbagel: node has to have an absolute path

Fallenbagel (changes requested):

README.md
elch01: Not required if you added the `-g` flag during npm install, node is then added to /bin and accessible directly. 🙂

fallenbagel: Thats for the absolute path of node binary (best practice for systemd servicefiles). -g flag install might not be how everyone installs node. Very different ways are there.

You can add these instructions (just an example. Change wordings so it's clearer 😀 )
Something along the lines of,
- get your node path `which node`
- usually it's `/usr/bin/node` or in your case wherever-g installs it c:
- assuming it is replace execstart etc 

In systemd service files, it is always recommended to use absolute paths. Say the user installed node using NVM. It will not be in `bin`. It will be in their home directory. Meaning a systemd service cannot work with just a `node dist/index.js`.

Hope you understand my concerns. 😄
Just trynna make it as generic as possible so it applies to most distros. (Some distros have immutable paths so node installation using `-g` flag won't work)

elch01: Yes, the binary was what i meant haha sorry
That's true, i'll commit some more lines clarifying that 😄

Fallenbagel:
> How does this look? :)

Perfect. There're a few changes I wanna do to the prerequisites and the installation of node so mind if I push those to this PR?

elch01:
> Perfect. There're a few changes I wanna do to the prerequisites and the installation of node so mind if I push those to this PR?

Go right ahead!

elch01:
Hello.

Any status on this PR?

README.md
```diff
@@ -63,24 +63,40 @@ yarn start
...
+- Git
+- npm
+
+npm:
 - Nodejs (atleast LTS version)
 - Yarn
-- Git
+  
+```bash
+npm install -g node yarn
```
fallenbagel, line 75: Please revise this to not download node through npm. And instead add this link https://nodejs.org/en/download/package-manager and also mention alternative: https://github.com/nvm-sh/nvm

Keep yarn as the only dependency to be installed using node AFTER installing it using the above methods. Might also want to add instructions for ubuntu/debain to remove `cmdlet` which is also called `yarn` before though.

Fallenbagel (changes requested):
Please make the changes that were commented

PR #447, opened 2023-08-07
Allows Jellyseerr to open jellyfin mobile app instead of website

### LINKED ISSUE #456: Mobile Site: 'Play on Jellyfin' directs to Jellyfin app if installed
### Description

Jellyseerr mobile web frontend is great, but neglects Jellyfin's official mobile application.

### Desired Behavior

Ideal mobile user journey map:
1. Setup of Jellyfin and Jellyseerr ecosystem
2. Request and Download/stream media from Jellyseerr web
3. Select media then 'Play on Jellyfin'
4. Jellyfin IOS/Android app opens for media viewing (as opposed to jellyfin web)

### Additional Context

_No response_

### Code of Conduct

CanardConfit:
#### Description

Currently Jellyseerr makes it possible to launch the media on Jellyseer (or plex) on the web but in the case of a mobile or a tablet, there is often an IOS or Android application which can very well be launched from jellyseerr, which avoids to launch the browser.

Urls schemas allow in particular to launch an application instead of a website. To do this, you must replace the http(s) protocol at the beginning of the URL and specify the application that you want to launch (it is often the application package that is used as the schema). Source: https://en.wikipedia.org/wiki/List_of_URI_schemes

In the case of the Jellyfin IOS application (https://github.com/jellyfin/jellyfin-expo), this is org.jellyfin.expo://.

The idea of my pull-request is therefore to allow Jellyseerr to launch an application from its url scheme if it is detected that the user is a mobile.

The Jellyfin app doesn't support Url Schema queries, so it doesn't take the user to the movie page directly... (he ends up in Home) But that may be implemented in the future!

#### To-Dos

- [X] Successful build `yarn build`
- [X] Translation keys `yarn i18n:extract`

#### Issues Fixed or Closed

--

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "react-ace": "10.1.0",
     "react-animate-height": "2.1.2",
     "react-aria": "3.23.0",
+    "react-device-detect": "^2.2.3",
     "react-dom": "18.2.0",
     "react-intersection-observer": "9.4.3",
     "react-intl": "6.2.10",
--- a/src/components/Common/PlayButton/index.tsx
+++ b/src/components/Common/PlayButton/index.tsx
@@ -25,7 +25,10 @@ const PlayButton = ({ links }: PlayButtonProps) => {
         </>
       }
       onClick={() => {
-        window.open(links[0].url, '_blank');
+        window.open(
+          links[0].url,
+          links[0].url.startsWith('http') ? '_blank' : '_self'
+        );
       }}
     >
       {links.length > 1 &&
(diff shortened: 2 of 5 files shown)


CanardConfit:
Everything works, but I still have to find a way to check if the user has one application or another on his computer... Or if this is not possible, possibly add an option in these settings to know which app throw, like "Prefer to launch Jellyfin iOS instead of the website"

Nobody has any idea?

ronnoceel:
One option is to present a popup to users when they press "Open in Jellyfin" that has them choose if they want to open in a browser or in app. If the user selects app, we can assume it's installed. If they select app when it isn't installed, they get an error message (iOS functionality), which I believe in this situation is harmless. This should only be presented to users on iOS.

This is not an ideal solution but it unblocks this PR.

PR #454, opened 2023-08-16
fix(ui): permission to delete media file match API

xabufr:
#### Description

UI needs to have ADMIN permission to display delete file link, but [API only needs to have MANAGE_REQUESTS](https://github.com/Fallenbagel/jellyseerr/blob/b7282ce9908759975f426b442bd3ca96953399d8/server/routes/media.ts#L176).

#### To-Dos


xabufr:
needs more tests

PR #470, opened 2023-09-07
feat(auth): send real information on login

TOomaAh:
#### Description

When a person logs in, their real ip is not sent and the application specified is Axios.
So I added the correct application name and added the X-Forwarded-For header in the axios request headers during login.

#### To-Dos


Panda260:
Nice liddle Feature

Gauvino (approved):
Test successfully

gauthier-th (approved):

PR #580, opened 2023-11-28
feat: support for forwardAuth

### LINKED ISSUE #1221: Support for Forward Auth
### Description

### Desired Behavior

### Additional Context

### Code of Conduct

Janhouse:
#### Description
Add support for forwardAuth available with nginx, caddy, traefik, etc. which allows use of SSO without exposing any jellyseerr services to unauthenticated users.

Inspired by similar [PR in overseer](https://github.com/sct/overseerr/pull/3105) but slightly simplified, requiring existing user to authenticate.

#### To-Dos

#### Fixes
- Fixes #1221


gauthier-th (changes requested):
Could you please add a bit of documentation for this?

docs/extending-jellyseerr/forward-auth.md
```diff
@@ -0,0 +1,68 @@
+# Forward auth
+
+You can use [forward-auth](https://doc.traefik.io/traefik/middlewares/http/forwardauth/) mechanism to log into Jellyseer.
+
+This works by passing the authenticated user e-mail in `X-Forwarded-User` header by the auth server, therefore enabling single-sign-on (SSO) login.
+
+> The user has to exist, it will not be created automatically.
```
gauthier-th, line 7: Probably better to use a warning instead of a citation.
```suggestion
:::warning
The user has to exist, it will not be created automatically.
:::
```

server/middleware/auth.ts
```diff
@@ -21,6 +21,19 @@ export const checkUser: Middleware = async (req, _res, next) => {
...
+      where: { email: req.header('x-forwarded-user') },
+    });
+
+    if (user) {
+      req.user = user;
+    }
+
+    req.locale = user?.settings?.locale
+      ? user.settings.locale
+      : settings.main.locale;
```
gauthier-th, line 36: This is not necessary, it's a duplicate from below.

overseerr-api.yml
```diff
@@ -186,6 +187,9 @@ components:
         defaultPermissions:
           type: number
           example: 32
+        enableForwardAuth:
+          type: boolean
+          example: true
```
gauthier-th, line 192: You wrote the setting here, but it is not available anywhere.

You should add this as a general setting, by adding a new properties in the `MainSettings` type in `server/lib/settings.ts` and adding a new input in the settings page in `src/components/Settings/SettingsMain/index.tsx`. You should also check that this setting is enabled before allowing login with forwardAuth.

Or did you intend to not make a setting and set it enabled by default?

overseerr-api.yml
```diff
@@ -6939,3 +6947,4 @@ paths:
 security:
   - cookieAuth: []
   - apiKey: []
+  - forwardAuth: []
```
gauthier-th, line 6950: This is not in the OpenAPI spec?
https://spec.openapis.org/oas/latest.html#security-scheme-object-0

src/utils/localRequestHelper.ts
```diff
@@ -0,0 +1,18 @@
+import type { NextPageContext } from 'next/dist/shared/lib/utils';
+import type { GetServerSidePropsContext, PreviewData } from 'next/types';
+import type { ParsedUrlQuery } from 'querystring';
+
+export const getRequestHeaders = (
```
gauthier-th, line 5: Shouldn't this function be called something like `getAuthHeaders` instead of `getRequestHeaders`?

src/utils/localRequestHelper.ts
gauthier-th, line 1: Wdyt about renaming this file to something like `serverSidePropsHelpers`?

hiasr:
LMK if any help is needed here, would love to get this through!

Janhouse:
@hiasr @gauthier-th, feel free to make the necessary changes. I made them when I had some free time but that is currently not the case.

gauthier-th:
Closed in favor of #1341

PR #616, opened 2024-01-06
fix(jellyfinlogin.tsx): remove extra spaces from login info

jimdogx:
If a user accidentally adds a space to their username or password, they will not be able to login and they won't understand why. This change removes any spaces from the username, password and email when they submit their credentials.

#### Description

I was trying to log in and it kept saying my username/password were incorrect.  Turns out, tab completion or something had added a space at the end of my username.  The space is not visible of course, so it looked correct.  Re-typing the password several times didn't work until I realized the username had an extra space.

#### To-Dos

- [X] Successful build `yarn build`
- [] Translation keys `yarn i18n:extract`
- [] Database migration (if required)

#### Issues Fixed or Closed

- Fixes #XXXX

### DIFF
--- a/src/components/Login/JellyfinLogin.tsx
+++ b/src/components/Login/JellyfinLogin.tsx
@@ -84,10 +84,10 @@ const JellyfinLogin: React.FC<JellyfinLoginProps> = ({
         onSubmit={async (values) => {
           try {
             await axios.post('/api/v1/auth/jellyfin', {
-              username: values.username,
-              password: values.password,
-              hostname: values.host,
-              email: values.email,
+              username: values.username.trim(),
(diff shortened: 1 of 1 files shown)


Fallenbagel:
Does jellyfin trim usernames/password when storing?

jimdogx:
> Does jellyfin trim usernames/password when storing?

As it turns out, the server does allow spaces.  I checked with the Jellyfin server team and the intention was to allow "John Doe" but not actually " John Doe" or "John Doe ", however the regex they use (^[\w\ \-'._@]+$) does allow for that.

Which means there could be people out there with extra spaces before/after their username (even though it was not intended to allow this).

PR #626, opened 2024-01-19
feat: display user score in discover view and filter it more flexable.

wjh000123:
#### Description
The score cannot be seen in discover view, but the api has voteAverage data, just show it on the page. And the filter slide has default step=1, which is not very flexable, for I might be interested in movies that is larger than 7.5, or so. Just make the score filter more flexable by set the step to 0.1.

##### [discover] before:
image

##### [discover] after:
image

##### [filter slide] after:
image

#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX

### DIFF
--- a/server/api/themoviedb/index.ts
+++ b/server/api/themoviedb/index.ts
@@ -1,6 +1,5 @@
 import ExternalAPI from '@server/api/externalapi';
 import cacheManager from '@server/lib/cache';
-import { getSettings } from '@server/lib/settings';
 import { sortBy } from 'lodash';
 import type {
   TmdbCollection,
@@ -99,7 +98,6 @@ interface DiscoverTvOptions {
 }
 
 class TheMovieDb extends ExternalAPI {
-  private locale: string;
   private region?: string;
   private originalLanguage?: string;
   constructor({
@@ -119,7 +117,6 @@ class TheMovieDb extends ExternalAPI {
         },
       }
     );
-    this.locale = getSettings().main?.locale || 'en';
     this.region = region;
     this.originalLanguage = originalLanguage;
   }
@@ -128,7 +125,7 @@ class TheMovieDb extends ExternalAPI {
     query,
     page = 1,
     includeAdult = false,
-    language = this.locale,
+    language = 'en',
   }: SearchOptions): Promise<TmdbSearchMultiResponse> => {
     try {
       const data = await this.get<TmdbSearchMultiResponse>('/search/multi', {
@@ -150,7 +147,7 @@ class TheMovieDb extends ExternalAPI {
     query,
     page = 1,
     includeAdult = false,
-    language = this.locale,
+    language = 'en',
     year,
   }: SingleSearchOptions): Promise<TmdbSearchMovieResponse> => {
     try {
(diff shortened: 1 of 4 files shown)


Gauvino:
Does adding a settings by user if they want to see the score or not so it's not in default built-in

wjh000123:
> Does adding a settings by user if they want to see the score or not so it's not in default built-in

That's OK, and where the setting section should this option be in?

Gauvino:
> > Does adding a settings by user if they want to see the score or not so it's not in default built-in
> 
> 
> 
> That's OK, and where the setting section should this option be in?

The most coherent for me I think is in the profile settings on the bottom left when clicking settings

Fallenbagel:
IMHO that makes discover posters very cluttered from an already cluttered view

wjh000123:
> IMHO that makes discover posters very cluttered from an already cluttered view

How about show the score on hover?

Fallenbagel:
> > IMHO that makes discover posters very cluttered from an already cluttered view
> 
> How about show the score on hover?

Dont we already have the watchlist button occupying that space on hover? Imagine already requested ones. It will have media type badge, watchlist button, score badge. It will also kind of make mobile/touch ux a bit harder

wjh000123:
> Dont we already have the watchlist button occupying that space on hover? Imagine already requested ones. It will have media type badge, watchlist button, score badge. It will also kind of make mobile/touch ux a bit harder

Yep, maybe we can do a little change to the UI. Maybe make the watchlist button more compact, like only show the icon. Another proposal is,  how about make the watchlist button the same style as request button, and below it. From my understanding, watchlist and request are some kind of similar action, which stands for the interest.

image

Gauvino:
> > Dont we already have the watchlist button occupying that space on hover? Imagine already requested ones. It will have media type badge, watchlist button, score badge. It will also kind of make mobile/touch ux a bit harder
> 
> 
> 
> Yep, maybe we can do a little change to the UI. Maybe make the watchlist button more compact, like only show the icon. Another proposal is,  how about make the watchlist button the same style as request button, and below it. From my understanding, watchlist and request are some kind of similar action, which stands for the interest.
> 
> 
> 
> image
> 
> 

By the picture you make I can say that it's impossible to make it like that cause don't forget that they're is other language than English and movie translation or series can be more longer so it will break everything I think

Gylesie:
I think displaying rating is such a nice feature to have, would save a ton of clicks into details. However, I agree the current draft design looks awfully clustered, as @Fallenbagel noted. I even think that the favourites icon should be redesigned too to a solid design, because currently it does not fit aesthetically with the overseer's style. I can't really think of a nice way of doing this. 

We could replace the "movie" "series" label with something more simple, but that would be a big change, and I kind of got used to the label and like it. Or maybe some kind of hover animation in place of the label, but not sure, that would not work on mobile UI that great.

Maybe lowering the font size a bit and max. length of the description?

Having it on the side close to the right border?

All of the options above dont feel right to me. Any other ideas?

A-Byte-Sized:
The score could be added on hover in a column underneath the favorite icon. 

image

This is a feature I'm dying to have personally. It's been awhile since there's been an update on this thread. Has it been abandoned or discussion moved elsewhere?

Fallenbagel:
> The score could be added on hover in a column underneath the favorite icon. 
> 
> image
> 

That would interfere with the blacklist button which is also located there 🤔

M0NsTeRRR:
Closing this PR as it has been in draft since 2024, and there is an ongoing discussion about which implementation to choose in #1269.

PR #421, opened 2023-07-04
feat: support for postgresql

zackhow:
Near the same as my PR in overseerr. Opening this here for others to test out. I haven't run into any issues myself so a second set of eyes would be good.

I also tested migrating an overseerr sqlite db but would also like more testing for that if possible.

PR #628, opened 2024-01-19
feat: add postgres support + migrations

### LINKED ISSUE #186: [Feature Request] Add support for external database connections
### Description

I'd love to be able to connect Jellyseerr to an external MySQL or Postgres-compatible database to allow for sharding, replication, and HA deployments. Currently Jellyseerr (and Overseerr) uses an internal SQLite database for request storage.

### Desired Behavior

Ideally, I'd like for there to be a set of environment variables that can be set in order to configure and use an external database connection. That way, the default (with no variables set) would be SQLite and no functionality is broken or removed for simple deployments like this. If the user so desires, they would be able to enable and configure the database, significantly expanding deployment options.

### Additional Context

_No response_

### Code of Conduct

dr-carrot:
#### To-Dos

#### Issues Fixed or Closed

- Fixes #186 

#### TODO:
- [x] An extra migration in the develop branch
- [x] Translation keys?
- [x] Docs
- [x] Test image - `drcarrot/jellyseerr-postgres:latest`
- [x] Move new docs from readme to the newer docs page
- [x] ~Switch date column types to remove timezone~ It is generally recommended to use `timestamp with timezone` aka `timestampz`


Fallenbagel:
Whats the difference between this and #421?

dr-carrot:
> Whats the difference between this and #421?

This PR contains all the changes in #421 plus changes from @ralgar that are also mentioned in that PR. Additionally, this PR contains the migration script required to get postgres running, documentation on how to configure postgres, and improved ssl configuration options

Fallenbagel:
Does this replace sqlite like the other pr did or make postgres optional? (Sorry haven't really had the time to review yet curious to know)

ralgar:
> Does this replace sqlite like the other pr did or make postgres optional? (Sorry haven't really had the time to review yet curious to know)

Both PRs make Postgres optional, using environment variables. SQLite is still the default.

I've only had a glance so far, but it looks pretty good. Appears to be everything from #421, but with all of the issues resolved (hopefully). The original PR is most likely dead, due to inactive author.

If someone can build a container from this PR, I'll gladly fire up a Kubernetes cluster and give it a thorough testing with different configurations.

dr-carrot:
> If someone can build a container from this PR, I'll gladly fire up a Kubernetes cluster and give it a thorough testing with different configurations.

Here's an image for it: `drcarrot/jellyseerr-postgres:latest`
It's been working fine so far, but I'd definitely appreciate some more testing on it

Fallenbagel:
> If someone can build a container from this PR, I'll gladly fire up a Kubernetes cluster and give it a thorough testing with different configurations.

Let me tag this. So a preview build will be created automatically

Fallenbagel:
> > If someone can build a container from this PR, I'll gladly fire up a Kubernetes cluster and give it a thorough testing with different configurations.
> 
> Here's an image for it: `drcarrot/jellyseerr-postgres:latest`
> It's been working fine so far, but I'd definitely appreciate some more testing on it

Oh you built it. No need for preview tag? 🤔

dr-carrot:
> > > If someone can build a container from this PR, I'll gladly fire up a Kubernetes cluster and give it a thorough testing with different configurations.
> > 
> > 
> > Here's an image for it: `drcarrot/jellyseerr-postgres:latest`
> > It's been working fine so far, but I'd definitely appreciate some more testing on it
> 
> Oh you built it. No need for preview tag? 🤔

Honestly, I'm not really sure how to get the configured github CI to do that. I'd be happy to do it another way with a little direction, but this was easiest for me

Fallenbagel:
> Honestly, I'm not really sure how to get the configured github CI to do that. I'd be happy to do it another way with a little direction, but this was easiest for me

I just have to tag this pr as a `preview-*` tag and ci would that in my repo

dr-carrot:
> > Honestly, I'm not really sure how to get the configured github CI to do that. I'd be happy to do it another way with a little direction, but this was easiest for me
> 
> I just have to tag this pr as a `preview-*` tag and ci would that in my repo

I'll give that a go later.
Currently looking into a bug where a single movie request spawns a bunch of media requests

ralgar:
I'm not very experienced with database stuff, but I think there's still quite a bit of work to be done in the migrations. I notice only the initial migration has been ported over to Postgres, and it looks quite heavily modified when there's really only a few keywords that need changing.

I was under the impression that they all should be ported over in a 1:1 manner (as close as possible), not only for functionality, but for maintainability (you should be able diff SQLite vs Postgres effectively, for example).

---

I'm currently getting this error with your container in my reproducible environment, when I attempt to login to Jellyseerr via Jellyfin auth:

```
2024-01-21T18:40:43.979Z [error][Auth]: column User_settings.pushoverSound does not exist
```

Seems to me that it's because of the missing migrations, specifically `1697393491630-AddUserPushoverSound.ts`.

dr-carrot:
> **Edit:** I see you have a commit that accounts for this column. Is it possible that your container is just outdated?
There should be a new build up under `:latest` or more specifically `:main-38`. I will try to get the preview-* build working tomorrow
As far as multiple vs 1 file for, I feel like it makes sense for the first version with postrgres support to only have the one file to create the db in a working (🤞) state. It's not like you would be able to downgrade to a previous version without switching back to sqlite

ralgar:
> There should be a new build up under `:latest` or more specifically `:main-38`. I will try to get the preview-* build working tomorrow

Nice, the new container now logs in successfully. I'll play around with it over the next couple of days, and report back.

>As far as multiple vs 1 file for, I feel like it makes sense for the first version with postrgres support to only have the one file to create the db in a working (🤞) state. It's not like you would be able to downgrade to a previous version without switching back to sqlite

That's true. Like I said I'm not very experienced with database stuff on the application side, so I wasn't sure if it mattered or not. Seems to work fine now anyway!

Fallenbagel:
> > **Edit:** I see you have a commit that accounts for this column. Is it possible that your container is just outdated?
> 
> There should be a new build up under `:latest` or more specifically `:main-38`. I will try to get the preview-* build working tomorrow 

I am pushing the preview build dont worry

Fallenbagel:
> I fixed the multiple requests issue (`:latest` and `main-39`). I will continue running/using it to look for more issues

Should I update the `:preview-postgres`?

dr-carrot:
> Should I update the `:preview-postgres`?
Sure, if you don't mind (though I really don't mind maintaining the other container)
I have been running the latest postgres version for the past few days and it's been working properly thus far

ralgar:
I just finished setting this up, and initial impressions are decent. Adding a couple of movies and series and it seems to work pretty well for the most part.

I do have one issue though - Rick and Morty Season 1 doesn't get marked 'Requested', or show up in 'Recent Requests'. Only seems to happen with that show (so far), and only when season 1 is selected as part of the request. It's definitely an issue with the Postgres implementation, as it doesn't happen when I toggle the env var back to `sqlite` using the same container (`main-39`).

aleksasiriski:
I just wanted to start working on this and found this PR. Great work @dr-carrot. I'll test it in my K8s cluster right away, and afterwards see what code changes were made. Hoping to get this merged ASAP!

aleksasiriski:
@Fallenbagel could you make another tag with these new changes and to push the arm build?

aleksasiriski:
@dr-carrot btw I think that "chore" should never be used if the code changes, those additional debug checks go into "fix"
chore is used when no code changes (so docs or README for example)

dr-carrot:
> @dr-carrot btw I think that "chore" should never be used if the code changes, those additional debug checks go into "fix" chore is used when no code changes (so docs or README for example)

At my job, we usually use it to mean something like "changes that do not affect the application's behavior". But I don't mind abiding by a different interpretation here.

---
The latest build (`:latest`/`:main-50`) includes a small change to the initial migration adding a few `not null` constraints (I considered adding a standalone migration for it, but LMK if you disagree). It fixes the "Rick and Morty" issue reported by @ralgar

Fallenbagel:
> @Fallenbagel could you make another tag with these new changes and to push the arm build?
> until then I made this: `ghcr.io/tminaorg/jellyseerr-postgres:latest` if someone want to test arm64 (no armv7 here)

Will update the preview tag today

Fallenbagel:
Edit: `:preview-postgres` should be updated in approx. 1-2hrs

aleksasiriski:
> Edit: `:preview-postgres` should be updated in approx. 1-2hrs

That tag is only amd64 (probably because this branch needs to merge the change in main for the CI)

Fallenbagel:
> > Edit: `:preview-postgres` should be updated in approx. 1-2hrs
> 
> That tag is only amd64 (probably because this branch needs to merge the change in main for the CI)

No. CI is not required. Preview workflow is separate. I will have to see why it did not build for other archs on the preview workflow

Fallenbagel:
Im very confused. New tags I tested is building for all archs

aleksasiriski:
Hmm, I didn't try pulling on my K8s cluster but I looked on dockerhub and it stated only amd64 for that tag

Fallenbagel:
> Hmm, I didn't try pulling on my K8s cluster but I looked on dockerhub and it stated only amd64 for that tag

Yeah it does say that. And it only built amd64. What I meant was when I created a preview tag of another pr it build all three platforms. Not sure why this pr isn't

aleksasiriski:
@dr-carrot I've been using the latest commits for the past few days (since I posted the first comment) and it's been working great. K8s deployment with 2 replicas, config is mounted in an init container and copied over to jellyseerr so it's completely stateless and declarative as well as highly available. I encountered an issue with requesting BoJack Horseman s1 (while I had all other seasons already downloaded) but I guess that is the same issue as Rick and Morty one, so should be fixed with the newest commits.

onedr0p:
@aleksasiriski I still wouldn't run this application with more than a single replica given that it runs tasks like scanning Plex/JF, Radarr, Sonarr etc... Having more than one replica will runs these tasks X amount of times where X is the number of replicas. This app was **not** architected to be highly available.

aleksasiriski:
> @aleksasiriski I still wouldn't run this application with more than a single replica given that it runs tasks like scanning Plex/JF, Radarr, Sonarr etc... Having more than one replica will runs these tasks X amount of times where X is the number of replicas. This app was **not** architected to be highly available, while postgres is a path forward to this happening it's still not scalable.

That is correct, but I don't mind running those tasks twice given that I can add PodDisruptionBudget and in case one of my nodes goes down the service stays up (;

ralgar:
> The latest build (`:latest`/`:main-50`) includes a small change to the initial migration, adding a few `not null` constraints (I considered adding a standalone migration for it but did not. LMK if you disagree). It fixes the "Rick and Morty" issue reported by @ralgar

Yup, seems to be resolved with the latest build.

I tested Rick and Morty S1, Bojack Horseman S1 (reported by @aleksasiriski), and the Oppenheimer movie, which all displayed the same problem. They're all correctly marked as requested now.

Haven't noticed any other issues so far, but I'll keep testing it over the next few days.

douglasparker:
@Fallenbagel I am setting up a new Jellyseerr instance.

Can I run `:preview-postgres` until this is merged? I am using emby and it looks like people report it's stable for them.

I have incremental backups being made every 5 minutes.

I just don't want to use the preview image if it isn't going to get merged.

onedr0p:
@Fallenbagel Everything is working here as well. Thanks all for adding postgres support!

buroa:
Looks good on my end, but can we get `/app/config/settings.json` info stored as well? Seems like it should be in the database.

dr-carrot:
> Looks good on my end, but can we get `/app/config/settings.json` info stored as well? Seems like it should be in the database.

I think migrating the config file to the db is worth its own issue and PR. But it is something worth looking in to after this is merged

ralgar:
> Looks good on my end, but can we get `/app/config/settings.json` info stored as well? Seems like it should be in the database.

Not sure I agree with this idea. IMO config should be in files and only data should be in the DB, it's simpler and more transparent that way. What benefit do you envision from moving the configuration into the DB?

From my end, all I see is that it will break (and seriously complicate) the declarative configuration workflow that I and @aleksasiriski are both currently using:

>config is mounted in an init container and copied over to jellyseerr so it's completely stateless and declarative

Jellyseerr is the easiest app to configure in my cluster, precisely because of this JSON file. Just copy it into an emptyDir and run `envsubst` on it, and you're good to go. Editing the ConfigMap only takes a few seconds. With *arr apps I have to dig into the DB or source code every time I want to change something, it's not much fun.

onedr0p:
> Not sure I agree with this idea. IMO config should be in files and only data should be in the DB, it's simpler and more transparent that way. What benefit do you envision from moving the configuration into the DB?

I agree with what you're saying to a point however there's a bunch of configuration in this config file that could be offloaded to the database. Obviously things like database, port and some other config should live in env and/or a config file but when it comes to a app that implements sqlite or pg most configuration should live in the database. Otherwise you start mixing configuration between env, file and database and that's not great.

> config is mounted in an init container and copied over to jellyseerr so it's completely stateless and declarative

This is pretty janky however I do it sometimes as well across other apps but there's no guarantee that this config file won't change in the future which will end up breaking this pattern. For example sabnzbd is all one configuration file and the app can run migrations on the config file. If you take a snapshot of this config, do the init container dance with envsubst, and then sabnzbd releases an update that migrates the config to a version your config that you snapshotted is going to be out of date and will potentially break the app.

Another issue you run into is that you essentially make it so configuration in the UI is pointless because those values live in your init container envsubst solution. Any time you want to make a configuration change you must restart the application.

> From my end, all I see is that it will break (and seriously complicate) the declarative configuration workflow that I and @aleksasiriski are both currently using

If the config is in the DB and you're already backing up the database there's no reason to do this. It's not like your deploying new instances of jellyseerr every hour and _need_ declarative configuration for this app. Once it's set up that configuration lives in the database and should be backed up.

Don't get me wrong, I love applications that have full declarative configuration. I don't think jellyseerr really benefits from it due to the reasons I laid out.

douglasparker:
I don't see a point in offloading configuration to the database. It just makes it harder to make changes.

Config files can be live reloaded without a reboot and there isn't a performance benefit to adding it to the database.

onedr0p:
> I don't see a point in offloading configuration to the database. It just makes it harder to make changes.

The amount of times I've to go into the Sonarr or Radarr database to change config is pretty much zero. They expose enough options in the config file to make sure the app starts, so you can change config in the UI.

Basically anything that requires a app restart should be in env or config file, anything that can change during run time should be in the database.

> I don't know, I just don't see a real good reason for the added complexity.

The main problem I'm trying to solve is close to what @aleksasiriski and @ralgar is doing but I don't want to deal with the nuances they are faced with in their solution of templating the config file, which is possibly dealing with upstream config changes and not having access to change settings in the UI and have those reflected in the template.

But I get it, why change it if things work now for the majority of people 🤷‍♂️

aleksasiriski:
Let's move this to a discussion so anyone interested can talk about it there :D I'll join with my perspective as well.

dr-carrot:
@Fallenbagel id appreciate a review when you (or anyone else) gets the chance. I haven’t seen or heard about any issues for a while and I’d prefer to get this merged before more migrations are required

Fallenbagel:
> @Fallenbagel id appreciate a review when you (or anyone else) gets the chance. I haven’t seen or heard about any issues for a while and I’d prefer to get this merged before more migrations are required

There is one issue. Not sure if it's isolated to this pr (but I have not been able to reproduce it on sqlite) #652

Also ill review when I can

dr-carrot:
> There is one issue. Not sure if it's isolated to this pr (but I have not been able to reproduce it on sqlite) #652

I had a look and I don’t think it’s caused by this pr directly. I think it’s either caused by a user being renamed or a bad migration. I’m going to write up an SQLite to Postgres migration guide when I have a little time

douglasparker:
> > There is one issue. Not sure if it's isolated to this pr (but I have not been able to reproduce it on sqlite) #652
> 
> I had a look and I don’t think it’s caused by this pr directly. I think it’s either caused by a user being renamed or a bad migration. I’m going to write up an SQLite to Postgres migration guide when I have a little time

Yep, I'm pretty sure it's because of old tags being in place from when I used Overseerr.

I had assumed that it would see that `ID: Username` didn't exist, so it would make a new tag- and then I could retag stuff made from Overseerr.

It turns out that it simply maps based on `ID` and not the whole tag.

I've gone ahead and wiped out my tags (Not worth the trouble to fix considering some of my users still don't have an ID yet).

Issue #652 should no longer be an issue.

Extremely excited to get this merged into Jellyseerr!

Thank you so much @dr-carrot and @Fallenbagel!

server/utils/DbColumnHelper.ts
```diff
@@ -0,0 +1,21 @@
+import { isPgsql } from '@server/datasource';
+import type { ColumnOptions, ColumnType } from 'typeorm';
+import { Column } from 'typeorm';
+
+const pgTypeMapping: { [key: string]: ColumnType } = {
+  datetime: 'timestamp with time zone',
```
danshilm, line 5: Can this be a timestamp without time zone to match how the datetime is stored in sqlite? I haven't tested that, so am not sure what the behaviour would be.

Edit: looks like [a fork is doing just that](https://github.com/Fat-Cats-LLC/overseerr/commit/866022e91f8edd803e57b833acf2453a64421cc9).

danshilm:
Something that I think would be nice is to update the dev setup to include everything that's needed to work with Jellyseer whether it's with the sqlite or postgres database. That would include:
- including a postgres service in the docker compose
- setting up a `.env.example` file for devs to know what their `.env` file should look like
	-  maybe use `cross-env` so Windows users can also easily contribute?
- update the migration scripts in package json to generate migrations for both sqlite and postgres

danshilm:
There's also a fix I saw a fork have to make that I'm curious to know if someone using this pull's docker image had to do?
[The commit in question](https://github.com/Fat-Cats-LLC/overseerr/commit/90afe067b081e29a9996923ae5fdc2a00a9ed639)

The issue seems to be TypeORM not loading the MediaRequest relations on a Media entity, and therefore dumping all the requests for a Media whenever it's updated and saved.

server/utils/DbColumnHelper.ts
dr-carrot: I'll give it a try and make sure it works first

dr-carrot:
> There's also a fix I saw a fork have to make that I'm curious to know if someone using this pull's docker image had to do? [The commit in question](https://github.com/Fat-Cats-LLC/overseerr/commit/90afe067b081e29a9996923ae5fdc2a00a9ed639)
> 
> The issue seems to be TypeORM not loading the MediaRequest relations on a Media entity, and therefore dumping all the requests for a Media whenever it's updated and saved.

There was something similar in this PR, but it didn't always work quite right. After some experimenting I found that using more strict constraints makes typeorm behave itself in that instance. I haven't heard of or had any issues with it since

server/utils/DbColumnHelper.ts
dr-carrot: Sqlite doesn't support that. In fact, that fork dropped sqlite completely

dr-carrot: [From the readme](https://github.com/Fat-Cats-LLC/overseerr?tab=readme-ov-file#current-features):
> Uses PostgreSQL over SQLite (SQLite database functionality is completely removed!)

danshilm: My question was about Postgres actually. not using a different data type for Sqlite. 
More specifically, whether this can be done without any issues since imo it's always a good idea to store timestamps without the time zone:
```suggestion
  datetime: 'timestamp without time zone',
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "jellyseerr",
-  "version": "0.1.0",
+  "version": "1.7.0",
```
fallenbagel: Please change this back to 0.1.0

CHANGELOG.md
fallenbagel, line 1: Remove this file from this pr

server/utils/DbColumnHelper.ts
fallenbagel: I believe it's best practice to always store time in db without the timezones

captmicr0:
I've been using this branch for a couple months and semi-regularly get these two errors in my Postgres logs.
I did migrate from an existing sqlite database so I'm guessing this error is the result of that.

```
[5593] ERROR:  null value in column "mediaId" of relation "media_request" violates not-null constraint

[5593] DETAIL:  Failing row contains (541, 2, 2024-06-05 00:53:25.891558-04, 2024-06-05 00:53:25.933927-04, movie, null, 3, 3, f, null, null, null, null, null, f).

[5593] STATEMENT:  UPDATE "media_request" SET "mediaId" = $1, "updatedAt" = CURRENT_TIMESTAMP WHERE "id" = $2
```

```
[5593] ERROR:  syntax error at or near ")" at character 1610

[5593] STATEMENT:  SELECT "media"."id" AS "media_id", "media"."mediaType" AS "media_mediaType", "media"."tmdbId" AS "media_tmdbId", "media"."tvdbId" AS "media_tvdbId", "media"."imdbId" AS "media_imdbId", "media"."status" AS "media_status", "media"."status4k" AS "media_status4k", "media"."createdAt" AS "media_createdAt", "media"."updatedAt" AS "media_updatedAt", "media"."lastSeasonChange" AS "media_lastSeasonChange", "media"."mediaAddedAt" AS "media_mediaAddedAt", "media"."serviceId" AS "media_serviceId", "media"."serviceId4k" AS "media_serviceId4k", "media"."externalServiceId" AS "media_externalServiceId", "media"."externalServiceId4k" AS "media_externalServiceId4k", "media"."externalServiceSlug" AS "media_externalServiceSlug", "media"."externalServiceSlug4k" AS "media_externalServiceSlug4k", "media"."ratingKey" AS "media_ratingKey", "media"."ratingKey4k" AS "media_ratingKey4k", "media"."jellyfinMediaId" AS "media_jellyfinMediaId", "media"."jellyfinMediaId4k" AS "media_jellyfinMediaId4k", "watchlist"."id" AS "watchlist_id", "watchlist"."ratingKey" AS "watchlist_ratingKey", "watchlist"."mediaType" AS "watchlist_mediaType", "watchlist"."title" AS "watchlist_title", "watchlist"."tmdbId" AS "watchlist_tmdbId", "watchlist"."createdAt" AS "watchlist_createdAt", "watchlist"."updatedAt" AS "watchlist_updatedAt", "watchlist"."requestedById" AS "watchlist_requestedById", "watchlist"."mediaId" AS "watchlist_mediaId" FROM "media" "media" LEFT JOIN "watchlist" "watchlist" ON "watchlist"."mediaId"="media"."id" AND ("media"."id"= "watchlist"."mediaId" and "watchlist"."requestedById" = $1) WHERE "media"."tmdbId" in ()
```
This one seems to be because there is no value provided for the `WHERE "media"."tmdbId" in ()` part of the statement.

Lebowski89:
Keen to get this merged. Has been working great for me. Jumped over from overseerr just for this postgres support.

croneter:
Same. What's blocking a merge appart from an obvious rebase..?

Fallenbagel (changes requested):

Fallenbagel:
> Same. What's blocking a merge appart from an obvious rebase..?

I added some review comments long ago. Forgot to submit it lol.

Also our focus for now is to get #815 in

nodadyoushutup:
I know that this is not the current priority to get in, but do we have even a hand wave idea of potentially how long it may be? I can continue to use the preview container but I am eagerly awaiting this not being a preview feature :)

joaopedrocg27:
Is there a timeframe to merge this? Since running in kubernetes this is a deal breaker for now

RyuunosukeDS3:
Chiming in to say, I migrated from Overseerr to Jellyseerr for this. I'm also running in Kubernetes, and I'm trying to move everything I can to postgres.

juanlurie:
I’d like to add this to my docker stack as well. Quite keen to have it working with Postgres

gauthier-th:
@Fallenbagel could you run the GitHub checks please?

server/datasource.ts
```diff
@@ -23,16 +60,60 @@ const prodConfig: DataSourceOptions = {
...
+  if (process.env.NODE_ENV === 'production') {
+    if (isPgsql) {
+      return postgresProdConfig;
+    }
+    return prodConfig;
+  } else if (isPgsql) {
+    return postgresDevConfig;
+  }
+  return devConfig;
+}
```
gauthier-th: Seems odd to write the condition this way
```suggestion
  if (process.env.NODE_ENV === 'production') {
    return isPgsql ? postgresProdConfig : prodConfig;
  } else {
    return isPgsql ? postgresDevConfig : devConfig;
  }
```

README.md
gauthier-th, line 1: We moved our documentation to the `docs/` folder. Could you move this please?

server/utils/DbColumnHelper.ts
```diff
@@ -0,0 +1,21 @@
+import { isPgsql } from '@server/datasource';
+import type { ColumnOptions, ColumnType } from 'typeorm';
+import { Column } from 'typeorm';
+// TODO cleanup this file
```
gauthier-th: Is this done?

server/utils/DbColumnHelper.ts
```diff
@@ -0,0 +1,21 @@
...
+const pgTypeMapping: { [key: string]: ColumnType } = {
+  datetime: 'timestamp without time zone',
+};
+
+export function resolveDbType(pgType: ColumnType): ColumnType {
```
gauthier-th, line 8: Why create a function that is only use once and without much content in it, instead of adding this in DbAwareColumn?

server/utils/DbColumnHelper.ts
```diff
@@ -0,0 +1,21 @@
...
+  if (isPgsql && pgType.toString() in pgTypeMapping) {
+    return pgTypeMapping[pgType.toString()];
+  }
+  return pgType;
+}
+
+export function DbAwareColumn(columnOptions: ColumnOptions) {
```
gauthier-th, line 15: And why is this function not used anywhere in this PR?

gauthier-th (changes requested):

gauthier-th:
@nodadyoushutup @joaopedrocg27 I expect this to be merged in approximately 1 month. I can't promise anything though 😅

gauthier-th:
@dr-carrot any update on this?

server/utils/DbColumnHelper.ts
dr-carrot: Removed

server/utils/DbColumnHelper.ts
dr-carrot: It's from the OG creator of the psql stuff. I removed it since it works differently now

server/utils/DbColumnHelper.ts
dr-carrot: It is now

README.md
dr-carrot: I moved it. Let me know if it makes sense

server/datasource.ts
dr-carrot: Yeah... Updated it

docs/getting-started/database-config.mdx
gauthier-th, line 1: I think this should be in the `docs/extending-jellyseerr` folder instead of `docs/getting-started` because it does not describe a setup process, but how to configure an additional tool.

docs/getting-started/database-config.mdx
```diff
@@ -0,0 +1,52 @@
...
+# Configuring the Database
+
+Jellyseerr supports SQLite and PostgreSQL. The database connection can be configured using the following environment variables:
+
+#### SQLite Options
+
+```dotenv
+DB_TYPE="sqlite" # Which DB engine to use, either "sqlite" or "postgres". The default is "sqlite"
+CONFIG_DIRECTORY="config" # The path to the config directory where the db file is stored
+DB_LOG_QUERIES="false" # Whether to log the DB queries for debugging
```
gauthier-th: Maybe write explicitly the default options?

docs/getting-started/database-config.mdx
```diff
@@ -0,0 +1,52 @@
...
+DB_USE_SSL="false" # Whether to enable ssl for database connection
+
+# The following options can be used to further configure ssl:
+DB_SSL_REJECT_UNAUTHORIZED="true" # Whether to reject ssl connections with unverifiable certificates i.e. self-signed certificates without providing the below settings
+DB_SSL_CA= # The CA certificate to verify the connection, provided as a string
+DB_SSL_CA_FILE= # The path to a CA certificate to verify the connection
+DB_SSL_KEY= # The private key for the connection in PEM format, provided as a string
+DB_SSL_KEY_FILE= # Path to the private key for the connection in PEM format
+DB_SSL_CERT= # Certificate chain in pem format for the private key, provided as a string
+DB_SSL_CERT_FILE= # Path to certificate chain in pem format for the private key
```
gauthier-th: Same here. Write explicitly what is the default value, and if it mandatory (because for instance the default value of `DB_USER` is not `jellyseerr`).

gauthier-th (changes requested):

gauthier-th:
Could you please also take a look at the cypress tests? Looks like the PR is breaking them. Could you find out why and fix your code or the tests accordingly?

docs/getting-started/database-config.mdx
dr-carrot: I made some changes to hopefully make it a little more clear

server/entity/Media.ts
```diff
@@ -129,10 +130,10 @@ class Media {
   @UpdateDateColumn()
   public updatedAt: Date;
 
-  @Column({ type: 'datetime', default: () => 'CURRENT_TIMESTAMP' })
+  @DbAwareColumn({ type: 'datetime', default: () => 'now()' })
   public lastSeasonChange: Date;
 
-  @Column({ type: 'datetime', nullable: true })
+  @DbAwareColumn({ type: 'datetime', default: () => 'now()' })
```
gauthier-th: Genuine question: why is this `DbAwareColumn` only necessary here?

dr-carrot: It seems to break the cypress tests when the db gets seeded. I've seen a different type of column used recently, but I'm not sure I could get a way with simply swapping the annotation

gauthier-th: So this has been implemented just to pass the cypress tests?
It doesn't feel right to modify the codebase just to be compliant with the tests.
Couldn't something else be changed in the cypress tests to fix this issue? If you're struggling to find a solution i can take a deeper look into it.

fallenbagel: > It doesn't feel right to modify the codebase just to be compliant with the tests.

I agree. Especially cypress tests.

dr-carrot: It does fix the cypress tests, but I think that's because it also fixes something with typeorm. I didn't see the issue with uses of the @CreationDateColumn but I'd need to make sure that doesn't break existing SQLite or anything and I'm not sure when I'll have time to do that

dr-carrot: Shockingly I had time today. So postgres doesn't support a column type of datetime. This doesn't seem to be an issue at runtime, but it does cause issues with certain non-runtime uses of the typeorm utilities. In the case of cypress, it's the call to `datasource.synchronize()`.
When using the special date columns (`@CreateDateColumn`), typeorm is able to figure out which type to use dynamically, but it can't for regular columns. So if we still want to play by the rules of typeorm then I believe we need the date field to be dynamic in a way that is like this.
Surprisingly, these 2 fields in `Media` seem to be the only `Dates` that are set manually in the code.
With the latest changes, I was able to run the cypress tests successfully against both sqlite and postgres

gauthier-th: Thanks, it makes more sense now.
Let me do some more tests and it should be ready to merge.

gauthier-th (changes requested):
I just came across this error while requesting a media:
```
SQLITE_ERROR: no such function: now
```
Could you please fix it?

server/entity/Media.ts
```diff
@@ -127,10 +132,10 @@ class Media {
   @UpdateDateColumn()
   public updatedAt: Date;
 
-  @Column({ type: 'datetime', default: () => 'CURRENT_TIMESTAMP' })
+  @DbAwareColumn({ type: 'datetime', default: () => 'CURRENT_TIMESTAMP' })
   public lastSeasonChange: Date;
 
-  @Column({ type: 'datetime', nullable: true })
+  @DbAwareColumn({ type: 'datetime', default: () => 'CURRENT_TIMESTAMP' })
   public mediaAddedAt: Date;
```
gauthier-th: I'm sorry, again an issue on my setup :(
I just tested with my setup on 2.1.0, and i have this error while starting Jellyseerr:
```
2024-12-04T21:19:34.233Z [error]: QueryFailedError: SQLITE_CONSTRAINT: NOT NULL constraint failed: temporary_media.mediaAddedAt
    at Statement.handler (\jellyseerr\node_modules\.pnpm\typeorm@0.3.11_pg@8.11.0_sqlite3@5.1.4_encoding@0.1.13__ts-node@10.9.1_@swc+core@1.6.5_@swc+h_wzz2xcsisc7bz3xl7invzcnodu\node_modules\typeorm\src\driver\sqlite\SqliteQueryRunner.ts:113:26)
```
This is probably coming from the removal of the `nullable: true`. Some media may have a null `mediaAddedAt`.
I suggest adding back the `nullable: true`, and add a comment to explain why it is needed.
```suggestion
  @DbAwareColumn({ type: 'datetime', default: () => 'CURRENT_TIMESTAMP', nullable: true })
  public mediaAddedAt: Date;
```

gauthier-th (changes requested):

server/datasource.ts
```diff
@@ -1,18 +1,55 @@
-import 'reflect-metadata';
+import fs from 'fs';
+import * as process from 'process';
```
gauthier-th: You don't need to import that, `process` is available in the global context by default.
```suggestion
```

gauthier-th (changes requested):

gauthier-th (approved):
LGTM! Thanks for your contribution :smile: 

I'd like a few more people to test this before we merge this into develop, and especially to test with SQLite, because the database is at the core of the app and SQLite is what will be used by most of the users.
I just updated the preview tag: `preview-postgres`

gauthier-th:
Hey everyone, could a few of you test and give me feedback to make sure everything works fine on various setups?

Gylesie:
From the current :develop tag to this feature tag it failed with this SQLite error:

```
[info]: Commit Tag: 00e18a272a8842f0c3d59924f37967f9cbfb413e 
[info]: Starting Overseerr version develop-00e18a272a8842f0c3d59924f37967f9cbfb413e 
[debug][Settings Migrator]: Checking migration '0001_migrate_hostname.js'... 
[debug][Settings Migrator]: Checking migration '0002_migrate_apitokens.js'... 
[debug][Settings Migrator]: Checking migration '0003_emby_media_server_type.js'... 
[info][Notifications]: Registered notification agents 
[info][Jobs]: Scheduled jobs loaded 
[info][Server]: Server ready on port 5055 
[debug][Jobs]: Starting scheduled job: Download Sync 
[info][Jobs]: Starting scheduled job: Jellyfin Recently Added Scan 
[info][Jellyfin Sync]: Jellyfin Sync Starting {"sessionId":"xxxxxx"}
[error][Jellyfin Sync]: Sync interrupted {"errorMessage":"SQLITE_ERROR: no such column: User_settings.region"}
```

gauthier-th:
@dr-carrot last fix I hope :pray: 
We merged #993 a few days ago, this PR splits the `region` field into two `discoverRegion` and `streamingRegion` in the User entity, hence the error above.
Could you please rebase and fix it?

gottsman:
> On previous note, I think https://github.com/Fallenbagel/jellyseerr/blob/00e18a272a8842f0c3d59924f37967f9cbfb413e/docs/extending-jellyseerr/database-config.mdx got it all covered. I'll give it a try and report back after 6 days.

A clean install worked for me so I tried an existing migration.  A question on the documentation... From the notes, it mentions using this command to import the data. What is the significance of "/pgloader/pgloader.load" volume mount? Where would I find that file?

`docker run --rm -v config/db.sqlite3:/db.sqlite3:ro -v pgloader/pgloader.load:/pgloader.load ghcr.io/ralgar/pgloader:pr-1531 pgloader --with "quote identifiers" --with "data only" /db.sqlite3 postgresql://{{DB_USER}}:{{DB_PASS}}@{{DB_HOST}}:{{DB_PORT}}/{{DB_NAME}}`

gauthier-th (approved):
LGTM!
@Fallenbagel do you want to review before we merge this?

Fallenbagel (approved):

gauthier-th:
@dr-carrot thanks again for your contribution!

PR #680, opened 2024-03-09
Adding an easy login to simplify logging in as a default user

phyzical:
Adds logic to allow an admin to enable "easy login", when enabled it will let you choose a user to be logged in as when choosing the default login option.

Idea behind this is to use this when running behind another layer of authentication, i.e authelina or behind a vpn. 

So basically just removes the need for and usernames or passwords for general users

image
image


phyzical:
Happy to change the wording around "easy" if theres a better term you can think of

gauthier-th:
Don't you think “default login” would be clearer than “easy login”?

gauthier-th:
Removing usernames and passwords will introduce significant security risks, even if it's just one of the available options. This could lead to potential vulnerabilities in the future. For reference, even Jellyfin removed this feature some time ago due to the major security concerns it posed.

phyzical:
at the end of the day if a user makes a username and password with username password they will be hacked too.

i get why you would want to deny this. but stupid people will be always be stupid.

/shrug

PR #748, opened 2024-05-13
fix(logging): handle media server connection refused error/toast

Fallenbagel:
#### Description
Properly log as connection refused if the jellyfin/emby server is unreachable. Previously it used to throw a credentials error which lead to a lot of confusion.

image
image

#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX


Fallenbagel:
**TODO:**
- [ ] refactor this PR to use custom error types and error codes instead of abusing error messages

server/types/error.ts
gauthier-th, line 1: Why do you create this new ApiError? I understand the `/auth/jellyfin` route can fail for different reason and this was unclear before, but why this whole new thing?
Every other routes in the application are managed without needing this ApiError thing, it seems me weird to implement it just for this use case (and leave the rest with the current behavior, which is not the same as this PR)

In my opinion you could just have solved it with different HTTP error codes, depending on the error:
- `INVALID_URL`: 502 Bad Gateway: "The server was acting as a gateway or proxy and received an invalid response from the upstream server."
- `INVALID_CREDENTIALS`: 401 Unauthorized: The authentication has failed
- `INVALID_URL`: 403 Forbidden: The request was understood, but the user don't have the required permission

This would also simplify the PR.

fallenbagel: That was how it was before. 
See:
https://github.com/Fallenbagel/jellyseerr/blob/develop/src%2Fcomponents%2FLogin%2FJellyfinLogin.tsx#L96-L100

https://github.com/Fallenbagel/jellyseerr/blob/develop/server%2Froutes%2Fauth.ts#L415-L453

Using http error codes. But it turned into a big mess. And the new error type is not just for this route

I want to refactor all errors to use a custom error type so we can use an enum. It's much more readable that way imo than using code numbers (or how it was implemented before where error messages were abused to death. 

Making the code more readable would mean easier for future contributors too. In addition, we have a few errors, in theory, is the same http error code, so it makes it a whole mess too. Sort of unreadable, and difficult to manage the code login. That was my reasoning. Bad idea?

fallenbagel: Although this PR does not make use of the new error type anywhere else, my plan is to make a follow up pr to start using the custom error+error code enums.

In addition, it also makes it easier to add logic in the frontend since we won't have to worry about arbitrary error messages or similar http codes when we can just use the error enum

gauthier-th: I can see your point. I don't think it's a bad idea, I just wanted to know why this choice. And if you think it would make the code more readable well let's go

gauthier-th: One other comment though: why ApiErrorCode and AuthErrorCode? Couldn't this be merged into only one enum? Will there be that much of AuthErrorCode? It could be confusing to have both while both are an error code returned by the api.

fallenbagel: > One other comment though: why ApiErrorCode and AuthErrorCode? Couldn't this be merged into only one enum? Will there be that much of AuthErrorCode? It could be confusing to have both while both are an error code returned by the api.

That was something I was wondering. Whether to seperate or not. Not sure if there would be other auth errors but just incase for future I added. But better to merge them right? 🤔

gauthier-th: I think so yeah. It's more simple like that and if we don't see any use case it's probably not necessary (at least for now)

gauthier-th (commented):
I don't want to be annoying, but why rename to NetworkError? An authentication error isn't really a network related error?
ApiError seemed fine to me.

Fallenbagel:
Oh so for other errors that aren't api related, we make more classes and enums? 🤔

gauthier-th:
> Oh so for other errors that aren't api related, we make more classes and enums? 🤔

I was thinking that any error returned by the server would be ApiError, since our server is the api?

Fallenbagel:
> > Oh so for other errors that aren't api related, we make more classes and enums? 🤔
> 
> I was thinking that any error returned by the server would be ApiError, since our server is the api?

What about external api requests

Fallenbagel:
> External API requests are always made by the server anyway?
> I would have thought that we have only one type of error returned by all the endpoints of the server.

Makes sense. Let me update real quick

gauthier-th (approved):
LGTM!

PR #762, opened 2024-05-23
fix(jellyfinapi): use external api class for jellyfin api requests

### LINKED ISSUE #387: ~25 requests to jellyfin every 5 minutes?
### Description

Pihole on the local network revealed about 25 dns queries to jellyfin every 5 minutes. I traced this back to jellyseerr. Why so many?

This results in ~ 7200 queries per day from jellyseerr alone.

Let me know if I can help with any logs.

### Version

1.5.0

### Steps to Reproduce

Setup docker container with jellyfin and jellyseer and connect them.
let it run.

### Screenshots

Screenshot 2023-05-19 at 00-42-18 Pi-hole - pihole

Pi-hole data from several days. Jellyfin domain receives abnormal requests

### Logs

_No response_

### Platform

desktop

### Device

/

### Operating System

/

### Browser

/

### Additional Context

_No response_

### Code of Conduct

### LINKED ISSUE #728: DNS spam
### Description

There seems to be an issue that makes Jellyseerr spam my DNS with aaaa requests for my internal Jellyfin URL. Not sure why there is even a request for an aaaa record since I don't use ipv6.

I tested this on 1.7.0 because that is the version I had. Updated to 1.8.1 because I thought maybe it was fixed in the meantime but it wasn't.

Not sure what triggers this. Maybe it's a sync job. This has been going on for a while but I only just noticed because my primary DNS would ratelimit the IP for this machine just long enough for my secondary DNS to take over and serve the requests. Then the secondary DNS would ratelimit the IP just in time for the primary to take over again.

### Version

1.7.0., 1.8.1

### Steps to Reproduce

I used Pi hole to monitor traffic volume
On the host where Jellyseerr is installed I used: sudo tcpdump -i ens18 'dst port 53' to monitor for outgoing DNS requests.
I am running Jellyseerr in a docker container so I added this to my compose file
` extra_hosts:
      - "internal.mrga.dev:IP"
`
This adds the local ip in the container hosts file but I don't like it as a permanent solution.

### Screenshots

Screenshot 2024-04-25 163944
Screenshot 2024-04-25 164027
secondary dns Screenshot 2024-04-25 170603

### Logs

```shell
Will add if needed.
```

### Platform

desktop

### Device

N/A

### Operating System

N/A

### Browser

N/A

### Additional Context

_No response_

### Code of Conduct

Fallenbagel:
#### Description
This PR refactors jellyfin api requests to be handled by the external api class to be consistent with how other external api requests are made. This refactor would also mean we can easily refactor jellyseerr to use apikeys instead of authtokens. In addition, it would be easier to add axios dns resolution caching

#### To-Dos

#### Issues Fixed or Closed

- Related to #728 
- Related to #387


server/constants/error.ts
```diff
@@ -1,5 +1,7 @@
 export enum ApiErrorCode {
   InvalidUrl = 'INVALID_URL',
   InvalidCredentials = 'INVALID_CREDENTIALS',
+  InvalidAuthToken = 'INVALID_AUTH_TOKEN',
   NotAdmin = 'NOT_ADMIN',
+  Generic = 'GENERIC',
```
gauthier-th: Just a little comment: why generic? Something like "Unknown" would be more meaningful in my opinion

.github/workflows/snap.yaml
```diff
@@ -2,7 +2,7 @@ name: Publish Snap
 
 # turn off edge snap builds temporarily and make it manual
 
-# on: 
+# on:
```
gauthier-th: Weird to have this change in this PR?

server/constants/error.ts
fallenbagel: Okay makes sense

.github/workflows/snap.yaml
fallenbagel: Ah i ran prettier on the whole repo and it must have fixed that too. I'll remove and fix that seperately

gauthier-th: Yep, and we should do a refactor to change all the errors returned by the api to the new ApiError

fallenbagel: yes. Will add that to the project board

server/constants/error.ts
```diff
@@ -1,5 +1,7 @@
 export enum ApiErrorCode {
   InvalidUrl = 'INVALID_URL',
   InvalidCredentials = 'INVALID_CREDENTIALS',
+  InvalidAuthToken = 'INVALID_AUTH_TOKEN',
   NotAdmin = 'NOT_ADMIN',
+  UNKNOWN = 'UNKNOWN',
```
gauthier-th: Not consistent with case?

gauthier-th (changes requested):

server/constants/error.ts
fallenbagel: Oh yeah!
I was half asleep when I made this change 💀

gauthier-th (approved):

PR #773, opened 2024-05-25
refactor(jellyfin): abstract jellyfin hostname, updated ui to reflect it, better validation

fallenbagel:
> [!WARNING]
BREAKING CHANGE: Jellyfin settings now does not include a hostname. Instead it abstracted it to ip, port, useSsl, and urlBase. However, migration of old settings to new settings should work automatically.
#### Description
This PR refactors and abstracts jellyfin hostname into, jellyfin ip, jellyfin port, jellyfin useSsl, and jellyfin urlBase. This makes it more consistent with how plex settings are stored as well. In addition, this improves validation as validation can be applied seperately to them instead of as one whole regex doing the work to validate the url.
UI was updated to reflect this.

image

#### To-Dos

~- [ ] Database migration (if required)~
- [x] Abstract settings migration logic
- [x] Figure out why cypress tests are failing

#### Issues Fixed or Closed

- Fixes #XXXX


server/constants/error.ts
```diff
@@ -2,4 +2,7 @@ export enum ApiErrorCode {
   InvalidUrl = 'INVALID_URL',
   InvalidCredentials = 'INVALID_CREDENTIALS',
   NotAdmin = 'NOT_ADMIN',
+  SyncErrorGroupedFolders = 'SYNC_ERROR_GROUPED_FOLDERS',
+  SyncErrorNoLibraries = 'SYNC_ERROR_NO_LIBRARIES',
+  Generic = 'GENERIC',
```
gauthier-th: This will cause merge conflict (the Generic)

gauthier-th (changes requested):

server/constants/error.ts
fallenbagel: Oh yeah I'll resolve it

gauthier-th (dismissed):
Works on my config. As it is a critical breaking change, could we have more people testing and reviewing this?

Fallenbagel:
> Works on my config. As it is a critical breaking change, could we have more people testing and reviewing this?

I will be creating a preview tag for this and try to get people to test this

Fallenbagel:
This pr should be available to test using the tag:
`preview-pr773`

server/lib/settings.ts
```diff
@@ -637,8 +639,56 @@ class Settings {
     const data = fs.readFileSync(SETTINGS_PATH, 'utf-8');
 
     if (data) {
-      this.data = merge(this.data, JSON.parse(data));
-      this.save();
+      const oldJellyfinSettings = JSON.parse(data).jellyfin;
+
+      // Migrate old settings
```
gauthier-th: A comment afterwards: couldn't this be abstracted somewhere else? Things might become messy here with more migration.
Could be interesting to have something doing config 1.8.1 to 1.9.0, 1.10.0 to 1.10.1, etc

Fallenbagel:
> In console log I get this
> 
> ```
> "Jellyfin login is disabled"
> ```
> 
> ```
> "cookie 'connect.sid' required"
> ```

A new preview tag should be available in about half an hour with a fix (when all the checks in this PR turns green). You'll just need to pull it.

server/lib/settings.ts
fallenbagel: Yes. We can create a migration class/function to handle setting migrations

sgtcoder:
Sounds good, let me know when it's updated.

Gauvino (approved):
only migration problem but, other else looks good !

sgtcoder:
I can check this out later today. I tried 3 days ago and it was having the same issue.

Fallenbagel:
> I can check this out later today. I tried 3 days ago and it was having the same issue.

There was a migration issue. I did not fix it 3 days ago. 
I just did.

gauthier-th:
@sgtcoder could you test it one more time please?

server/routes/settings/index.ts
```diff
@@ -252,11 +255,59 @@ settingsRoutes.get('/jellyfin', (_req, res) => {
   res.status(200).json(settings.jellyfin);
 });
 
-settingsRoutes.post('/jellyfin', (req, res) => {
+settingsRoutes.post('/jellyfin', async (req, res, next) => {
```
gauthier-th, line 258: I don't really understand what you're doing here. What is the point of all this new code?

server/routes/settings/index.ts
```diff
@@ -252,11 +255,59 @@ settingsRoutes.get('/jellyfin', (_req, res) => {
...
+    const jellyfinClient = new JellyfinAPI(
+      getHostname(tempJellyfinSettings),
+      admin.jellyfinAuthToken ?? '',
+      admin.jellyfinDeviceId ?? ''
+    );
+
+    const result = await jellyfinClient.getSystemInfo();
+
+    if (!result?.Id) {
+      throw new ApiError(result?.status, ApiErrorCode.InvalidUrl);
```
gauthier-th, line 280: Why throw an ApiError to finally catch it just after?

server/routes/settings/index.ts
```diff
@@ -252,11 +255,59 @@ settingsRoutes.get('/jellyfin', (_req, res) => {
...
+      });
+    } else {
+      logger.error('Something went wrong', {
+        label: 'API',
+        errorMessage: e.message,
+      });
+
+      return next({
+        status: 500,
+        message: 'Something went wrong',
```
gauthier-th: Don't we have an ApiErrorCode for this?

cypress/config/settings.cypress.json
gauthier-th, line 1: Is this linked to this PR?

server/routes/settings/index.ts
fallenbagel: This is relevant to how the new abstracted jellyfin hostname settings are saved. Previously when it was being saved, it didn't do a check. It just saves. Even if it's wrong.

Now it will check if the url is reachable before saving. This is consistent with how plex settings are saved if you scroll up a little

server/routes/settings/index.ts
fallenbagel: Should I just log it here then?

server/routes/settings/index.ts
fallenbagel: Oh yeah I missed out on that one

cypress/config/settings.cypress.json
fallenbagel: Yes. We changed how settings work. This changes how cypress settings works as well.

server/routes/settings/index.ts
gauthier-th: Ok, thanks for the clarification. Make sense.

cypress/config/settings.cypress.json
gauthier-th: Even all the `schedule` properties?

fallenbagel: > Even all the `schedule` properties?

Yes. Basically settings were missing all those and it made cypress tests fail only on this pr. Every other pr doesn't seem to care. Because i don't think the startJobs() was working until this pr when we changed how settings worked

server/routes/settings/index.ts
gauthier-th: Yes I think it's cleaner this way.

gauthier-th (approved):

PR #776, opened 2024-05-27
Update docker-compose.yml

robertocmfgt:
#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX


Fallenbagel (commented):
Uh what's the purpose of this?
This docker compose is for development purposes and that's why the config folder is being mounted onto the current directory and not inside another folder so `.:/app` is what's needed.

Similarly with other mounts as well.

PR #778, opened 2024-05-28
fix(auth): case-sensitive logins not updating authtokens

ThowZzy:
#### Description
Having a Jellyfin username with capital letters and logging into Jellyseerr with the username in lower case would prevent the auth token to be updated with a new one.

So, instead of comparing the usernames from the input field and the Jellyfin's auth response, we are now comparing the user ID stored in our DB and the one from Jellyfin's auth response.

#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX

### DIFF
--- a/server/routes/auth.ts
+++ b/server/routes/auth.ts
@@ -325,7 +325,7 @@ authRoutes.post('/jellyfin', async (req, res, next) => {
       await userRepository.save(user);
     }
     // User already exists, let's update their information
-    else if (body.username === user?.jellyfinUsername) {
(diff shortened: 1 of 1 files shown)


Fallenbagel (approved):
LGTM!

gauthier-th (approved):
Lgtm

gauthier-th:
@all-contributors please add @ThowZzy for code

Fallenbagel:
> @all-contributors please add @ThowZzy for code

Thanks. I just remembered as well 😀

Issue #96, opened 2022-05-20
[Feature Request] Music / Audio support


Pheggas:
#### Description
Hello. I'd like to use Jellyseerr but i only use Jellyfin for music (and music videos) so i can't right now. However, added music support would be really helpful of course.

#### Data grabbing
There are some great sources from where Jellyseerr can download new data - [Musicbrainz](https://musicbrainz.org/), [Last.FM](https://www.last.fm/), and also some media providers like [Qobuz](https://www.qobuz.com/)

I don't think this feature request needs additional info.

el-tiuri:
This, combined with a connection to Deemix (or possibly [AMR](https://github.com/RandomNinjaAtk/docker-amd)) would be great!

dcuellar:
Yes. This would be great. I've installed this and will be keeping track of it. I noticed that the version I have is 1.2.1. It was said that this feature is on the roadmap for 1.2.0. Does that mean it's already on there... because I don't see it. Or does it mean that it will happen in 1.2.x?

gouthamkumaran:
Music support with sync to lidarr would be great.

Extarys:
I'd like to find a project to keep track of what I want to buy in the future (wishlist) and get notifications about new albums from my favourite artists. This would be dope.

sunjam:
That sounds more like a totally different service for tracking artists.
Try finding what already exists and see if they offer an API.

CollinJoseph:
I would love to work on adding lidar support if no one is working on it already.

Fallenbagel:
> I would love to work on adding lidar support if no one is working on it already. 

You're more than welcome to do so c:

CollinJoseph:
I just wanted to make sure that no one else is working on it. Because it would be a waste of time. Thank you for letting me know :)

mukul-mehta:
Any updates on this? If not, I could pick it up and try

dcuellar:
> Any updates on this? If not, I could pick it up and try

Someone else has asked but they. have not followed up since. I've been eagerly waiting for this feature as it would make it perfect for my use. Hope you or someone picks it up.

CollinJoseph:
Sorry for the late reply. I have been quite busy with university and work
the past few months. And, I have made little progress.


A few reasons for my slow progress and lack of motivation are:


1. There is not a comprehensive database for music like TMDB. It might be
because of the genre I am listening to or the language.


[...]

PR #782, opened 2024-05-30
feat: add Lidarr and MusicBrainz support

Svagtlys:
This PR aims to add Lidarr and MusicBrainz support to Jellyseerr to address feature request #96 

I'm pretty new to this kind of coding, so I'm mostly just making changes, running a build, and fixing bugs as they come up. I'm using https://github.com/sct/overseerr/pull/3800 by @anon0002 as a guide/code source for the heftier bits like the Lidarr API classes.

In feature request #96 @zachbakerdev mentioned using Deezer instead of MusicBrainz as it has more info on covers, but I'm not sure that Lidarr supports that and I've not interacted with Deezer before, so I'm currently sticking with MusicBrainz. I'm happy to also add Deezer support at some point in the future.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,6 +66,7 @@
     "node-cache": "5.1.2",
     "node-gyp": "9.3.1",
     "node-schedule": "2.1.1",
+    "nodebrainz": "^2.1.1",
     "nodemailer": "6.9.1",
     "openpgp": "5.7.0",
     "plex-api": "5.3.2",
--- a/server/api/jellyfin.ts
+++ b/server/api/jellyfin.ts
@@ -37,7 +37,7 @@ interface JellyfinMediaFolder {
 }
 
 export interface JellyfinLibrary {
-  type: 'show' | 'movie';
+  type: 'show' | 'movie' | 'music';
   key: string;
   title: string;
   agent: string;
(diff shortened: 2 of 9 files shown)


zachbakerdev:
One of the other considerations I had was instead of using the cover art archive that Musicbrainz uses was searching on Musicbrainz for the albums, then mapping those albums to Deezer instead of the other way around. Just a consideration.

ano0002:
It seems like you forgot to add the type definitions I had to add so it will not work I think.
[I'm talking about this](https://github.com/ano0002/overseerr/blob/overseer-with-lidarr-support/server/types/nodebrainz.d.ts)
But since I don't know how jellyseer is setup you might not need it

Svagtlys:
@ano0002 I actually had a few questions about your implementation of the nodebrainz API, for example why didn't you copy the interfaces of TMDB with all of the Result interfaces?

I haven't added all of your work into this PR (yet) as I'm trying to learn a bit about Typescript/Javascript and how different people code as I'm doing this :D I'll admit I also don't have a deep understanding of how Jellyseerr functions so I'm trying to learn that as well to determine which nodebrainz features are needed.

ano0002:
> @ano0002 I actually had a few questions about your implementation of the nodebrainz API, for example why didn't you copy the interfaces of TMDB with all of the Result interfaces?
> 
> I haven't added all of your work into this PR (yet) as I'm trying to learn a bit about Typescript/Javascript and how different people code as I'm doing this :D I'll admit I also don't have a deep understanding of how Jellyseerr functions so I'm trying to learn that as well to determine which nodebrainz features are needed.

I'll try to answer all those question when I get some time as right now I have all my final exams.

ano0002:
> @ano0002 I actually had a few questions about your implementation of the nodebrainz API, for example why didn't you copy the interfaces of TMDB with all of the Result interfaces?
> 
> I haven't added all of your work into this PR (yet) as I'm trying to learn a bit about Typescript/Javascript and how different people code as I'm doing this :D I'll admit I also don't have a deep understanding of how Jellyseerr functions so I'm trying to learn that as well to determine which nodebrainz features are needed.

I didn't copy them because I didn't want to send the same kind of info. I could have reused it and modified it to take more parameters but that would have meant I'd have to rewrite every other time when the interface was already used so I chose to make a separate one.

PR #793, opened 2024-06-02
revert(api): ip forwarding breaks windows docker desktop setups when externally accessed

Fallenbagel:
#### Description
IP forwarding is broken on windows docker desktop in bridge mode. Windows does not support `network_mode: host`, therefore, this feature breaks external access completely. Until a better method can be found, #470 has be reverted.

IP forwarding is not an issue when using windows baremetal installs, however, all docker desktop windows users would be unable to use jellyseerr externally unless #470 is reverted.

#### To-Dos

#### Reverts

- Reverts #470

### DIFF
--- a/server/api/jellyfin.ts
+++ b/server/api/jellyfin.ts
@@ -123,24 +123,14 @@ class JellyfinAPI extends ExternalAPI {
 
   public async login(
     Username?: string,
-    Password?: string,
-    ClientIP?: string
+    Password?: string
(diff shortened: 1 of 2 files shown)


gauthier-th (approved):
lgtm

Fallenbagel:
Closing as this is not the case. Tests has shown that it does work

Fallenbagel:
Closing as this is not the case. Tests has shown that it does work

Issue #520, opened 2023-11-01
Movie is launched in 4K instead of 1080p when both available


zer0ish:
### Description

Everything is in Unraid dockers with Emby as the media server.

When I have a movie both in 1080p(in my 1080p folder) and a 4K of the movie(in a 4K folder) as seen in SC2 Jellyseer opens emby website to the 4K version even though I selected the 1080p version in Jellyseer. 

Both the 1080p and the 4K have unique Item Id's in emby as seen in SC3

From SC1, if I click Available or Play on emby, the expected behavior would be to open the 1080p version.
If I click the 4K available or the Play 4K on Emby it should open the 4K version.

All my 4K are in their own folder and all my 1080 are in their own folder(SC2) so clicking on either should know where the correct movie is since Jellyseer has the path to that movie and Emby has a unique item id for each.

### Version

1.7.0

### Steps to Reproduce

Got to the movie

Click Available

and then

Go to movie

Click 4K available

### Screenshots

![image](https://github.com/Fallenbagel/jellyseerr/assets/29928635/d3f46383-cf37-4eff-b1f0-f3d5b79aa88d)
![image](https://github.com/Fallenbagel/jellyseerr/assets/29928635/76407458-3393-4714-88b5-b941c2425b1e)
![image](https://github.com/Fallenbagel/jellyseerr/assets/29928635/90300809-82b0-45ef-b81f-11102e11316e)

### Platform

desktop

### Device

Windows Browser

### Operating System

Window 10 for the clientside, Unraid docker for the server side

### Browser

Brave but happens on Edge, chrome and Firefox too

### Code of Conduct

- [X] I agree to follow Overseerr's Code of Conduct

markschrik:
I have the same issue. Weird part is even users who do not have access to my 4K library, still can access it this way.

markschrik:
@zer0ish Funny thing, I noticed on accident. Someone created an issue for a movie available in HD and 4K. On the issue screen, if I click Play on Emby, it forwards to the HD version. If I click on Play in 4K on Emby, it also forwards to the HD version. This is getting confusing. Do you have this too?

zer0ish:
@markschrik I have my 4K movies library un-synced so users will go to the 1080p version. If they want the 4K ones, they can go to emby directly and select it from the 4K folder there instead of going through Jellyseer for 4K stuff.
At least until it's fixed.

Fallenbagel:
[...]

PR #805, opened 2024-06-10
fix(jellyfinscanner): assign only 4k available badge for a 4k request instead of both badges

gauthier-th:
#### Description

When you have a 4k server setup, and request a 4k item, when it becomes available it also sets the normal item as available thus not allowing the user to request for the normal item

#### To-Dos


gauthier-th:
It might be worth testing this PR to see if it solves #520.

Fallenbagel (approved):

PR #817, opened 2024-06-14
feat: remove "requested" status of unmonitored movies during scan

### LINKED ISSUE #695: Support monitor type for Radarr
### Description

Mostly an extension of #191, and related to #325.

I'd like to browse and discover content that seems interesting on Jellyseerr and if it seems worthy to watch add it to my 'backlog' but not necessarily start monitoring. Having a selection if a movie should be monitored or not be presented to at least the approving user (admin) would be a nice addition to prevent the library filling up with content that wouldn't be watched for a long time.

### Desired Behavior

Have the option as the admin user to select as which monitoring mode the movie should be added to Radarr.

### Additional Context

_No response_

### Code of Conduct

bonswouar:
#### Description

Quick & easy first implementation of Monitoring status support in Radarr:
When a movie - which was monitored before - is unmonitored, it won't appear as "requested" in Jellyseerr anymore

#### Issues Fixed or Closed

- https://github.com/Fallenbagel/jellyseerr/issues/695

## Summary by CodeRabbit

* **New Features**
  * Added "Remove Unmonitored Media" toggle in settings to automatically process movies that become unmonitored in Radarr, updating their status when enabled.

* **Bug Fixes**
  * Fixed documentation typo in scanner comments.

### DIFF
--- a/server/lib/scanners/baseScanner.ts
+++ b/server/lib/scanners/baseScanner.ts
@@ -240,7 +240,7 @@ class BaseScanner<T> {
 
   /**
    * processShow takes a TMDB ID and an array of ProcessableSeasons, which
-   * should include the total episodes a sesaon has + the total available
+   * should include the total episodes a season has + the total available
    * episodes that each season currently has. Unlike processMovie, this method
    * does not take an `is4k` option. We handle both the 4k _and_ non 4k status
    * in one method.
@@ -702,6 +702,21 @@ class BaseScanner<T> {
 
   get protectedBundleSize(): number {
     return this.bundleSize;
+  }
+
+  protected async processUnmonitoredMovie(tmdbId: number): Promise<void> {
+    const mediaRepository = getRepository(Media);
+    await this.asyncLock.dispatch(tmdbId, async () => {
+      const existing = await this.getExisting(tmdbId, MediaType.MOVIE);
+      if (existing && existing.status === MediaStatus.PROCESSING) {
+        existing.status = MediaStatus.UNKNOWN;
+        await mediaRepository.save(existing);
+        this.log(
+          `Movie TMDB ID ${tmdbId} unmonitored from Radarr. Media status set to UNKNOWN.`,
+          'info'
+        );
+      }
+    });
   }
 }
 
--- a/src/components/Settings/SettingsMain/index.tsx
+++ b/src/components/Settings/SettingsMain/index.tsx
@@ -64,6 +64,9 @@ const messages = defineMessages('components.Settings.SettingsMain', {
   validationApplicationUrlTrailingSlash: 'URL must not end in a trailing slash',
   partialRequestsEnabled: 'Allow Partial Series Requests',
   enableSpecialEpisodes: 'Allow Special Episodes Requests',
+  removeUnmonitoredEnabled: 'Remove Unmonitored Media',
+  removeUnmonitoredExplanation:
+    'Remove Movies from Jellyseerr that are not available and have been un-monitored since',
   locale: 'Display Language',
   youtubeUrl: 'YouTube URL',
   youtubeUrlTip:
@@ -175,6 +178,7 @@ const SettingsMain = () => {
             enableSpecialEpisodes: data?.enableSpecialEpisodes,
             cacheImages: data?.cacheImages,
             youtubeUrl: data?.youtubeUrl,
+            removeUnmonitoredEnabled: data?.removeUnmonitoredEnabled,
           }}
           enableReinitialize
           validationSchema={MainSettingsSchema}
@@ -195,6 +199,7 @@ const SettingsMain = () => {
                 enableSpecialEpisodes: values.enableSpecialEpisodes,
                 cacheImages: values.cacheImages,
                 youtubeUrl: values.youtubeUrl,
+                removeUnmonitoredEnabled: values.removeUnmonitoredEnabled,
               });
               mutate('/api/v1/settings/public');
               mutate('/api/v1/status');
@@ -530,6 +535,35 @@ const SettingsMain = () => {
                         setFieldValue(
                           'enableSpecialEpisodes',
                           !values.enableSpecialEpisodes
+                        );
+                      }}
+                    />
+                  </div>
+                </div>
+                <div className="form-row">
+                  <label
+                    htmlFor="removeUnmonitoredEnabled"
+                    className="checkbox-label"
+                  >
+                    <span className="mr-2">
+                      {intl.formatMessage(messages.removeUnmonitoredEnabled)}
+                    </span>
+                    <SettingsBadge badgeType="experimental" />
+                    <span className="label-tip">
+                      {intl.formatMessage(
+                        messages.removeUnmonitoredExplanation
+                      )}
+                    </span>
+                  </label>
+                  <div className="form-input-area">
+                    <Field
+                      type="checkbox"
+                      id="removeUnmonitoredEnabled"
+                      name="removeUnmonitoredEnabled"
+                      onChange={() => {
+                        setFieldValue(
+                          'removeUnmonitoredEnabled',
+                          !values.removeUnmonitoredEnabled
                         );
                       }}
                     />
--- a/cypress/config/settings.cypress.json
+++ b/cypress/config/settings.cypress.json
@@ -22,6 +22,7 @@
     "blocklistedTagsLimit": 50,
     "trustProxy": false,
     "mediaServerType": 1,
+    "removeUnmonitoredEnabled": false,
     "partialRequestsEnabled": true,
     "enableSpecialEpisodes": false,
     "locale": "en"
--- a/seerr-api.yml
+++ b/seerr-api.yml
@@ -239,6 +239,9 @@ components:
         partialRequestsEnabled:
           type: boolean
           example: false
+        removeUnmonitoredEnabled:
+          type: boolean
+          example: false
         localLogin:
           type: boolean
           example: true
--- a/server/interfaces/api/settingsInterfaces.ts
+++ b/server/interfaces/api/settingsInterfaces.ts
@@ -41,6 +41,7 @@ export interface PublicSettingsResponse {
   mediaServerType: number;
   partialRequestsEnabled: boolean;
   enableSpecialEpisodes: boolean;
+  removeUnmonitoredEnabled: boolean;
   cacheImages: boolean;
   vapidPublic: string;
   enablePushRegistration: boolean;
--- a/server/lib/scanners/radarr/index.ts
+++ b/server/lib/scanners/radarr/index.ts
@@ -79,14 +79,13 @@ class RadarrScanner
   }
 
   private async processRadarrMovie(radarrMovie: RadarrMovie): Promise<void> {
-    if (!radarrMovie.monitored && !radarrMovie.hasFile) {
-      this.log(
-        'Title is unmonitored and has not been downloaded. Skipping item.',
-        'debug',
-        {
-          title: radarrMovie.title,
-        }
-      );
+    const settings = getSettings();
+    if (
+      settings.main.removeUnmonitoredEnabled &&
+      !radarrMovie.monitored &&
+      !radarrMovie.hasFile
+    ) {
+      this.processUnmonitoredMovie(radarrMovie.tmdbId);
       return;
     }
 
--- a/server/lib/settings/index.ts
+++ b/server/lib/settings/index.ts
@@ -144,6 +144,7 @@ export interface MainSettings {
   mediaServerType: number;
   partialRequestsEnabled: boolean;
   enableSpecialEpisodes: boolean;
+  removeUnmonitoredEnabled: boolean;
   locale: string;
   youtubeUrl: string;
 }
@@ -196,6 +197,7 @@ interface FullPublicSettings extends PublicSettings {
   jellyfinServerName?: string;
   partialRequestsEnabled: boolean;
   enableSpecialEpisodes: boolean;
+  removeUnmonitoredEnabled: boolean;
   cacheImages: boolean;
   vapidPublic: string;
   enablePushRegistration: boolean;
@@ -402,6 +404,7 @@ class Settings {
         mediaServerType: MediaServerType.NOT_CONFIGURED,
         partialRequestsEnabled: true,
         enableSpecialEpisodes: false,
+        removeUnmonitoredEnabled: false,
         locale: 'en',
         youtubeUrl: '',
       },
@@ -690,6 +693,7 @@ class Settings {
       mediaServerType: this.main.mediaServerType,
       partialRequestsEnabled: this.data.main.partialRequestsEnabled,
       enableSpecialEpisodes: this.data.main.enableSpecialEpisodes,
+      removeUnmonitoredEnabled: this.data.main.removeUnmonitoredEnabled,
       cacheImages: this.data.main.cacheImages,
       vapidPublic: this.vapidPublic,
       enablePushRegistration: this.data.notifications.agents.webpush.enabled,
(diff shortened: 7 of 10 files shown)


Fallenbagel:
Converted to draft as the title states "WIP". Once it's ready feel free to make this pr ready for review

bonswouar:
@Fallenbagel Thanks I didn't think about the Draft feature

This is a WIP but it's basically working as I'd expect.
But I could update it with a more specific Monitoring status support if you think it's a good idea?

Only drawback is that if you have lots of Unmonitored movies in Radarr the scan should be slower, as it will request each unmonitored movie in db every time. Let me know if you have some thoughts about this

bonswouar:
FYI I've been using this for months and it works as expected.
I suspect some other users might need this functionality, couldn't we merge it?

_A more complete monitoring status support wouldn't be very useful imo, at least I didn't encounter any use case_

gauthier-th (commented):
I don't think this should be merged if it's only Radarr. Could you implement the same logic for Sonarr seasons?
And shouldn't there be a setting to enable/disable this behavior? You may have unmonitored a movie/tv show but still want to keep it as requested.

bonswouar:
@gauthier-th I've added the Setting, disabled by default

Also I've tried to add the same behaviour for Sonarr, but I encountered an issue:
- Some (specific) seasons are still shown as "Requested", I don't understand how as I get the new log message "Removing request for ..." for those. Plus I've added a check in `processShow()` directly, so I really don't see where that could happen in the code. If you've got any clue?

_Also, not really related to this issue afaik, but still looks like the same kind of problem:
I've noticed during my debugging that some Seasons are shown as "Available" although they don't contain any file in Sonarr (and are not monitored).
I'd guess it's related to `If the season is already marked as available, we force it to stay available (to avoid competing scanners)`, but I'm really not sure why this condition exists despite the comment (why not at least check if it has any episode instead of relying entirely on the previous status?!)_

src/i18n/locale/fr.json
gauthier-th, line 1: You should not edit these translations, only the English ones. Other translations than English are handled by our [Weblate](https://translate.jellyseerr.dev/).

src/components/Settings/SettingsMain/index.tsx
```diff
@@ -54,6 +54,7 @@ const messages = defineMessages('components.Settings.SettingsMain', {
   validationApplicationUrl: 'You must provide a valid URL',
   validationApplicationUrlTrailingSlash: 'URL must not end in a trailing slash',
   partialRequestsEnabled: 'Allow Partial Series Requests',
+  removeUnmonitoredFromRequestsEnabled: 'Remove Request for Movies/Seasons that have been un-monitored since',
```
gauthier-th: Maybe set something shorter? Like "Remove Unmonitored Media"
And then add a tip to describe what it does more precisely.

bonswouar:
I'll squash the commits when everything's approved

Meanwhile any hint on Sonarr weird behaviour is welcome! I probably missed something, don't know much about the project nor Typescript

server/lib/scanners/sonarr/index.ts
```diff
@@ -112,13 +116,50 @@ class SonarrScanner
...
+              existingRequest.seasons.forEach(async (requestedSeason) => {
+                if (requestedSeason.seasonNumber === season.seasonNumber) {
+                  this.log(
+                    `Removing request for Season ${season.seasonNumber} of ${sonarrSeries.title} as it is unmonitored`
+                  );
+                  await seasonRequestRepository.remove(requestedSeason);
+                }
+              });
+            });
+          }
```
gauthier-th: Here you remove the season request. However, you don't remove the request itself right? You should remove the request if there is no more season request in it.

gauthier-th: You should move this to a `processUnmonitoredSeason` like you did for `processUnmonitoredMovie`, to be consistent.

server/lib/scanners/baseScanner.ts
```diff
@@ -283,6 +285,10 @@ class BaseScanner<T> {
               ? MediaStatus.PARTIALLY_AVAILABLE
               : !season.is4kOverride && season.processing
               ? MediaStatus.PROCESSING
+              : settings.main.removeUnmonitoredFromRequestsEnabled &&
+                !season.monitored &&
+                season.episodes == 0
+              ? MediaStatus.UNKNOWN
```
gauthier-th: Why is this change necessary? Isn't the Sonarr scanner job enough?

gauthier-th (changes requested):
It looks like the way things are done are not consistent between Radarr and Sonarr.
Why do you change the status of the Media entity for movies while you remove the SeasonRequest for series?
It may be linked to the issue you experienced.

server/lib/scanners/sonarr/index.ts
bonswouar: It's already done with @AfterRemove in SeasonRequest entity, as far as I understand the code

bonswouar: It's already done with @AfterRemove in SeasonRequest entity, as far as I understand the code (?)

gauthier-th: Oh yes you're right, my bad.

bonswouar:
> It looks like the way things are done are not consistent between Radarr and Sonarr. Why do you change the status of the Media entity for movies while you remove the SeasonRequest for series? It may be linked to the issue you experienced.

For clarity, there are two potential "problems" with un-monitored medias that I've been trying to fix:

1. **Scenario 1**
Nothing is requested. But the movie was originally added (from Radarr directly) with the status monitored. It's been un-monitored later
    - In Jellyseerr it will always be shown as "Requested", which imo is a real issue (it really doesn't match Radarr behaviour, and the user can't know nor do anything about it, will never be able to really request it, etc)
      - => That's my first commit specifically for this problem, for Movies only _(and there is no Request in this case, despite the visible "Requested" status)_

3. **Scenario 2**
Something is requested, but manually un-monitored before it completes. _For Movies it's not really a problem imo, as the admin can as easily delete a request from Jellyseerr than un-monitor a Movie from Radarr._ But for Shows it becomes more complicated as **an admin can't EDIT a Request**
    - Thus if for example multiple Seasons of a Show are Requested, but only 1 kept & all the other ones unmonitored from Sonarr, the user won't be able to request again the other ones.
      - => That's the main goal of my second commit for Shows

~~I can certainly add 1/ for Shows for consistency, but not sure many users automatically import monitored Shows from Sonarr and un-monitor them before they're downloaded.~~ Actually I'm pretty sure that's also what [this change](https://github.com/Fallenbagel/jellyseerr/pull/817#discussion_r1837206552) does: Season not monitored && no episode = Status UNKNOWN
I hope my explanation & use cases do make sense!

bonswouar:
> There are several jobs that update the status of a Media: Radarr/Sonarr jobs, and Jellyfin/Emby/Plex jobs, hence the comment about competing jobs.

I still don't get the logic, why _"If the season is already marked as available, we force it to stay available"_? 
What if the season is not available anymore??!
The referencing code is:
```ts
(season.totalEpisodes === season.episodes && season.episodes > 0) ||
            existingSeason.status === MediaStatus.AVAILABLE
```
And I don't get why we don't check anything on this second line (i.e `existingSeason.status === MediaStatus.AVAILABLE && season.episodes > 0`)
But anyway that should be a separate issue I guess as it's not about monitor status

EDIT: Well I did try this simple change, and it seems to fix this issue for one season that was always shown as Available before (although it has been deleted some time ago)

server/lib/scanners/baseScanner.ts
bonswouar: It seems to be necessary. It **is** part of Sonarr scanner job though!
- At the beginning of the job I check if there is any SeasonRequest and delete it if necessary
- And here I check the Status of the Season, as every Season is parsed and its Status checked/updated, it should set un-monitored season with no file to UNKNOWN (as for Movies)

bonswouar:
@gauthier-th I went with your recommendations, and marked as resolved the comments I was sure about.
I'll let you have a look at my other answers

Not sure why/how, but during my last tests the problem seems fixed!
Although, after running 2 Sonarr scans, I still had:
```
Removing request for Season X of tmdbId XXXX as it is unmonitored
```
with the same Season/ID, **twice** in the log. Which means the first time didn't work, right?

Is there something I should know about async deletion in typescript or something? This seems a bit random

gauthier-th:
> I still don't get the logic, why "If the season is already marked as available, we force it to stay available"?
> What if the season is not available anymore??!

There is another job, "Media Availability Sync" that runs and mark the media as unavailable if necessary.

> Is there something I should know about async deletion in typescript or something? This seems a bit random

It could come from your `forEach` loops. For instance in the `processUnmonitoredSeason` you are using an async function inside a `forEach` loops, but `forEach` loops don't wait the end of the async function to run the next item. IMHO It's better to use a `for` loop to iterate over an array than a `forEach`, because of this async function concurrency. 

Let me review it one more time with the change you made and with your comments in mind.

server/lib/scanners/baseScanner.ts
```diff
@@ -618,6 +633,56 @@ class BaseScanner<T> {
...
+      existingRequests.forEach((existingRequest) => {
+        existingRequest.seasons.forEach(async (requestedSeason) => {
+          if (requestedSeason.seasonNumber === season.seasonNumber) {
+            this.log(
+              `Removing request for Season ${season.seasonNumber} of tmdbId ${tmdbId} as it is unmonitored`
+            );
+            await seasonRequestRepository.remove(requestedSeason);
+          }
+        });
+      });
```
gauthier-th: You should use `for` loops here instead of `forEach` if you want the `await` to be effective.

gauthier-th (changes requested):

bonswouar:
@gauthier-th I've updated with `for` loops, rebased from develop, and squashed everything into one cleaner commit.
Please let me know if some other changes are needed!

bonswouar:
> Is the issue with the log message repeated twice fixed?

After more tests... Apparently not :/
It seems `await seasonRequestRepository.remove(requestedSeason);` still has no effect (sometimes?!), i.e I have only 1 Season of 1 TV Show requested & deleted & unmonitored, and it still shows this message for **every** Sonarr scan, and this (season / show) request is **never** deleted at all.

Just for clear context, the log & remove is pretty simple:
```ts
this.log(
   `Removing request for Season ${season.seasonNumber} of tmdbId ${tmdbId} as it is unmonitored`
);
await seasonRequestRepository.remove(requestedSeason);
```

I've tried without the `await`, same issue.
My guess is that part of the problem might be because `processShow` is happening after `processUnmonitoredSeason`, with the "removed season" included ; but that seems weird that the request doesn't get removed still!

I've also tried a dirty workaround (skip adding the season to `processableSeasons`), but that didn't help either

---
_What's next could be a new Issue, but just in case it's related.._

I forgot about that, but as pointed out in [one of my previous comment](https://github.com/Fallenbagel/jellyseerr/pull/817#issuecomment-2466982799) the issue for this specific case is also that **the Season is still marked as Available**, although it's been deleted for some time now.
And if you're wondering, yes, Media Availability Sync job ran many times since this season got removed!

Could it be that Jellyseerr relies only Jellyfin's "basic season status" to check the status of a Season?
My guess would be: 
As the _(empty)_ season folder still exists, in Jellyfin the Season is still shown, but is empty. In Sonarr its status makes sense though, that's why I was expecting it to be based on Sonarr.. Or at least to check if there are any episodes available
Sounds a bit like [this issue](https://github.com/Fallenbagel/jellyseerr/issues/1126), but I have no idea what "misconfiguration" it could be (plus the "Play on Jellyfin" do link to the Season page, it just contains no episode).

I wanted to give a try with the official image to confirm I have the same issue, but got the error `no such column: User_settings.region`, that got my instance offline, apparently related to the latest commit.
I can't do much more tests right now, as the only instance I have is my "production" one

bonswouar:
I now realize that this last test _(and all other similar weird behaviour I had with deleted seasons)_ is out of scope of this PR.... It's only a Media Availability Sync issue.

Still I don't understand why remove does not remove, but it's not related to "an unmonitored tv show", only a "deleted tv show".

gauthier-th:
I may have an idea. Your `processUnmonitoredMovie` and `processUnmonitoredSeason` are not consistent.

In the [availability sync job](https://github.com/Fallenbagel/jellyseerr/blob/develop/server/lib/availabilitySync.ts), we don't really delete the media, but we set its status as `UNKNOWN`, for movies and shows. That's what you're doing in `processUnmonitoredMovie`, but not in `processUnmonitoredSeason`, you don't update the status of the seasons.

Also, you don't remove the request for the movie, but you do remove it for shows.

Try to update these 2 functions to mimic the behavior of `mediaUpdater` and `seasonUpdater` in the [availability sync job](https://github.com/Fallenbagel/jellyseerr/blob/develop/server/lib/availabilitySync.ts).

gauthier-th:
@Fallenbagel any idea why the media availability sync job mark the movies/seasons as `MediaStatus.UNKNOWN` instead of deleting the media?

bonswouar:
I understand!
Although I believe I've mixed things because of those shows that still appear as "Available" even if they're not _**anymore** (meaning, they've been deleted since!)_.. But that's a totally different issue, the availability sync job should take care or that, no matter if they're monitored or not.
I'll remove that part from my PR and will let you know if it seems to make sense afterwards, thanks for the feedbacks!

And about my "scenario 2" of [this previous comment](https://github.com/Fallenbagel/jellyseerr/pull/817#issuecomment-2469009526), I guess it's out of scope of this PR, and could be fixed / worked around by another feature request, for example admins being able to edit requests _(something like https://github.com/Fallenbagel/jellyseerr/issues/585)_

bonswouar:
@gauthier-th FYI I've removed the part about TV Shows from my commit (basically went back to my first iteration...)

I don't know how to manage unmonitored shows. Should we check for seasons only, or the whole show? Should we also set it (every episodes, or..?!) to `MediaStatus.UNKNOWN`? Etc.

As a reminder, the original bug (it was more a bug fix than a new feature imo) was that:
**Movies added** (and monitored) **in Radarr appear as "requested"** _(although there was no actual request)_, **and stay forever as requested** if not downloaded, even if unmonitored (**preventing actual users to really request them**!).

I have the issue with Radarr's Import Lists, but that could also happen if something is manually monitored on Radarr, and then unmonitored.

bonswouar:
> Movies added and monitored in radarr will not show as requested. Can you test this. We don't do bidirectional sync like this. 

Well I've just checked some random auto-imported (and monitored) movies in Radarr, I confirm it's still the case.

But that **can** make sense, specifically when they're unreleased: users shouldn't need/want to request movies that will already be automatically download I guess (at least in current Jellyseer's state; it could be useful if Movies can be requested by more than 1 user though)

bonswouar:
> If you believe so could you record and show me after adding a movie as monitored (a new movie) and then show me that exact same movie as marked as requested on jellyseerr?

So the ones from today don't appear as the sync didn't happen yet.
But from yesterday for example:

image
image

Fallenbagel:
> > If you believe so could you record and show me after adding a movie as monitored (a new movie) and then show me that exact same movie as marked as requested on jellyseerr?
> 
> So the ones added today don't appear as Requested as the sync didn't happen yet. But added yesterday for example:
> 
> image image
> 
> (But again, I think in current state of thing this is a wanted behaviour, as Monitored status isn't shown. Otherwise users will request "for nothing" all the unreleased movies that are already monitored! This could make sense only if more than 1 user is able to request a movie imo - requests could then be seen more like a "follow" movie)

Oh so it does change the status. I see. iirc it wasn't like this previously as I remember being able to request ones already added to radarr (but i guess this was changed recently and since I haven't looked into it yet did not notice it). I will have a look at it.

bonswouar:
Just a little reminder, if you have an idea for a better way of handling this let me know I could probably look into it.

Otherwise, I've been using my fork for more than a year and it works exactly as expected fyi

fallenbagel:
superseded by #2757

PR #831, opened 2024-06-23
feat: add environment variable for API key

AidanHilt:
#### Description
Adds an environment variable, API_KEY, that can be used to set the API key. If the environment variable is not set, the previous method is used.

#### To-Dos

#### Issues Fixed or Closed

This doesn't fix any filed issues, the need for this came up from my own projects.


gauthier-th:
For what use cases do you need this?

AidanHilt:
The main advantage of this setup is for people who use tools like: https://github.com/buildarr/buildarr-jellyseerr. Buildarr is a configuration management tool for the Jellyseerr and the -arr stack, and it can completely initialize Jellyseerr automatically, if provided with the API key. This makes it possible to generate that key externally, and then provide it to both Jellyseerr and a configuration management tool like Buildarr.

src/i18n/locale/en.json
```diff
@@ -986,7 +986,7 @@
   "components.Settings.plexlibraries": "Plex Libraries",
   "components.Settings.plexlibrariesDescription": "The libraries Jellyseerr scans for titles. Set up and save your Plex connection settings, then click the button below if no libraries are listed.",
   "components.Settings.plexsettings": "Plex Settings",
-  "components.Settings.plexsettingsDescription": "Configure the settings for your Plex server. Jellyseerr scans your Plex libraries to determine content availability.",
+  "components.Settings.plexsettingsDescription": "Configure the settings for your Plex server. Overseerr scans your Plex libraries to determine content availability.",
```
gauthier-th: This shouldn't be changed

AidanHilt:
I removed the change, pulled the latest changes from develop, and added documentation

gauthier-th (changes requested):
What you've done won't work if the API_KEY environment variable is changed after the 1st restart of Jellyseerr, unless you click on "regenerate" after setting it.

It seems odd to have to regenerate a key you've just defined manually, so you should add a check each time you start Jellyseerr to check whether there's an API_KEY environment variable, and if so change the main settings with this key.

AidanHilt:
Alright, I went ahead and added a check when the settings load, if the environment variable is set it will read that and use that at startup. I made sure to pull in the latest changes on develop. Documentation is still updated as well. Anything else I should adjust or any changes to be made?

gauthier-th (approved):
LGTM

Fallenbagel (approved):

PR #844, opened 2024-06-28
fix: abort media availability sync job if auth token invalid/connection lost

Fallenbagel:
#### Description
This fix aborts the media availability sync job if the following conditions are met: 
- auth token has expired
- connection to jellyfin/emby has been lost

Previously, the sync job will continue even if auth token was invalid or connection was lost, thereby, resulting in removal of series/movies that were never removed on jellyfin/emby/sonarr/radarr. This also removed the requests. With the current fix, the sync job should refuse to run unless the auth token is valid.

#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX

### DIFF
--- a/server/lib/availabilitySync.ts
+++ b/server/lib/availabilitySync.ts
@@ -73,29 +73,50 @@ class AvailabilitySync {
         });
       }
 
-      if (mediaServerType === MediaServerType.PLEX) {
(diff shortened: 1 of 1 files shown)


Fallenbagel:
Why is the lint & test build workflow not running...

PR #845, opened 2024-06-28
fix: abort availability sync job if auth token invalid/connection lost

Fallenbagel:
#### Description
This fix aborts the media availability sync job if the following conditions are met:
- auth token has expired
- connection to jellyfin/emby has been lost

Previously, the sync job will continue even if auth token was invalid or connection was lost, thereby, resulting in removal of series/movies that were never removed on jellyfin/emby/sonarr/radarr. This also removed the requests. With the current fix, the sync job should refuse to run unless the auth token is valid.

supersedes #844 

#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX

### DIFF
--- a/server/lib/availabilitySync.ts
+++ b/server/lib/availabilitySync.ts
@@ -73,29 +73,50 @@ class AvailabilitySync {
         });
       }
 
-      if (mediaServerType === MediaServerType.PLEX) {
-        if (admin && admin.plexToken) {
-          this.plexClient = new PlexAPI({ plexToken: admin.plexToken });
-        } else {
-          logger.error('Plex admin is not configured.');
-        }
-      } else if (
-        mediaServerType === MediaServerType.JELLYFIN ||
-        mediaServerType === MediaServerType.EMBY
-      ) {
-        if (admin) {
-          this.jellyfinClient = new JellyfinAPI(
-            getHostname(),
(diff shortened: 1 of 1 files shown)


Fallenbagel:
Why is the lint & test build workflow no longer running...

server/lib/availabilitySync.ts
```diff
@@ -73,29 +73,50 @@ class AvailabilitySync {
...
-        mediaServerType === MediaServerType.JELLYFIN ||
-        mediaServerType === MediaServerType.EMBY
-      ) {
-        if (admin) {
-          this.jellyfinClient = new JellyfinAPI(
-            getHostname(),
-            admin.jellyfinAuthToken,
-            admin.jellyfinDeviceId
-          );
+      switch (mediaServerType) {
```
gauthier-th, line 76: Why change the if statement with a switch?

fallenbagel: Because when i added in the try block for connection checking, it ended up getting really nested and hard to read. Should i change it back?

gauthier-th: No no it's ok, I was just wondering why did you change it

gauthier-th (approved):
LGTM, not the most elegant way of doing it but I don't see any other one.

PR #868, opened 2024-07-09
feat(jellyfinapi): switch to API tokens instead of auth tokens

gauthier-th:
#### Description

This PR changes the way Jellyseerr interacts with Jellyfin. Instead of using auth tokens from an admin user to connect to Jellyfin and perform all its requests, Jellyseerr will now use API tokens.

The API token will be automatically created from the previous auth tokens to ensure a smooth migration.

#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX

### DIFF
--- a/server/api/jellyfin.ts
+++ b/server/api/jellyfin.ts
@@ -93,9 +93,7 @@ export interface JellyfinLibraryItemExtended extends JellyfinLibraryItem {
 }
 
 class JellyfinAPI extends ExternalAPI {
-  private authToken?: string;
   private userId?: string;
-  private jellyfinHost: string;
 
   constructor(jellyfinHost: string, authToken?: string, deviceId?: string) {
     let authHeaderVal: string;
@@ -114,9 +112,6 @@ class JellyfinAPI extends ExternalAPI {
         },
       }
     );
-
-    this.jellyfinHost = jellyfinHost;
-    this.authToken = authToken;
   }
 
   public async login(
--- a/server/api/jellyfin.ts
+++ b/server/api/jellyfin.ts
@@ -405,6 +405,23 @@ class JellyfinAPI extends ExternalAPI {
       throw new ApiError(e.cause?.status, ApiErrorCode.InvalidAuthToken);
     }
   }
+
+  public async createApiToken(appName: string): Promise<string> {
+    try {
+      await this.axios.post(`/Auth/Keys?App=${appName}`);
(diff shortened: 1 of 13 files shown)


Gauvino:
Work's really great !

server/routes/user/index.ts
```diff
@@ -501,25 +501,21 @@ router.post(
...
-        ],
+        select: ['id', 'jellyfinDeviceId', 'jellyfinUserId'],
         order: { id: 'ASC' },
       });
+
+      const hostname = getHostname();
       const jellyfinClient = new JellyfinAPI(
-        getHostname(),
-        admin.jellyfinAuthToken ?? '',
+        hostname,
```
fallenbagel, line 510: Why not inline it?

gauthier-th: Because the `hostname` variable was already used later in the code ([here](https://github.com/Fallenbagel/jellyseerr/pull/868/files/850fc7516b7dff89e5b2147fbc045a8f9f158b37#diff-20f9b030a139bdbd17c8a67a791feeef0c2a22aac142d68c15777cef4e6bf792L522)), so instead of replacing every reference to it i just moved it here

fallenbagel: Ah didn't see that there. Makes sense

server/api/jellyfin.ts
```diff
@@ -93,9 +93,7 @@ export interface JellyfinLibraryItemExtended extends JellyfinLibraryItem {
 }
 
 class JellyfinAPI extends ExternalAPI {
-  private authToken?: string;
   private userId?: string;
-  private jellyfinHost: string;
 
   constructor(jellyfinHost: string, authToken?: string, deviceId?: string) {
     let authHeaderVal: string;
```
fallenbagel, line 99: (not sure how to comment on the line below this)

We could remove the client, device, deviceId, version now that authorization this is handled using api token
https://gist.github.com/nielsvanvelzen/ea047d9028f676185832e51ffaf12a6f#examples

Fallenbagel (changes requested):

server/api/jellyfin.ts
gauthier-th: Isn't this still used to display logs info in Jellyfin?

fallenbagel: Oh wait. You're right. When users login, that will use these auth headers too. My bad, didn't think about that

Fallenbagel (approved):

PR #875, opened 2024-07-14
fix: Length of undefined on users warnings

franciscofsales:
#### Description

On my instance, I was getting a frontend error where the were some issues with trying to check the length of undefined.
I tracked it down to a function that was not completely checking its dependencies and just added a more extensive check that should prevent this issue from happening again.

I didn't find anyone else with the same issue but regardless, this PR is a tiny improvement.

This PR does not change any functionality or affect any of the existing ones

image

#### To-Dos
N/A

#### Issues Fixed or Closed
Could not find anyone experiencing the same issue.

### DIFF
--- a/src/components/Layout/UserWarnings/index.tsx
+++ b/src/components/Layout/UserWarnings/index.tsx
@@ -17,47 +17,45 @@ interface UserWarningsProps {
 const UserWarnings: React.FC<UserWarningsProps> = ({ onClick }) => {
   const intl = useIntl();
   const { user } = useUser();
-  if (!user) {
+  //check if a user has warnings
+  if (!user || !user.warnings || user.warnings.length === 0) {
     return null;
   }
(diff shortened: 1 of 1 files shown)


gauthier-th:
Could you tell use how to reproduce this issue?

franciscofsales:
Sure, so I run it in the a kuberetes cluster and there's a custom ingress and error handler, this issue happens when the `/api/v1/auth/me` request does not return the expected 403 response and instead returns something else, which was my cause given that I run https://github.com/tarampampam/error-pages on my setup

gauthier-th (approved):
lgtm

gauthier-th:
https://github.com/all-contributors please add @franciscofsales for code

gauthier-th:
@all-contributors please add @franciscofsales for code

PR #881, opened 2024-07-20
fix: required valid email in user setting

### LINKED ISSUE #825: UI doesn't consider "Local Sign-In" when disabled for certain features
### Description

Some features in the UI do not consider the "Local Sign-In" settings when it is disabled.

- In the user list, the "Create Local User" button is shown when setting is disabled.
  - We should hide the button when the setting is disabled
- In the user profile settings (/users/:userId/settings), the email is required. This was added in #803 but does not check the setting. 
  - The email should not be required when the setting is disabled
- The Password tab in the user profile settings (/users/:userId/settings) is shown. The password in that tab is the password for "Local users".
  - We should hide the tab when the setting is disabled

### Version

1.9.2

### Steps to Reproduce

1. Go to the Jellyseerr setting page in the users tab (/settings/users)
2. Disable the "Local Sign-In" setting and save your changes
3. View the issues described in the description section above behave like the setting is still enabled

### Screenshots

_No response_

### Logs

_No response_

### Platform

desktop

### Device

all

### Operating System

all

### Browser

all

### Additional Context

_No response_

### Code of Conduct

Jacouille:
#### Description
Updated user email setting to be required only when the local login setting is enabled. Otherwise we don't require a valid email address. I've also added a few french i18n.

#### To-Dos

#### Issues Fixed or Closed

- Fixes #825


gauthier-th:
Hi, thanks for the PR.
Unfortunately, we had already planned to make some changes to this email setting. As you noticed, there are a few inconsistencies with this email field but also with other things related to local login and importing users from Jellyfin/Emby/Plex, and we'd like to fix all of these at the same time.

Jacouille:
Sounds good. Is there a place I can follow the progress of said changes?

gauthier-th:
I just opened #900. Feel free to take a look and add a review.
PS: for translations, it happens on our weblate: https://jellyseerr.borgcube.de/projects/jellyseerr/jellyseerr-frontend/

Jacouille:
Thanks for the update! The changes look good to me!
I'll take a look at the translation website

Issue #183, opened 2022-07-09
Support login with OIDC


michaelhthomas:
### Description

Authentication providers like Authelia, Auth0, Authentik, Keycloak, etc. make managing user permissions across selfhosted services much easier. 

### Desired Behavior

I would like to be able to configure a custom OIDC provider for logins, and have this option appear on the Jellyseer login page. [This PR](https://github.com/sct/overseerr/pull/1941) from Overseerr implements support for generic OIDC login. Since OIDC is already supported for Plex SSO, supporting other external authentication services is fairly straightforward.

### Additional Context

Here is a screencast from the linked Overseerr PR showing what this feature would look like when implemented.

https://cln.sh/wGbPIdYUKas2QId0cqxw

### Code of Conduct

- [X] I agree to follow Overseerr's Code of Conduct

michaelhthomas:
I'd be happy to try my hand at re-implementing this for Jellyseerr, so long as all are onboard.

Fallenbagel:
> I'd be happy to try my hand at re-implementing this for Jellyseerr, so long as all are onboard.

Any help would be much appreciated c:

michaelhthomas:
@Fallenbagel Would you say it's alright to go ahead and remove the Plex OAuth login stuff in here as well? It's intertwined with the authentication logic a little more than I'd like.

Fallenbagel:
> @Fallenbagel Would you say it's alright to go ahead and remove the Plex OAuth login stuff in here as well? It's intertwined with the authentication logic a little more than I'd like.

If the plex OAuth is removed, wouldn't that remove support for plex?

michaelhthomas:
Login with Plex, yes. If the goal is to keep that support in place, then I can definitely leave it.

Fallenbagel:
> Login with Plex, yes. If the goal is to keep that support in place, then I can definitely leave it. I could also see if it's possible to just detangle the Plex login a bit so that it's isolated like the rest of the auth providers.

Yeah the goal is for plex to be supported as well

michaelhthomas:
Got it. I'll try separating plex login into its own component so all the login logic isn't on the main `Login` component

ados8:
Testing with Authentik and it fails the OIDC login due to exceeded threshold.
![image](https://user-images.githubusercontent.com/49982433/194679826-06aa8dcc-e6b0-4ede-b318-9b10e094717d.png)

Jellyseerr presents an error to sign in.
[...]

PR #883, opened 2024-07-21
feat: add linked accounts page

michaelhthomas:
#### Description

Adds a "Linked Accounts" page to the user settings, allowing managing of linked media server (Plex, Jellyfin) accounts, including linking an account to your existing local user, or removing a linked account to become a local user. This page doesn't add much new functionality (except the ability to unlink accounts), but is intended to lay the groundwork for a new and improved implementation of #183 which allows the secure linking of OIDC accounts with existing local accounts.

|  Plex Linked Accounts Page  |  Link Account Dropdown  |
|-----------------------------|---------------------------|
|Plex Linked Accounts page|Dropdown menu showing Plex linking option|

|  Jellyfin Linked Accounts Page  |  Link Jellyfin Modal  |
|--------------------------------|----------------------|
|Jellyfin Linked Accounts page|Link Jellyfin Modal|

|  Emby Linked Accounts Page  |  Link Emby Modal  |
|-------------------------------|--------------------|
|Emby Linked Accounts Page|Link Emby Modal|

#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX


michaelhthomas:
Some outstanding questions / TODOs:

- Should it be possible to link Plex accounts with a different email address? Right now I'm leaning towards no for simplicity's sake, since otherwise we would have to change how the auth linking works as well.
- Need to make sure you can't lock yourself out of your account. Particularly, need to make sure that users with a media server account do not unlink it without first setting a password
- Emby? I haven't tested it at all and it currently doesn't have a logo, so will probably just show the jellyfin one
- Will need help testing this, particularly with Plex since I don't personally use it

server/entity/User.ts
gauthier-th, line 1: Why these modifications of the strings of the User table?

src/components/Common/Dropdown/index.tsx
gauthier-th, line 1: Don't we already have a Dropdown component?

server/entity/User.ts
gauthier-th: And the modifications of the type to add `| null` when it is already an optional field?

gauthier-th (commented):
Just a few comments for now.
And you will add Emby once the PR with Jellyfin/Emby as different server types will be merged right?
Otherwise it's great I like it, good job!

server/entity/User.ts
michaelhthomas: Great question! TypeORM treats null differently than undefined, and you have to set values to null explicitly during update operations to clear the set value, whereas undefined does not change the value. Unfortunately this resulted in some annoying changes to other files to deal with the possible null values (in the type, they aren't actually nullable in practice)

michaelhthomas:
> And you will add Emby once the PR with Jellyfin/Emby as different server types will be merged right?

Yep, that's the plan currently. Hopefully that PR will make emby a lot easier to integrate.

src/components/Common/Dropdown/index.tsx
michaelhthomas: Yeah there's `ButtonWithDropdown`. Unfortunately, that one is more like a button group, with a primary button and then a secondary one to trigger the dropdown. This component is basically the same thing but just has a single button which triggers the dropdown. I could probably update them to use the same components for the actual dropdown part though if you think that would be worthwhile.

server/entity/User.ts
gauthier-th: Good to know! But why leave the field as optional with `?` then?
For instance here: `field?: string | null`, the field type will be inferred as `string | null | undefined`, which will also be annoying right? Especially if TypeORM makes sure this field always exist and is null without a value.

src/components/Common/Dropdown/index.tsx
gauthier-th: I just want to make sure that we don't create a generic component to be used only once.
We use a simple HTML selection in the Discover Movie page filter, for example. Couldn't you use the same thing since you only use text in your dropdown?
https://github.com/Fallenbagel/jellyseerr/blob/develop/src/components/Discover/DiscoverMovies/index.tsx#L80-L116

server/entity/User.ts
michaelhthomas: It's not too terribly annoying, since checking `field != null` will coerce `undefined` as well. Additionally, that field is actually undefined in a number of places, such as when creating or updating a user in the database with a partial record.

src/components/Common/Dropdown/index.tsx
michaelhthomas: The HTML select is a little different from a dropdown, since the select actually lets you chose an active option, instead of just showing a list of buttons. I'll definitely go ahead and unify the `Dropdown` and `ButtonWithDropdown` components, though, so there isn't any unnecessary duplication.

michaelhthomas: An additional benefit of unifying them will be that the `ButtonWithDropdown` component will be using the `headless-ui` dropdown (instead of the current custom implementation) which should help improve the accessibility of that component.

michaelhthomas:
Unified the two dropdown components and addressed a few other issues with the frontend.

michaelhthomas:
Rebased on `feat-server-type-setup` and added support for emby

Fallenbagel:
> Rebased on `feat-server-type-setup` and added support for emby

That pr has now been merged

Gauvino:
Why this PR is still in draft any reason for this @michaelhthomas ?

gauthier-th:
Hi @michaelhthomas, could you please rebase this on develop and make it ready for review if everything is ok?

TheRedCyclops:
Is there a preview tag for testing on docker?

gauthier-th:
> Is there a preview tag for testing on docker?

I just created one: `preview-linked-accounts` (will be available in ~30min)

overseerr-api.yml
```diff
@@ -4343,6 +4343,100 @@ paths:
...
+        - in: path
+          name: userId
+          required: true
+          schema:
+            type: number
+      responses:
+        '204':
+          description: Unlinking account succeeded
+        '404':
+          description: User does not exist
```
gauthier-th: I think a 400 error is missing here

overseerr-api.yml
```diff
@@ -4343,6 +4343,100 @@ paths:
...
```
gauthier-th: I think a 400 error is missing here

server/routes/user/usersettings.ts
```diff
@@ -267,6 +285,265 @@ userSettingsRoutes.post<
...
+            'Cannot unlink media server accounts for the primary administrator.',
+        });
+      }
+
+      const hasPassword = !!(
+        await userRepository.findOne({
+          where: { id: user.id },
+          select: ['id', 'password'],
+        })
+      )?.password;
```
gauthier-th: This is not really clear to read.
And why another call to the database just to check the password? Can't you select the password in the previous call that is retrieving the user?

server/routes/user/usersettings.ts
```diff
@@ -267,6 +285,265 @@ userSettingsRoutes.post<
...
+    } catch (e) {
+      logger.error('Failed to link account to user.', {
+        label: 'API',
+        ip: req.ip,
+        error: e,
+      });
+      if (
+        e instanceof ApiError &&
+        (e.errorCode == ApiErrorCode.InvalidCredentials ||
+          e.errorCode == ApiErrorCode.NotAdmin)
```
gauthier-th: This `NotAdmin` API error code is never thrown here

server/routes/user/usersettings.ts
```diff
@@ -267,6 +285,265 @@ userSettingsRoutes.post<
...
+            'Cannot unlink media server accounts for the primary administrator.',
+        });
+      }
+
+      const hasPassword = !!(
+        await userRepository.findOne({
+          where: { id: user.id },
+          select: ['id', 'password'],
+        })
+      )?.password;
```
gauthier-th: Same as before

src/types/error.ts
gauthier-th, line 1: I don't think you should implement one more way to handle errors.
We currently have `ApiErrorCode`, but I don't really like it as it is very verbose to handle it from the client. We also have a lot of other places where error handling is inconsistent.
I'll talk about it with @Fallenbagel but I think we have to do something about this.

src/components/Common/ButtonWithDropdown/index.tsx
gauthier-th, line 1: Have you checked that the refactor of this ButtonWithDropdown component doesn't affect anything on the existing UI?

src/components/Common/Modal/index.tsx
```diff
@@ -29,11 +29,16 @@ interface ModalProps {
   secondaryDisabled?: boolean;
   tertiaryDisabled?: boolean;
   tertiaryButtonType?: ButtonType;
+  okButtonProps?: React.ButtonHTMLAttributes<HTMLButtonElement>;
+  cancelButtonProps?: React.ButtonHTMLAttributes<HTMLButtonElement>;
+  secondaryButtonProps?: React.ButtonHTMLAttributes<HTMLButtonElement>;
+  tertiaryButtonProps?: React.ButtonHTMLAttributes<HTMLButtonElement>;
```
gauthier-th, line 35: Looks like these 3 properties are not used

src/components/UserProfile/UserSettings/UserLinkedAccountsSettings/LinkJellyfinModal.tsx
```diff
@@ -0,0 +1,187 @@
...
+              okButtonProps={{ type: 'submit', form: 'link-jellyfin-account' }}
+              okText={
+                isSubmitting
+                  ? intl.formatMessage(messages.saving)
+                  : intl.formatMessage(messages.save)
+              }
+              okDisabled={isSubmitting || !isValid}
+              onOk={() => handleSubmit()}
+              title={intl.formatMessage(messages.title, { mediaServerName })}
+              dialogClass="sm:max-w-lg"
```
gauthier-th, line 135: Genuine question (I did not check how it looks like): why is it the only place in Jellyseerr where we could need to have a smaller dialog? It goes against consistency?

src/components/UserProfile/UserSettings/UserLinkedAccountsSettings/index.tsx
```diff
@@ -0,0 +1,273 @@
...
+        username: user.plexUsername,
+      });
+    if (user.userType === UserType.EMBY && user.jellyfinUsername)
+      accounts.push({
+        type: LinkedAccountType.Emby,
+        username: user.jellyfinUsername,
+      });
+    if (user.userType === UserType.JELLYFIN && user.jellyfinUsername)
+      accounts.push({
+        type: LinkedAccountType.Emby,
```
gauthier-th: Typo?

src/components/UserProfile/UserSettings/UserLinkedAccountsSettings/index.tsx
```diff
@@ -0,0 +1,273 @@
...
+import { useIntl } from 'react-intl';
+import useSWR from 'swr';
+import LinkJellyfinModal from './LinkJellyfinModal';
+
+const messages = defineMessages(
+  'components.UserProfile.UserSettings.UserLinkedAccountsSettings',
+  {
+    linkedAccounts: 'Linked Accounts',
+    linkedAccountsHint:
+      'These external accounts are linked to your Jellyseerr account.',
```
gauthier-th: You are using `applicationName` in `LinkJellyfinModal.tsx` but not here. Why?

src/components/Common/Dropdown/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+  switch (buttonType) {
+    case 'ghost':
+      styleClass +=
+        ' bg-transparent rounded hover:bg-gradient-to-br from-indigo-600 to-purple-600 text-white focus:border-gray-500 focus:text-white';
+      break;
+    default:
+      styleClass +=
+        ' bg-indigo-600 rounded hover:bg-indigo-500 focus:border-indigo-700 focus:text-white';
+  }
```
gauthier-th: `rounded` class is on both button types.
It feels a bit unusual to write class names this way. Why not write this with a ternary inside the className?

src/components/Common/Dropdown/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+  let dropdownClasses: string;
+
+  switch (dropdownType) {
+    case 'ghost':
+      dropdownClasses =
+        'bg-gray-800 border border-gray-700 bg-opacity-80 p-1 backdrop-blur';
+      break;
+    default:
+      dropdownClasses = 'bg-indigo-600 p-1';
+  }
```
gauthier-th: Same a above

src/components/Common/Dropdown/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+  switch (buttonType) {
+    case 'ghost':
+      dropdownButtonClasses +=
+        ' bg-transparent border-gray-600 hover:border-gray-200 focus:border-gray-100 active:border-gray-100';
+      break;
+    default:
+      dropdownButtonClasses +=
+        ' bg-indigo-600 border-indigo-500 bg-opacity-80 hover:bg-opacity-100 hover:border-indigo-500 active:bg-indigo-700 active:border-indigo-700 focus:ring-blue';
+  }
```
gauthier-th: Same as above

gauthier-th (commented):
Here are a few more remarks after a detailed review. I still have to test it though.
Feel free to challenge me if you disagree on something.

src/types/error.ts
fallenbagel: I agree. We need to come up with a better way. The thing i implemented wasn't good either

src/components/UserProfile/UserSettings/UserLinkedAccountsSettings/index.tsx
gauthier-th: image
Yep this is wrong, my account is a Jellyfin account.

src/components/Common/Dropdown/index.tsx
michaelhthomas: I wrote this so long ago that I honestly don't remember what the reasoning was. I'm inclined to believe there were probably more variants at some point? Or maybe this was copied from somewhere. In any case, this could definitely be more succinct, and I'll look into it.

src/components/UserProfile/UserSettings/UserLinkedAccountsSettings/index.tsx
michaelhthomas: Good point, I'll update this to use `applicationName` as well

server/routes/user/usersettings.ts
michaelhthomas: ~~This was since the `password` field has `select: false` set, and as far as I can tell there's no easy way to select all of the default fields + the password. If there's a better way to do this, I would be happy to do that instead.~~ I found a better way.

src/components/Common/ButtonWithDropdown/index.tsx
michaelhthomas: I have done a cursory check, but I'll look through more thoroughly to be sure.

src/types/error.ts
michaelhthomas: I'll look into using the `ApiError` type on the server and handling errors that way, it should be good enough for this use case.

michaelhthomas:
@gauthier-th Updated to address feedback, thanks for reviewing!

src/components/UserProfile/UserSettings/UserLinkedAccountsSettings/LinkJellyfinModal.tsx
michaelhthomas: I don't really think this hurts consistency, and honestly I think there are some other dialogs in the app that could benefit from a smaller size. The current default size is optimized for the content-dense settings modals, but things like confirmation dialogs (and this modal) look really silly since they're so wide.

If you'd prefer, I can implement this as a size prop with variants within the modal component (like `size: 'sm' | 'md' | 'lg'`), so that there's a standard for different size modals.

gauthier-th: That's fine like this, i was just curious about it.

gauthier-th (approved):
LGTM.
A few more tests are welcome though.

gauthier-th:
@michaelhthomas I think this will be merged soon. Could you please rebase on develop and fix the conflicts? :pray:

michaelhthomas:
@gauthier-th Happy New Year! I've updated this PR and it should be ready to review/merge.

Fallenbagel:
Let me create a preview of this as well

gauthier-th (approved):
LGTM
Thanks you very much for your contribution!

Gauvino:
chrome_JURQ6N9gw5
Just an error message from node but I think the problem come from the latest pr switching to node22
Over-all, looking great, no bug found, thank you really for all the work !

BoBeR182:
@fallenbagel any timeline set for when this will merge to main?

fallenbagel:
> @fallenbagel any timeline set for when this will merge to main?

Once it is reviewed by me.

fallenbagel (approved):
LGTM

gauthier-th (approved):

fallenbagel:
@all-contributors please add @michaelhthomas for code

PR #896, opened 2024-07-26
fix: rewrite the rate limit utility

gauthier-th:
#### Description

This PR fixes the rateLimit function that was allowing X parallel requests instead of limiting to X requests per second.

#### To-Dos


Gauvino (commented):
Finally fixed let's go, now work really great

Fallenbagel (approved):

PR #900, opened 2024-07-26
fix: remove email requirement for the user, and use the username if no email provided

gauthier-th:
#### Description

This PR remove the email requirement for users, but only from the user's perspective. From server side, Jellyfin/Plex username will be used when no email is provided, since a lot of Jellyseerr features need this email field.

#### To-Dos


src/components/UserProfile/UserSettings/UserGeneralSettings/index.tsx
```diff
@@ -93,9 +93,14 @@ const UserGeneralSettings = () => {
...
 
   const UserGeneralSettingsSchema = Yup.object().shape({
-    email: Yup.string()
-      .email(intl.formatMessage(messages.validationemailformat))
-      .required(intl.formatMessage(messages.validationemailrequired)),
+    email:
+      user?.id === 1
+        ? Yup.string()
+            .email(intl.formatMessage(messages.validationemailformat))
+            .required(intl.formatMessage(messages.validationemailrequired))
```
Jacouille, line 100: I'd recommend swapping these two as we'd like to show the required message when the field is empty. Currently, the email validation will be displayed when the field is empty

server/routes/auth.ts
```diff
@@ -320,15 +320,18 @@ authRoutes.post('/jellyfin', async (req, res, next) => {
...
+        email: body.email || account.User.Name,
         jellyfinUsername: account.User.Name,
         jellyfinUserId: account.User.Id,
         jellyfinDeviceId: deviceId,
         jellyfinAuthToken: account.AccessToken,
         permissions: Permission.ADMIN,
         avatar: account.User.PrimaryImageTag
           ? `${jellyfinHost}/Users/${account.User.Id}/Images/Primary/?tag=${account.User.PrimaryImageTag}&quality=90`
-          : gravatarUrl(body.email ?? '', { default: 'mm', size: 200 }),
+          : gravatarUrl(body.email || account.User.Name, {
```
fallenbagel, line 331: Why was this added? Afaik gravatar only uses emails

server/routes/auth.ts
```diff
@@ -371,7 +374,7 @@ authRoutes.post('/jellyfin', async (req, res, next) => {
       if (account.User.PrimaryImageTag) {
         user.avatar = `${jellyfinHost}/Users/${account.User.Id}/Images/Primary/?tag=${account.User.PrimaryImageTag}&quality=90`;
       } else {
-        user.avatar = gravatarUrl(user.email, {
+        user.avatar = gravatarUrl(user.email || account.User.Name, {
```
fallenbagel, line 377: See above

server/routes/auth.ts
```diff
@@ -426,7 +425,10 @@ authRoutes.post('/jellyfin', async (req, res, next) => {
         permissions: settings.main.defaultPermissions,
         avatar: account.User.PrimaryImageTag
           ? `${jellyfinHost}/Users/${account.User.Id}/Images/Primary/?tag=${account.User.PrimaryImageTag}&quality=90`
-          : gravatarUrl(body.email, { default: 'mm', size: 200 }),
+          : gravatarUrl(body.email || account.User.Name, {
```
fallenbagel, line 428: See above

server/routes/user/index.ts
```diff
@@ -108,7 +121,7 @@ router.post(
       }
 
       const passedExplicitPassword = body.password && body.password.length > 0;
-      const avatar = gravatarUrl(body.email, { default: 'mm', size: 200 });
+      const avatar = gravatarUrl(email, { default: 'mm', size: 200 });
```
fallenbagel, line 124: Here too

Fallenbagel (changes requested):

server/routes/auth.ts
gauthier-th: Yep, but the `gravatar-url` package throws an error if the provided string is empty, but send a default profile picture when the provided a string/email doesn't have a corresponding profile picture.
I could replace the `|| account.User.Name` by something like `|| 'none'` if you thing it makes more sense.

fallenbagel: Email will always be there though. And if it's not there an empty string would give the default image (atleast it used to. Does it not now?)

gauthier-th: Email should be here yes. I added this because during one of my test I came across an empty email and `gravatar-url` throwed an error because of it.
See https://github.com/sindresorhus/gravatar-url/blob/main/index.js#L4-L6

fallenbagel: Ah makes sense. I assumed since overseerr was also just passing in email only

Fallenbagel (approved):

PR #907, opened 2024-08-01
refactor: Proxy and cache avatar images

JoaquinOlivero:
#### Description
This PR changes how Jellyseerr handles profile images. Now they are proxied from the source and cached in the disk.
The caching mechanism uses the already existing "ImageProxy" class which I tweaked.

Also, I added a new endpoint "/avatarproxy" that takes the URL from the images, caches them if needed, and then returns them.

#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX


server/lib/imageproxy.ts
```diff
@@ -27,29 +27,37 @@ class ImageProxy {
...
+                recursive: true,
+              });
+              deletedImages += 1;
+            }
           }
         }
       }
+    } catch (e) {
+      if (e.code === 'ENOENT') {
+        logger.error(e.message);
```
gauthier-th: Could you please add a more explicit error message?

server/lib/imageproxy.ts
```diff
@@ -187,16 +211,25 @@ class ImageProxy {
...
+            : '/'
+          : '') +
         (path.startsWith('/') ? path.slice(1) : path);
       const response = await this.fetch(href);
       const arrayBuffer = await response.arrayBuffer();
       const buffer = Buffer.from(arrayBuffer);
 
-      const extension = path.split('.').pop() ?? '';
-      const maxAge = Number(
+      const extension = (response.headers.get('content-type') ?? '').replace(
```
gauthier-th: What is the purpose of modifying the extension to content-type?
Won't this create issues with the `writeToCacheDir` function that uses it for the filename?

src/components/UserProfile/ProfileHeader/index.tsx
```diff
@@ -1,9 +1,10 @@
 import Button from '@app/components/Common/Button';
+import CachedImage from '@app/components/Common/CachedImage';
 import type { User } from '@app/hooks/useUser';
 import { Permission, useUser } from '@app/hooks/useUser';
 import defineMessages from '@app/utils/defineMessages';
 import { CogIcon, UserIcon } from '@heroicons/react/24/solid';
-import Image from 'next/image';
+// import Image from 'next/image';
```
gauthier-th: You can remove this

gauthier-th (changes requested):
Well done!

server/lib/imageproxy.ts
```diff
@@ -171,6 +171,21 @@ class ImageProxy {
...
+    const cacheKey = this.getCacheKey(path);
+
+    try {
+      const directory = join(this.getCacheDirectory(), cacheKey);
+      await promises.rm(directory, { recursive: true });
+      logger.info(`Cleared ${path} image from cache 'avatar'`, {
+        label: 'Image Cache',
+      });
+    } catch (e) {
+      logger.error(e.message, { label: 'Image Cache' });
```
gauthier-th: Same here, more explicit error message

gauthier-th (changes requested):

server/lib/imageproxy.ts
JoaquinOlivero: I had to do it this way because the URL of the profile pictures from Jellyfin and Emby don't have a file extension, causing issues. However, I recognize that getting the image file extension from the content-type header needs to be handled better, because by just using .replace('/image', '') can cause further issues.

server/lib/imageproxy.ts
```diff
@@ -315,6 +329,31 @@ class ImageProxy {
...
+      case 'image/tiff':
+        return 'tiff';
+      case 'image/webp':
+        return 'webp';
+      default:
+        break;
+    }
+
+    return '';
+  }
```
gauthier-th: I don't have a strong opinion on this, but shouldn't we use a library to do this conversion (like [this one](https://github.com/broofa/mime?tab=readme-ov-file#quick-start))?
I feel like this is an easy way to mess with the [official specification](https://www.iana.org/assignments/media-types/media-types.xhtml#image).

@Fallenbagel any opinion on this?

fallenbagel: I agree

JoaquinOlivero: My bad, because I 100% agree. I was going to ask whether using that library for this situation was acceptable and forgot to do it and left it as it was.

gauthier-th (dismissed):

Gauvino:
No problem at all work's great !

src/i18n/locale/en.json
```diff
@@ -820,6 +820,7 @@
   "components.Settings.SettingsJobsCache.runnow": "Run Now",
   "components.Settings.SettingsJobsCache.sonarr-scan": "Sonarr Scan",
   "components.Settings.SettingsJobsCache.unknownJob": "Unknown Job",
+  "components.Settings.SettingsJobsCache.useravatars": "User Avatars",
```
Gauvino:
```suggestion
  "components.Settings.SettingsJobsCache.useravatars": "User Avatars",
```
This doesn't seems good to me, maybe "User(s) Avatars or "User Avatar"

gauthier-th (dismissed):

gauthier-th (changes requested):
Could you try to replace `.startsWith('https://gravatar.com')` with `.startsWith('https://gravatar.com/')` (note the `/` at the end) to see if it fixes the GitHub security warnings?

src/components/Common/CachedImage/index.tsx
```diff
@@ -16,8 +16,15 @@ const CachedImage = ({ src, ...props }: ImageProps) => {
   if (typeof imageUrl === 'string' && imageUrl.startsWith('http')) {
     const parsedUrl = new URL(imageUrl);
 
-    if (parsedUrl.host === 'image.tmdb.org' && currentSettings.cacheImages) {
-      imageUrl = imageUrl.replace('https://image.tmdb.org', '/imageproxy');
+    console.log(parsedUrl);
```
gauthier-th: I think you forgot to remove that

src/components/Common/CachedImage/index.tsx
```diff
@@ -16,8 +16,15 @@ const CachedImage = ({ src, ...props }: ImageProps) => {
...
+
+    if (parsedUrl.host === 'image.tmdb.org') {
+      if (currentSettings.cacheImages)
+        imageUrl = imageUrl.replace('https://image.tmdb.org', '/imageproxy');
+    }
+
+    if (parsedUrl.protocol === 'http:') {
```
gauthier-th: Why only http?
Shouldn't this be an `else` from the `if (parsedUrl.host === 'image.tmdb.org') { ... }`?
IF it's tmdb image THEN cache image but only if image caching enabled ELSE cache image

gauthier-th (approved):
LGTM!

Fallenbagel (approved):
Have not tested yet but LGTM

Gauvino:
Could the action run for a docker image be done please ?

gauthier-th:
> Could the action run for a docker image be done please ?

Done with the `preview-cache-avatar` tag

Fallenbagel (approved):

Fallenbagel (changes requested):
Some issues were reported (on issues index) in discord so until that is fixed

package.json
```diff
@@ -62,6 +62,7 @@
     "formik": "^2.4.6",
     "gravatar-url": "3.1.0",
     "lodash": "4.17.21",
+    "mime": "3",
```
Gauvino, line 65: You couldn't make mime: 4 to work ?

JoaquinOlivero: No because it's an es module and Jellyseerr uses cjs

Fallenbagel (approved):
Everything looks good!

gauthier-th (approved):

PR #932, opened 2024-08-16
refactor: replace streaming providers' names with their equivalent logos

### LINKED ISSUE #919: "Currently streaming on" to show the streamers' logos
### Description

It would be great if "currently streaming on" section, could show the streamers' logos instead of just a text-based name.

This would make it much more apparent to users when content is available on other streaming services - and in turn, hopefully prevent them from requesting content may not necessarily be required.

Thank you

### Desired Behavior

Show logos of the "Currently Streaming On" service providers, instead of listing them in text. See attached screenshot example.

image

### Additional Context

Not sure if this helps - but if TMDB is the source used, I understand from their support forums that `logo_path` can be used in conjunction with  `get/watch/providers/tv` api to return the logos.

### Code of Conduct

JoaquinOlivero:
#### Description
This PR changes the name of the different streaming services with their corresponding logos. The TMDB API provides the image URLs for the logos. 
Also, a tooltip with the provider's name is shown when hovering over the logo.

image
image

#### To-Dos

#### Issues Fixed or Closed

- Fixes #919


gauthier-th (changes requested):
To be consistent with other external link images below, could you apply a hover transition to the images so that the opacity is set to 1 when they are hovered?

gauthier-th (approved):
LGTM

PR #975, opened 2024-09-22
feat: add Content-Security-Policy Header, allows setting frame-ancestor domains and moved font local

### LINKED ISSUE #455: Authentification error Jellyseerr iFrame in Jellyfin
### Description

Hello everyone, 

I have my jellyfin server and my jellyseerr server accessible via a different dynamic dns (noIP) in https behind an Apache reverse proxy.

I've integrated jellyseerr as an iframe instead of the trailer tab.

I can see the login page but I can't connect, error 401 unauthorized, I've tried lots of solutions without success. When I access https://JELLYSEERR.ddns.net directly, I can connect without any problem. I was able to figure out that the problem was due to cookies and probably to the sharing of cookies between domains without resolving the problem.

I've tested the connection via API with `X-Api-Key`, everything works but the connection is auto on the Owner account, not usable as I'm sharing the server with my family/friends. I haven't managed to configure the connection on an account other than the owner.

I also have a configuration with the same config that works perfectly with Ombi in iframe on jellyfin in https. 

Do you have any ideas on how to enable authentication? Or (perhaps the best) keep the API connection but be able to choose auto connection to another account (without admin rights)? 
J'ai essayé énormément de choses sans succès je commence à sécher 

Thanks

### Version

1.6.0

### Steps to Reproduce

Authentification in iframe

### Screenshots

_No response_

### Logs

_No response_

### Platform

desktop

### Device

PC

### Operating System

Debian

### Browser

Chrome, Safari, Edge, Firefox

### Additional Context

_No response_

### Code of Conduct

Ruakij:
#### Description

I have added Content-Security-Policy Header in a very moderate mode not to break any current features.
Through Content-Security-Policy we can allow specific origins or "Frame-Ancestors" to embed jellyseer as e.g. iframe.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors

Currently, correct usage in an iframe is blocked by strict or lax setting of SameSite on the session- and, if enabled, the  csrf-cookie, which results in the cookie not getting send in a request which originates from a cross-origin.
https://github.com/Fallenbagel/jellyseerr/blob/edfd80444c3d8b9d82998f8a7bbb0b3496701602/server/index.ts#L163-L175

This setting is incompatible with CSRF-Protection as its not possible to have the protection without SameSite strict.

I am unsure if we should complexity remove the SameSite attribute or leave it with the condition when Frame-Ancestor Domains is unset like i have done currently.

As my setup of Content-Security-Policy also blocked the google-fonts request, i simply moved it to local with the additional benefit of speed and privacy.
If you rather want to, we can remove most of the Content-Security-Policy not to hinder future features, but its a good thing for security usually to block anything we dont expect like loading external javascript.

The setting is a space-separated string of domains.
Screenshot-Settings
When the browser blocks the request, it looks something like this:
Screenshot-iframe-blocked
And here when having added the domain:
Screenshot-iframe-works

#### To-Dos

#### Issues Fixed or Closed

- Fixes #455


ThomasVicot:
Hi, 
I tried to install this version via the to do,
pnpm build
pnpm i18n:extract
pnpm start 
In ‘General Settings’ I don't have the new feature ‘Frame-Ancestor Domains’.
Have I forgotten a step?

jellyseerr

Ruakij:
@ThomasVicot I have checked, the setting doesnt require a migration, if the field doesnt appear, you did something wrong.
All you should have to do is checkout the branch and `pnpm dev` for the development build.

Fallenbagel:
> Hi, 
> I tried to install this version via the to do,
> pnpm build
> pnpm i18n:extract
> pnpm start 
> In ‘General Settings’ I don't have the new feature ‘Frame-Ancestor Domains’.
> Have I forgotten a step?
> 
> jellyseerr
> 

The todo is not the steps to checkout prs. Thats for the PR authors todos...

to checkout prs you do:
http://stackoverflow.com/questions/27567846/ddg#30584951

src/components/Settings/SettingsMain/index.tsx
```diff
@@ -44,6 +44,9 @@ const messages = defineMessages('components.Settings.SettingsMain', {
   csrfProtectionTip: 'Set external API access to read-only (requires HTTPS)',
   csrfProtectionHoverTip:
     'Do NOT enable this setting unless you understand what you are doing!',
+  cspFrameAncestorDomains: 'Frame-Ancestor Domains',
+  cspFrameAncestorDomainsTip:
+    'Domains to allow embedding Jellyseer as iframe, object or embed. Incompatible with CSRF-Protection',
```
gauthier-th, line 49: Shouldn't you specify here that this is a space-separated list of domains?

src/styles/globals.css
```diff
@@ -53,6 +53,10 @@
 }
 
 @layer components {
+  *:disabled {
+    opacity: 0.7;
+  }
+
```
gauthier-th, line 59: Why this change?

src/pages/_app.tsx
```diff
@@ -18,11 +18,13 @@ import { MediaServerType } from '@server/constants/server';
 import type { PublicSettingsResponse } from '@server/interfaces/api/settingsInterfaces';
 import type { AppInitialProps, AppProps } from 'next/app';
 import App from 'next/app';
+import { Inter } from 'next/font/google';
 import Head from 'next/head';
 import { useEffect, useState } from 'react';
 import { IntlProvider } from 'react-intl';
 import { ToastProvider } from 'react-toast-notifications';
 import { SWRConfig } from 'swr';
+const inter = Inter({ subsets: ['latin'] });
```
gauthier-th, line 27: Maybe you could add a newline after the imports?

gauthier-th (commented):
> I am unsure if we should complexity remove the SameSite attribute or leave it with the condition when Frame-Ancestor Domains is unset like i have done currently.

IMHO I think it's ok the way you've done it. @M0NsTeRRR may have a more stronger opinion on the subject.

src/components/Settings/SettingsMain/index.tsx
Ruakij: There are more possibilities to setting this than just space separated as this is directly put into the header.
But maybe we could link to the specification: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors

src/styles/globals.css
Ruakij: Before, the text-box wasnt visibly disabled apart from not being able to interact with it, so this makes it a bit more visible

M0NsTeRRR:
> As my setup of Content-Security-Policy also blocked the google-fonts request, i simply moved it to local with the additional benefit of speed and privacy.

Fully agree

> If you rather want to, we can remove most of the Content-Security-Policy not to hinder future features, but its a good thing for security usually to block anything we dont expect like loading external javascript.

Fully agree

> This setting is incompatible with CSRF-Protection as its not possible to have the protection without SameSite strict.

Not exactly. CSRF protection shouldn't affect the CSP header, but it does in this case because it isn't configured properly. CSRF protection is meant to prevent logged-in users from unknowingly executing actions (such as via a JavaScript fetch when they visit a webpage). Ideally, CSRF protection should only apply to form pages and shouldn't impact iframes, as iframes are used only for displaying data not form. If we don't want to disable CSRF for CSP (we should do that), we need to fix the CSRF system in this way.

src/components/Settings/SettingsMain/index.tsx
gauthier-th: Oh ok. That's a good idea yes

Ruakij:
> If we don't want to disable CSRF for CSP (we should do that), we need to fix the CSRF system in this way.

Yeah i worded that too broadly, CSP in general is fine, but CSP ancestor-domains/iframes in combination with SameSite is incompatible.
SameSite is just very restrictive and doesnt allow any domain exceptions.

So the CSRF cookie would simply need SameSite none.
We should probably explicitly set SameSite none for all cookies going forward as we should be able to rely on CSP ancestor-domains preventing clickjacking (which could afaik. otherwise bypass CSRF protected forms with iframes).

So i guess we can change https://github.com/Fallenbagel/jellyseerr/blob/257cab387791ae4f9836091cd8daddb76b3a42dd/server/index.ts#L157-L174 `sameSite: true` to `'none'`?

Another thing should probably then be to be able to bypass CSRF protection with API-keys to have read/write API, but protected for user-actions, but thats for the future.

src/styles/globals.css
gauthier-th: Did you check that this does not have any other unexpected behavior? It could have undesired effects on some text fields. and other fields too*

We usually do styling using TailwindCSS, inside the `class` attribute of the elements

M0NsTeRRR:
> So the CSRF cookie would simply need SameSite none.

Jellyseerr is a stateless application that uses [signed double submit cookies](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#signed-double-submit-cookie-recommended) to prevent CSRF attacks. According to the [proposed standard](https://www.ietf.org/archive/id/draft-ietf-httpbis-rfc6265bis-15.html#name-strict-and-lax-enforcement), we should set the `SameSite` attribute to `strict` or `lax`, though `lax` might not cover all of our cases. [This mitigation](https://www.ietf.org/archive/id/draft-ietf-httpbis-rfc6265bis-15.html#top-level-navigations) is suggested for page that need authentication, but it’s not as simple as turning off `SameSite` (which we want to avoid, as it would disable CSRF protection). If we want to iframe unauthenticated pages, we can disable CSRF for those pages, as it's not necessary.

> Another thing should probably then be to be able to bypass CSRF protection with API-keys to have read/write API, but protected for user-actions, but thats for the future.

Yes

Ruakij:
Yeah i havent thought this through fully.
If we want the application to be useable (at least for user-actions like requesting), we cannot use Cookie-based CSRF protection which relies on SameSite as on other mode than none, it doesnt work through iframes.
iframe navigations arent top-level navigations.

So maybe using the signed-token approach via header or somewhere an attacker cannot obtain it and/or get the browser to submit it via Cross-Origin.

I think for now we can implement this PRs iframe-support with disabling the current cookie-based CSRF-protection and then gradually move towards a more compatible system?

M0NsTeRRR:
> If we want the application to be useable (at least for user-actions like requesting), we cannot use Cookie-based CSRF protection
As I mentioned before, CSRF protection should be disabled for non-action requests. By 'action,' I mean any operation that updates the application, like POST, PATCH, or DELETE requests. GET requests don't require CSRF protection, which would allow us to use iframes without encountering this issue.

> So maybe using the signed-token approach via header or somewhere an attacker cannot obtain it and/or get the browser to submit it via Cross-Origin.
There is no API to read initial page HTTP response header.

> I think for now we can implement this PRs iframe-support with disabling the current cookie-based CSRF-protection and then gradually move towards a more compatible system?
That’s up to the maintainer, but in my opinion, I disagree. Adding a 'feature' (iframe) that disables security, potentially allowing unauthorized admin access via an HTTP POST request on the create user endpoint, could lead to serious issues. I previously submitted a CVE for network equipment that had this vulnerability (about 20 lines of code) and was rated as a 8.8 CVE.

Fallenbagel:
> That’s up to the maintainer, but in my opinion, I disagree. Adding a 'feature' (iframe) that disables security, potentially allowing unauthorized admin access via an HTTP POST request on the create user endpoint, could lead to serious issues. I previously submitted a CVE for network equipment that had this vulnerability (about 20 lines of code) and was rated as a 8.8 CVE.

As the maintainer, I would like to avoid security vulnerabilities/CVEs as much as possible, so I'm with @M0NsTeRRR on this.

OdinVex:
I would've liked to have seen something about the fonts being made local, any particular reason they're still remote?

gauthier-th:
> I would've liked to have seen something about the fonts being made local, any particular reason they're still remote?

No particular reason. You can submit a PR.

OdinVex:
> > I would've liked to have seen something about the fonts being made local, any particular reason they're still remote?
> 
> No particular reason. You can submit a PR.

Are the fonts referenced licensed to prevent packaging?

gauthier-th:
> > > I would've liked to have seen something about the fonts being made local, any particular reason they're still remote?
> > 
> > 
> > No particular reason. You can submit a PR.
> 
> Are the fonts referenced licensed to prevent packaging?

I don't think so: https://fonts.google.com/specimen/Inter/license

OdinVex:
It looks as though including the license (attributing to the font, so there's no misunderstanding of what license it is under) is required for OFL licenses. I've never had to add a font to a project so I'm not sure how best to do so.

gauthier-th:
> It looks as though including the license (attributing to the font, so there's no misunderstanding of what license it is under) is required for OFL licenses. I've never had to add a font to a project so I'm not sure how best to do so.

IMO you could create a `fonts/` folder inside `src/assets/` with 2 files for the font and for its license

OdinVex:
> > It looks as though including the license (attributing to the font, so there's no misunderstanding of what license it is under) is required for OFL licenses. I've never had to add a font to a project so I'm not sure how best to do so.
> 
> IMO you could create a `fonts/` folder inside `src/assets/` with 2 files for the font and for its license

I've no idea if it needs to be mentioned somewhere in the UI, but that does seem to cover. I think @fallenbagel or whoever makes decisions regarding this fork could approve?

gauthier-th:
> I think fallenbagel or whoever makes decisions regarding this fork could approve?

I'm a maintainer, and yes I do approve :wink:

The license clearly states that it's ok to provide the license as a standalone file:

> Original or Modified Versions of the Font Software may be bundled, redistributed and/or sold with any software, provided that each copy contains the above copyright notice and this license. These can be included either as stand-alone text files, human-readable headers or in the appropriate machine-readable metadata fields within text or binary files as long as those fields can be easily viewed by the user.

OdinVex:
I meant I don't know whether the license had to be viewable from the UI or not (just included for the distribution of the server). Sounds good then. :)

Issue #890, opened 2024-07-24
[Feature Request] Split "Discover Region" and "Currently Streaming on"


DrMcCoy:
### Description

Currently, the "Currently Streaming On" information seems to be based on the Discover Region setting. I.e. when I set it to Germany, it shows where it's streaming in Germany. When it's set to All Regions, it shows...US? Production country? That's not clear, but it's not all services in all countries, that's for sure.

I'd like to see these concepts split, so that I can see where something is streaming in Germany (if any), but not filter away things in the Discover section.

### Desired Behavior

I'd like two different settings: "Discover Region" and, perhaps, "Streaming Region". The former is solely for filtering in the Discover section, while the latter changes which region the "Currently Streaming On" information is based off of.

### Code of Conduct

- [X] I agree to follow Jellyseerr's Code of Conduct

DrMcCoy:
Oh, nice, thanks for implementing this! :)

Just saw it referenced in the release announcement mail. Going to look at how it looks in Jellyseerr tomorrow.

(For some reason I didn't get any update mails for this ticket, though, even though I'm the author and still subscribed to this ticket? That's weird.)

gauthier-th:
> (For some reason I didn't get any update mails for this ticket, though, even though I'm the author and still subscribed to this ticket? That's weird.)

Probably because no message was sent here. Just references to this issue.

DrMcCoy:
Okay, having looked at the new version of Jellyseerr, this is exactly what I wanted. Thanks! :)

PR #993, opened 2024-10-02
feat(usersettings): add separate setting for streaming region

gauthier-th:
#### Description

Currently, the "Currently Streaming On" information is based on the Discover Region setting. This PR adds a new setting to specify which region should be used to display the streaming region.

New setting in general settings as well as in user settings:
image

#### To-Dos

#### Issues Fixed or Closed

- Re #890


server/lib/settings/migrations/0004_migrate_region_setting.ts
```diff
@@ -0,0 +1,17 @@
+import type { AllSettings } from '@server/lib/settings';
+
+const migrateHostname = (settings: any): AllSettings => {
```
fallenbagel: Why is this called `migrateHostname`?

Fallenbagel (changes requested):

server/lib/settings/migrations/0004_migrate_region_setting.ts
gauthier-th: Copy/paste error, let me fix it

Fallenbagel (approved):
LGTM

PR #1016, opened 2024-10-16
fix: rewrite avatarproxy and CachedImage

### LINKED ISSUE #1012: /avatarproxy prefer hostname/ip address over external url if provided
### Description

The user avatar in Jellyfin is retrieved from the path `/avatarproxy/jellyfin.domain.com`

### Desired Behavior

The user avatar can be retrieved from the hostname/IP address using the following path: `/avatarproxy/http://hostname:port`

### Additional Context

Tested the path using `/avatarproxy/http://jellyfin:8096/...` for Jellyfin user avatar and it works. This was not an issue in v1.19.2

### Code of Conduct

### LINKED ISSUE #1013: Avatar Proxy allows requesting any resource from literally every origin
### Description

I just noticed that the avatar proxy allows proxying/loading literally any request - no matter the original ressource's origin or filetype. 

### Version

2.0.0

### Steps to Reproduce

Try requesting (e. g.): 
https://{YOUR-JELLYSEER-HOST}/avatarproxy/https://ash-speed.hetzner.com/100MB.bin

### Screenshots

_No response_

### Logs

_No response_

### Platform

desktop

### Device

MacBook Pro (doesn't matter tho)

### Operating System

macOS 15.0.1

### Browser

Zen Browser (Firefox)

### Additional Context

The avatar proxy should only allow loading ressources from the configured Jellyfin host (or it's external URL) and maybe also check MIME types

### Code of Conduct

### LINKED ISSUE #1017: [Bug] Wrong user picture in request list
### Description

Jellyseerr get the wrong picture for user in the request list

### Version

2.0.0

### Steps to Reproduce

1. Goto Requests
2. Look to the user changes

### Screenshots

image

### Logs

_No response_

### Platform

desktop

### Device

PC

### Operating System

Windows 11

### Browser

Chrome

### Additional Context

_No response_

### Code of Conduct

gauthier-th:
#### Description

Avatar proxy was allowing every request to be proxied, no matter the original ressource's origin or filetype. This PR fixes it be allowing only relevant resources to be cached, i.e. Jellyfin/Emby images and TMDB images.

#### To-Dos

#### Issues Fixed or Closed

- Fixes #1012
- Fixes #1013
- Fixes #1017

### DIFF
--- a/server/routes/avatarproxy.ts
+++ b/server/routes/avatarproxy.ts
@@ -11,8 +11,8 @@ router.get('/*', async (req, res) => {
   let imagePath = '';
   try {
     const jellyfinAvatar = req.url.match(
-      /(.*?)(\/Users\/\w+\/Images\/Primary\/\?tag=\w+&quality=90)$/
-    )?.[2];
+      /.*?(\/Users\/\w+\/Images\/Primary\/?\?tag=\w+&quality=90)$/
+    )?.[1];
     if (!jellyfinAvatar) {
       throw new Error('Provided URL is not a Jellyfin avatar.');
     }
--- a/server/routes/avatarproxy.ts
+++ b/server/routes/avatarproxy.ts
@@ -1,4 +1,6 @@
+import { MediaServerType } from '@server/constants/server';
 import ImageProxy from '@server/lib/imageproxy';
+import { getSettings } from '@server/lib/settings';
 import logger from '@server/logger';
 import { getHostname } from '@server/utils/getHostname';
 import { Router } from 'express';
@@ -14,7 +16,14 @@ router.get('/*', async (req, res) => {
       /(\/Users\/\w+\/Images\/Primary\/?\?tag=\w+&quality=90)$/
     )?.[1];
     if (!jellyfinAvatar) {
-      throw new Error('Provided URL is not a Jellyfin avatar.');
+      const mediaServerType = getSettings().main.mediaServerType;
+      throw new Error(
+        `Provided URL is not ${
+          mediaServerType === MediaServerType.JELLYFIN
+            ? 'a Jellyfin'
+            : 'an Emby'
+        } avatar.`
+      );
     }
 
     const imageUrl = new URL(jellyfinAvatar, getHostname());
--- a/server/routes/avatarproxy.ts
+++ b/server/routes/avatarproxy.ts
@@ -1,5 +1,6 @@
 import ImageProxy from '@server/lib/imageproxy';
 import logger from '@server/logger';
+import { getHostname } from '@server/utils/getHostname';
 import { Router } from 'express';
 
 const router = Router();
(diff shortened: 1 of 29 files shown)


server/routes/avatarproxy.ts
```diff
@@ -1,15 +1,25 @@
...
 router.get('/*', async (req, res) => {
-  const imagePath = req.url.startsWith('/') ? req.url.slice(1) : req.url;
-
+  let imagePath = '';
   try {
+    const jellyfinAvatar = req.url.match(
+      /(\/Users\/\w+\/Images\/Primary\/?\?tag=\w+&quality=90)$/
+    )?.[1];
+    if (!jellyfinAvatar) {
+      throw new Error('Provided URL is not a Jellyfin avatar.');
```
fallenbagel: what would you say towards getting the `mediaServerType` here when the error is being thrown? It would be less confusing to the users?

gauthier-th: Yep

server/routes/avatarproxy.ts
```diff
@@ -1,15 +1,25 @@
...
 const router = Router();
 
 const avatarImageProxy = new ImageProxy('avatar', '');
 // Proxy avatar images
 router.get('/*', async (req, res) => {
-  const imagePath = req.url.startsWith('/') ? req.url.slice(1) : req.url;
-
+  let imagePath = '';
   try {
+    const jellyfinAvatar = req.url.match(
```
M0NsTeRRR, line 15: This regex should be set on router part and not inside the route no ?

And if you want to keep it here the regex look like wrong as it's missing `^`, would be catched : https://test.fake/Users/efefe/Images/Primary/?tag=efefe&quality=90 (even if it's useless)

src/components/Common/CachedImage/index.tsx
```diff
@@ -4,24 +4,34 @@ import Image from 'next/image';
...
+  if (type === 'tmdb') {
+    // tmdb stuff
+    imageUrl =
+      currentSettings.cacheImages && !src.startsWith('/')
+        ? src.replace('https://image.tmdb.org', '/imageproxy')
+        : src;
+  } else if (type === 'avatar') {
+    // jellyfin avatar (in any)
+    const jellyfinAvatar = src.match(
+      /(\/Users\/\w+\/Images\/Primary\/?\?tag=\w+&quality=90)$/
```
M0NsTeRRR, line 30: same issue for the regex above missing `^`

src/components/Common/CachedImage/index.tsx
```diff
@@ -4,24 +4,34 @@ import Image from 'next/image';
...
-      if (currentSettings.cacheImages)
-        imageUrl = imageUrl.replace('https://image.tmdb.org', '/imageproxy');
-    } else if (parsedUrl.host !== 'gravatar.com') {
-      imageUrl = '/avatarproxy/' + imageUrl;
-    }
+  if (type === 'tmdb') {
+    // tmdb stuff
+    imageUrl =
+      currentSettings.cacheImages && !src.startsWith('/')
+        ? src.replace('https://image.tmdb.org', '/imageproxy')
```
M0NsTeRRR: src.replace doesn't check if this string is present at the beginning so `https://fake.test/https://image.tmdb.org?pic=mypic` would be converted to `https://fake.test/imageproxy?pic=mypic`

server/routes/avatarproxy.ts
gauthier-th: > This regex should be set on router part and not inside the route no ?

It could be set in the router path yes. But I did it this way to be consistent with what is made on front-end. Both ways seems fine to me.
And having it in the router path would require to reconstruct the URL with the URL param and the URL query

src/components/Common/CachedImage/index.tsx
gauthier-th: Here this is the intended behavior.
Avatars were previously stored like this: `https://domain.jellyfin.com/possible-subpath/Users/XXX/Images/Primary?tag=XXX&quality=90`. This PR is removing the `https://domain.jellyfin.com/possible-subpath` part, but until every users log in, matching strictly the new syntax will break their avatars (avatars are updated at login)

src/components/Common/CachedImage/index.tsx
gauthier-th: Fixed

Fallenbagel (approved):
Lgtm

Fallenbagel:
Test. Seems to be working as intended

PR #1036, opened 2024-10-19
feat: add more logs to migrations and create a settings backup

gauthier-th:
#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX

### DIFF
--- a/server/lib/settings/migrator.ts
+++ b/server/lib/settings/migrator.ts
@@ -13,11 +13,12 @@ export const runMigrations = async (
   let migrated = settings;
 
   try {
-    // let's create a backup of the settings
-    fs.writeFileSync(
-      SETTINGS_PATH.replace('.json', '.old.json'),
-      JSON.stringify(settings, undefined, ' ')
-    );
+    // we read old backup and create a backup of currents settings
+    const BACKUP_PATH = SETTINGS_PATH.replace('.json', '.old.json');
+    const oldBackup = fs.existsSync(BACKUP_PATH)
+      ? fs.readFileSync(BACKUP_PATH)
+      : null;
+    fs.writeFileSync(BACKUP_PATH, JSON.stringify(settings, undefined, ' '));
 
     const migrations = fs
       .readdirSync(migrationsDir)
@@ -58,6 +59,10 @@ export const runMigrations = async (
         // something went wrong while saving file
         throw new Error('Unable to save settings after migration.');
       }
+    } else if (oldBackup) {
+      // no migration occured
+      // we save the old backup (to avoid settings.json and settings.old.json being the same)
+      fs.writeFileSync(BACKUP_PATH, oldBackup.toString());
     }
   } catch (e) {
     logger.error(
--- a/server/lib/settings/migrator.ts
+++ b/server/lib/settings/migrator.ts
@@ -1,7 +1,7 @@
 /* eslint-disable no-console */
 import type { AllSettings } from '@server/lib/settings';
 import logger from '@server/logger';
-import fs from 'fs';
+import fs from 'fs/promises';
 import path from 'path';
 
 const migrationsDir = path.join(__dirname, 'migrations');
@@ -15,14 +15,16 @@ export const runMigrations = async (
   try {
     // we read old backup and create a backup of currents settings
     const BACKUP_PATH = SETTINGS_PATH.replace('.json', '.old.json');
-    const oldBackup = fs.existsSync(BACKUP_PATH)
-      ? fs.readFileSync(BACKUP_PATH)
-      : null;
-    fs.writeFileSync(BACKUP_PATH, JSON.stringify(settings, undefined, ' '));
+    const backupExists = await fs
+      .access(BACKUP_PATH, fs.constants.F_OK)
+      .then(() => true)
+      .catch(() => false);
+    const oldBackup = backupExists ? await fs.readFile(BACKUP_PATH) : null;
+    await fs.writeFile(BACKUP_PATH, JSON.stringify(settings, undefined, ' '));
 
-    const migrations = fs
-      .readdirSync(migrationsDir)
-      .filter((file) => file.endsWith('.js') || file.endsWith('.ts'));
+    const migrations = (await fs.readdir(migrationsDir)).filter(
+      (file) => file.endsWith('.js') || file.endsWith('.ts')
+    );
 
     const settingsBefore = JSON.stringify(migrated);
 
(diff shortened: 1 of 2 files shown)


Fallenbagel (approved):

server/lib/settings/migrator.ts
```diff
@@ -10,19 +10,42 @@ export const runMigrations = async (
...
-
   let migrated = settings;
 
   try {
+    // we read old backup and create a backup of currents settings
+    const BACKUP_PATH = SETTINGS_PATH.replace('.json', '.old.json');
+    const oldBackup = fs.existsSync(BACKUP_PATH)
+      ? fs.readFileSync(BACKUP_PATH)
+      : null;
+    fs.writeFileSync(BACKUP_PATH, JSON.stringify(settings, undefined, ' '));
```
fallenbagel: I think we should use `fs.promises` to avoid any potential race conditions.

So something like:
```suggestion
await fs.promises.writeFile(BACKUP_PATH, JSON.stringify(settings, undefined, ' '));
```

server/lib/settings/migrator.ts
```diff
@@ -10,19 +10,42 @@ export const runMigrations = async (
...
+
+    const migrations = fs
+      .readdirSync(migrationsDir)
+      .filter((file) => file.endsWith('.js') || file.endsWith('.ts'));
+
```
fallenbagel: Same thing here as above. `fs.promises`

```suggestion
const migrations = await fs.promises.readdir(migrationsDir);
const validMigrations = migrations.filter(
      (file) => file.endsWith('.js') || file.endsWith('.ts')
    );
```

server/lib/settings/migrator.ts
```diff
@@ -10,19 +10,42 @@ export const runMigrations = async (
...
     const settingsBefore = JSON.stringify(migrated);
 
     for (const migration of migrations) {
```
fallenbagel, line 32: If applying the above suggestion then this will be:
```suggestion
for (const migration of validMigrations) {
```

server/lib/settings/migrator.ts
```diff
@@ -10,19 +10,42 @@ export const runMigrations = async (
...
-      migrated = await migration(migrated);
+      try {
+        logger.debug(`Checking migration '${migration}'...`, {
+          label: 'Settings Migrator',
+        });
+        // eslint-disable-next-line @typescript-eslint/no-var-requires
+        const fn = require(path.join(migrationsDir, migration)).default;
+        const newSettings = await fn(migrated);
```
fallenbagel: What do you think of using dynamic imports here? Wouldn't it be better for ts compatibility:
```suggestion
        const { default: migrationFn } = await import(path.join(migrationsDir, migration));
        const newSettings = await migrationFn(migrated);
```

server/lib/settings/migrator.ts
```diff
@@ -10,19 +10,42 @@ export const runMigrations = async (
...
+        if (JSON.stringify(migrated) !== JSON.stringify(migrated)) {
```
fallenbagel: shouldn't this be:
```suggestion
if (JSON.stringify(migrated) !== JSON.stringify(newSettings)) {
```
I guess I am a bit confused as to why you're comparing it with itself

server/lib/settings/migrator.ts
```diff
@@ -36,6 +59,10 @@ export const runMigrations = async (
         // something went wrong while saving file
         throw new Error('Unable to save settings after migration.');
       }
+    } else if (oldBackup) {
```
fallenbagel, line 69: We dont need this else if using `fs.promises`

server/lib/settings/migrator.ts
```diff
@@ -36,6 +59,10 @@ export const runMigrations = async (
...
+      // no migration occured
+      // we save the old backup (to avoid settings.json and settings.old.json being the same)
+      fs.writeFileSync(BACKUP_PATH, oldBackup.toString());
```
fallenbagel: Once again, `fs.promises`.
```suggestion
await fs.promises.writeFile(SETTINGS_PATH, JSON.stringify(migrated, null, 2));
```

Would it be better to parse this json with type safety?
```ts
const fileSaved = JSON.parse(await fs.promises.readFile(SETTINGS_PATH, 'utf-8')) as AllSettings;
```

Fallenbagel (changes requested):

Fallenbagel:
We can ignore the broken cypress test. That is because trending api is broken and it's currently showing people instead of media

server/lib/settings/migrator.ts
gauthier-th: Yep that would be better

server/lib/settings/migrator.ts
gauthier-th: Yes we do? Why shouldn't we need this?

The point of this `else if` is to save the old backup. Because if no migration occurred the `settings.old.json` will then have the same content as the `settings.json` (because we saved current `settings.json` in `settings.old.json` it at the beginning). We don't want that, we want the `settings.old.json` to keep the value of the previous migration.

server/lib/settings/migrator.ts
gauthier-th: Yep idk how it end up like this

server/lib/settings/migrator.ts
gauthier-th: We could but I don't really see the point of this since this variable is only used once to be compared after being stringified.

server/lib/settings/migrator.ts
```diff
@@ -10,32 +10,65 @@
...
+      );
+      const fileSaved = JSON.parse(await fs.readFile(SETTINGS_PATH, 'utf-8'));
       if (JSON.stringify(fileSaved) !== settingsAfter) {
         // something went wrong while saving file
         throw new Error('Unable to save settings after migration.');
       }
+    } else if (oldBackup) {
+      // no migration occured
+      // we save the old backup (to avoid settings.json and settings.old.json being the same)
+      await fs.writeFile(BACKUP_PATH, oldBackup.toString());
```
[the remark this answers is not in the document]
gauthier-th: Irrelevant. "The file may have changed since was checked": that's the point of saving it again here.

gauthier-th:
Fixed some of your comments.

server/lib/settings/migrator.ts
fallenbagel: Oh yeah I misunderstood. Ignore this comment

Fallenbagel (approved):

PR #1045, opened 2024-10-21
fix: cache Jellyfin/Emby avatars from API

gauthier-th:
#### Description

Previously, avatars were cached using image links from Jellyfin/Emby. Now, avatar images are obtained directly from the API to avoid some configuration bugs.

#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX

### DIFF
--- a/server/lib/imageproxy.ts
+++ b/server/lib/imageproxy.ts
@@ -135,6 +135,7 @@ class ImageProxy {
   private cacheVersion;
   private key;
   private baseUrl;
+  private headers: HeadersInit | null = null;
 
   constructor(
     key: string,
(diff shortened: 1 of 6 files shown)


Fallenbagel (approved):

Panda260:
Hey since that PR is merged my avatar is no longer availible 
image

Fallenbagel:
> Hey since that PR is merged my avatar is no longer availible 
> image
> 

Did you log out and back in? And why are you on develop tag

Panda260:
relogin helped
But it would also be good if you didn't have to log in again for the profile picture to be reloaded.

Fallenbagel:
> relogin helped
> But it would also be good if you didn't have to log in again for the profile picture to be reloaded.

That's how our system is designed because we don't want to be hitting jellyfin again and again to update avatar. So it only updates when you login.

PR #1062, opened 2024-10-31
fix(externalapi): extract basic auth and pass it through header

### LINKED ISSUE #1027: Cannot pass BasicAuth in URLs
### Description

When trying to connect a service, jellyseerr no longer accepts basic auth credentials as part of the URL string.

### Version

2.0.1

### Steps to Reproduce

Try to connect any service eg Sonarr to Jellyseerr. Pass basic auth as part of the string and observe the below log-line after failing to connect.

### Screenshots

_No response_

### Logs

```shell
[error][Radarr]: [Radarr] Failed to retrieve profiles: Request cannot be constructed from a URL that includes credentials: https://username:Password@urlhere/radarr/api/v3/qualityProfile?apikey=apiKeyHere
```

### Platform

desktop

### Device

Server

### Operating System

Ubuntu

### Browser

Firefox

### Additional Context

As per @Fallenbagel on discord. "2.0.0 introduced fetch and url constructor doesn't like basic auth in url."

### Code of Conduct

Fallenbagel:
#### Description
This PR adds extraction of basic authentication credentials from the URL and then pass the credentials as the `Authorization` header. And then credentials are removed from the URL before being passed to fetch. This is done because fetch request cannot be constructed using a URL with credentials.

I have not tested whether the basic authentication works though. Will create a preview called `:preview-fix-basic-auth`

#### To-Dos

- [ ] Test if it works

#### Issues Fixed or Closed

- Fixes #1027

### DIFF
--- a/server/api/externalapi.ts
+++ b/server/api/externalapi.ts
@@ -32,13 +32,27 @@ class ExternalAPI {
       this.fetch = fetch;
     }
 
-    this.baseUrl = baseUrl;
-    this.params = params;
+    const url = new URL(baseUrl);
+
     this.defaultHeaders = {
       'Content-Type': 'application/json',
       Accept: 'application/json',
+      ...((url.username || url.password) && {
+        Authorization: `Basic ${Buffer.from(
+          `${url.username}:${url.password}`
+        ).toString('base64')}`,
+      }),
       ...options.headers,
     };
+
+    if (url.username || url.password) {
+      url.username = '';
+      url.password = '';
+      baseUrl = url.toString();
+    }
+
+    this.baseUrl = baseUrl;
+    this.params = params;
     this.cache = options.nodeCache;
   }
 


server/api/externalapi.ts
```diff
@@ -32,13 +32,27 @@ class ExternalAPI {
...
+    const url = new URL(baseUrl);
+
     this.defaultHeaders = {
       'Content-Type': 'application/json',
       Accept: 'application/json',
+      ...((url.username || url.password) && {
+        Authorization: `Basic ${Buffer.from(
+          `${url.username}:${url.password}`
+        ).toString('base64')}`,
+      }),
```
gauthier-th, line 44: To be more clear, WDYT of something like this instead?
```suggestion
          Authorization:
            url.username || url.password
              ? `Basic ${Buffer.from(
                  `${url.username}:${url.password}`
                ).toString('base64')}`
              : undefined,
```

fallenbagel: Yeah I guess it is more concise and clear 🤔
Should I change it to this?

gauthier-th: IMO that's better, but it's a small thing

Fallenbagel:
Smh. Never use github to apply suggestions. Let me fix the formatting

server/api/externalapi.ts
```diff
@@ -32,13 +32,28 @@ class ExternalAPI {
...
+
     this.defaultHeaders = {
       'Content-Type': 'application/json',
       Accept: 'application/json',
+      Authorization:
+        url.username || url.password
+          ? `Basic ${Buffer.from(`${url.username}:${url.password}`).toString(
+              'base64'
+            )}`
+          : undefined,
```
fallenbagel:
```suggestion
  ...(url.username || url.password) && {
    Authorization: `Basic ${Buffer.from(`${url.username || ''}:${url.password || ''}`).toString('base64')}`
  },
```
Wouldn't it be better to pass the authorisation header conditionally as opposed to passing an empty string (undefined won't work here).

It doesn't include an empty Authorization header when there are no credentials. It's more semantically correct as there are no Authorization header at all vs an empty one. And it's cleaner in terms of HTTP headers being sent.

@gauthier-th

gauthier-th (approved):

PR #1066, opened 2024-11-01
Autofix: Anime Series Type setting not applied correctly when sending request to Sonarr

lattaai1:
I have identified the issue in the `addSeries` method of the `SonarrAPI` class. The `seriesType` parameter was not being properly set when adding a new series to Sonarr. I've modified the code to use the `options.seriesType` value when creating a new series, which should resolve the issue of series always being set as 'Anime' regardless of the selected option in Jellyseerr. 
> [!CAUTION]  
> Disclaimer: This fix was created by Latta AI and you should never merge before you check the correctness of generated code!
> 
> The fix provided by Latta AI might not be complete and it can serve as an inspiration.**

---
This bug was fixed for free by Latta AI - https://latta.ai/ourmission

If you no longer want Latta AI to attempt fixing issues on your repository, you can block this account.


gauthier-th:
no

PR #1067, opened 2024-11-01
fix(server/settings): write settings to a temp file then move

Zariel:
When writing the settings.json file ensure that the file is fully written by writing it to temporary file before renaming it to the final settings path. This should prevent issues where the config gets lost due to the file being corrupted.

#### Description
When running in K8S both Jellyseerr and Overseerr have suffered from losing the config due to pod restarts, this change makes it so that the settings.json is first wirtten to a temporary file then it is renamed to its final name. This should help prevent cases where a partial or unflushed write cause the loss of the whole settings.

#### To-Dos

### DIFF
--- a/server/lib/settings/index.ts
+++ b/server/lib/settings/index.ts
@@ -682,10 +682,9 @@ class Settings {
   }
 
   public async save(): Promise<void> {
-    await fs.writeFile(
-      SETTINGS_PATH,
-      JSON.stringify(this.data, undefined, ' ')
-    );
+    const tmp = SETTINGS_PATH + '.tmp';
+    await fs.writeFile(tmp, JSON.stringify(this.data, undefined, ' '));
+    await fs.rename(tmp, SETTINGS_PATH);
   }
 }
 


gauthier-th (approved):
LGTM, even if I wouldn't have thought that saving the settings would cause any problems though, because the settings aren't updated very often.
@Fallenbagel wdyt about this?

Zariel:
> LGTM, even if I wouldn't have thought that saving the settings would cause any problems though, because the settings aren't updated very often. @Fallenbagel wdyt about this?

I've just moved to Jellyseerr from Overseerr and didn't do a find references but Overseerr was saying settings very frequently. Running in K8s with ceph-block storage I would regularly be losing the settings to the point where I started trying to figure out what to do about it. The only things I can think of is the process being killed mid write or something happening where the kernel buffer doesnt get flushed, ive not been able to nail down the exact sequence of events but I thought would prevent users from being left with corrupted settings which ends up requiring a fresh setup.

gauthier-th:
@all-contributors please add @Zariel for code

PR #1073, opened 2024-11-03
chore: migrate to docker compose v2

M0NsTeRRR:
#### Description

Migrate to docker compose v2.
I've updated the documentation and the filename as docker compose v2 looks for an other filename by [default](https://docs.docker.com/compose/intro/compose-application-model/#the-compose-file).

User would need to run compose v2 (v1 is deprecated since 2021) and should read [this](https://docs.docker.com/compose/releases/migrate/#using-compose-v2-with-docker-in-docker) before upgrading.


gauthier-th (approved):
lgtm

Fallenbagel (approved):

PR #1084, opened 2024-11-11
fix(blacklist): request data only when modal is shown, remove useless ratelimit and lazy load blacklist

gauthier-th:
#### Description

For admin users, the button to blacklist a media (used on every media card) was displaying a Modal, that was requesting data BEFORE the modal was displayed. This resulted in dozens of additional requests everytime media cards were displayed.

A useless ratelimit has also been removed on the blacklist endpoint.

Joins between the Media table and the Blacklist table will be lazy loaded instead of being eager loaded.

#### To-Dos

### DIFF
--- a/overseerr-api.yml
+++ b/overseerr-api.yml
@@ -4142,6 +4142,21 @@ paths:
         '412':
           description: Item has already been blacklisted
   /blacklist/{tmdbId}:
+    get:
+      summary: Get media from blacklist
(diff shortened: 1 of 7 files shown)


Fallenbagel:
Finally. Two weeks of agony ended

Fallenbagel (approved):
LGTM

PR #1052, opened 2024-10-23
docs: added missing sub_filters in nginx subpath

C4J3:
Fix posters not showing up when using sub-folder (at least with NPM)

#### Description

Fixes title card images not appearing if using subfolders.
#### Screenshot (if UI-related)

Before:
image

After:
image

#### Aditional Notes

I know this is not officially supported but I figured this was a byproduct of the recent update and it just wasn't thought about. Small change but for idiots like me it'll save 10-100~ mins of trying to troubleshoot 😅.

PR #1093, opened 2024-11-13
Update reverse-proxy.mdx

riptide981:
add avatarproxy sub_filter for better visualization


riptide981:
Duplicated with #1052

PR #1096, opened 2024-11-13
fix(usersettings): allow unset email and add more explicit email error message

### LINKED ISSUE #1088: Warning for unset E-mail
### Description

Is it normal to create a new local user and the email cannot be changed or added once it is set or not set during the new user creation process?

> If the email is not set during the user creation process, a warning that the email is not set will appear when logging in and the user will not be able to add a new email.

### Version

2.0.1

### Steps to Reproduce

1. Create a new local user
2. Fill in only the username and password
3.Logging in new user prompts no E-mail settings
4. Personal settings to add E-mail and save
5.Refresh the page, the E-mail disappears, and the warning about not setting E-mail still exists.
6. Use the administrator account to modify the user interface with the same error as above.

### Screenshots

_No response_

### Logs

_No response_

### Platform

desktop

### Device

Macbook Pro

### Operating System

MacOS 12.6.6(User),Ubuntu 22.04(Server Docker)

### Browser

Edge

### Additional Context

This feature works fine in version 1.9.3

### Code of Conduct

### LINKED ISSUE #1094: Cannot change user settings without email address
### Description

Hi, 
when you want to change user specific settings for other users, without enter an email address you get the following error:

image
image

### Version

2.1.0

### Steps to Reproduce

1. Goto Users
2. Click edit user
3. Change as example the weekly limits 
4. Click safe

### Screenshots

_No response_

### Logs

_No response_

### Platform

desktop

### Device

PC

### Operating System

Windows 11

### Browser

Chrome

### Additional Context

_No response_

### Code of Conduct

gauthier-th:
#### Description

This PR prevents some weird behavior when someone try to remove the email from a local user, and add a more explicit message when there is a conflict between two email/username.

#### To-Dos

#### Issues Fixed or Closed

- Fixes #1088
- Fixes #1094


server/routes/user/usersettings.ts
```diff
@@ -99,11 +100,29 @@ userSettingsRoutes.post<
...
+        user.email = req.body.email;
+        if (
+          !user.username &&
+          user.email !== oldEmail &&
+          !oldEmail.includes('@')
+        ) {
+          user.username = oldEmail;
+        }
+      } else if (req.body.username) {
+        user.email = oldUsername || user.email;
```
fallenbagel, line 122: Wouldn't it better to use nullish Coalescing here to avoid assigning falsify values like an empty string when user.email should never be empty?

```suggestion
        user.email = oldUsername ?? user.email;
```

gauthier-th: I don't think so. On the contrary here with nullish coalescing this could assign an empty string if oldUsername is empty. We want the email to always have a value, it is mandatory.

Fallenbagel (approved):

PR #1104, opened 2024-11-15
fix: use less strict validation for external URLs

### LINKED ISSUE #1068: External URL validation is too strict
### Description

My Jellyfin, Radarr and Sonarr are all accessed on `http://melorae` within my local network. But attempting to enter this in the "External URL" doesn't work because "You must provide a valid URL".

### Version

2.0.1

### Steps to Reproduce

Settings -> Jellyfin -> External URL
enter any domain without a suffix

### Screenshots

_No response_

### Logs

_No response_

### Platform

desktop

### Device

PC

### Operating System

Ubuntu 24 (server), Windows 10 (client)

### Browser

Firefox

### Additional Context

_No response_

### Code of Conduct

gauthier-th:
#### Description

Default URL validation from the Yup module doesn't allow URLs like "http://custom-host", while it is a correct value for an external URL.

#### To-Dos

#### Issues Fixed or Closed

- Fixes #1068

### DIFF
--- a/src/components/Settings/RadarrModal/index.tsx
+++ b/src/components/Settings/RadarrModal/index.tsx
@@ -131,7 +131,7 @@ const RadarrModal = ({ onClose, radarr, onSave }: RadarrModalProps) => {
     ),
     externalUrl: Yup.string()
       .matches(
-        /^https?:\/\/(www\.)?[-a-zA-Z0-9@:%._+~#=]{1,256}(\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_+.~#?&//=]*))?$/i,
+        /^https?:\/\/(www\.)?[-a-zA-Z0-9@:%._+~#=]{1,256}(\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_+.~#?&/=]*))?$/i,
         intl.formatMessage(messages.validationApplicationUrl)
       )
       .test(
--- a/src/components/Settings/RadarrModal/index.tsx
+++ b/src/components/Settings/RadarrModal/index.tsx
@@ -130,7 +130,10 @@ const RadarrModal = ({ onClose, radarr, onSave }: RadarrModalProps) => {
       intl.formatMessage(messages.validationMinimumAvailabilityRequired)
     ),
     externalUrl: Yup.string()
-      .url(intl.formatMessage(messages.validationApplicationUrl))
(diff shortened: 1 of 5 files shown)


src/components/Settings/RadarrModal/index.tsx
```diff
@@ -130,7 +130,10 @@ const RadarrModal = ({ onClose, radarr, onSave }: RadarrModalProps) => {
       intl.formatMessage(messages.validationMinimumAvailabilityRequired)
     ),
     externalUrl: Yup.string()
-      .url(intl.formatMessage(messages.validationApplicationUrl))
+      .matches(
+        /^https?:\/\/(www\.)?[-a-zA-Z0-9@:%._+~#=]{1,256}(\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_+.~#?&//=]*))?$/i,
```
fallenbagel: What do you think of

```suggestion
        /^https?:\/\/(([-a-zA-Z0-9._+%]+(:[-a-zA-Z0-9._+%]+)?@)?([-a-zA-Z0-9._]+)(\.[a-zA-Z]{2,6})?)(:\d{1,5})?(\/[-a-zA-Z0-9()@:%_+.~#?&/=]*)?$/i,
```

(My best effort to not allow special characters in domain name as this could be syntaxically invalid [correct me if im wrong] but still allow basic auth to be included in the url)

Tested this using:
https://regexr.com/

Feel free to resolve this conversation if im wrong about it or made a mistake in my regex

gauthier-th: I took the regex from here: https://stackoverflow.com/a/3809435, and just edited it so it's not mandatory to have a TLD.
Here is the thing i tested:

```
http://domain
http://domain.com
http://domain.com:9999
http://domain.com/base
http://domain.com:9999/base
http://x.x.x.x:9999
```

The two regex match these URLs. Do you have any example of something your regex is trying to solve than mine doesn't?

fallenbagel: http://user:pass@example.com

Also not to allow
http://foo@bar.com
Or
http://foo#bar.com

gauthier-th: http://foo@bar.com/ is allowed by your regex

Fallenbagel (dismissed):

Fallenbagel (approved):

PR #1138, opened 2024-12-06
fix: use tmdb first as metadata provider and fallback to tvdb

### LINKED ISSUE #1137: If a provider lookup fails for an ID, there is no fallback to other providers
### Description

https://github.com/Fallenbagel/jellyseerr/blob/9fa47cbba2a5beff703603861cd163be0a70b100/server/lib/scanners/jellyfin/index.ts#L213-L221

If a show has both IDs assigned but the tvdb does not exist or the lookup fails, the tmdb id is not used to do a fallback lookup, leading to the show not being recognized,

Additionally, since Jellyfin uses TMDB as main provider, I suggest switching the order of lookups to prefer TMDB over TVDB.

### Version

2.1.0

### Steps to Reproduce

1. Add a bad TVDB ID to a show in Jellyfin
2. Add a correct TMDB ID to show in Jellyfin
3. Rescan
4. Show is not recognized

### Screenshots

_No response_

### Logs

_No response_

### Platform

desktop

### Device

Computer

### Operating System

Windows

### Browser

Firefox

### Additional Context

_No response_

### Code of Conduct

gauthier-th:
#### Description

This PR changes the order of the metadata provider to TMDB first and then fallback to TheTVDB if no TMDB metadata is available. Previously, TheTVDB was used first and there was no fallback if the latter failed.

#### To-Dos

#### Issues Fixed or Closed

- Fixes #1137

### DIFF
--- a/server/lib/scanners/jellyfin/index.ts
+++ b/server/lib/scanners/jellyfin/index.ts
@@ -216,7 +216,9 @@ class JellyfinScanner {
             tvId: Number(metadata.ProviderIds.Tmdb),
           });
         } catch {
-          /* empty */
+          this.log('No TMDb ID found for this title.', 'debug', {
+            jellyfinitem,
+          });
         }
       }
       if (!tvShow && metadata.ProviderIds.Tvdb) {
@@ -500,7 +502,9 @@ class JellyfinScanner {
           }
         });
       } else {
-        this.log(`failed show: ${metadata.Name}`);
+        this.log('No TMDb or TVDb ID found for this title.', 'debug', {
+          jellyfinitem,
+        });
       }
     } catch (e) {
       this.log(
--- a/server/lib/scanners/jellyfin/index.ts
+++ b/server/lib/scanners/jellyfin/index.ts
@@ -216,7 +216,7 @@ class JellyfinScanner {
             tvId: Number(metadata.ProviderIds.Tmdb),
           });
         } catch {
-          this.log('No TMDb ID found for this title.', 'debug', {
+          this.log('Unable to find TMDb ID for this title.', 'debug', {
             jellyfinitem,
           });
         }
@@ -227,7 +227,9 @@ class JellyfinScanner {
             tvdbId: Number(metadata.ProviderIds.Tvdb),
           });
         } catch {
-          /* empty */
+          this.log('Unable to find TVDb ID for this title.', 'debug', {
+            jellyfinitem,
+          });
         }
       }
 
@@ -502,7 +504,7 @@ class JellyfinScanner {
           }
         });
       } else {
-        this.log('No TMDb or TVDb ID found for this title.', 'debug', {
+        this.log('No information found for this show', 'debug', {
           jellyfinitem,
         });
       }
--- a/server/lib/scanners/jellyfin/index.ts
+++ b/server/lib/scanners/jellyfin/index.ts
@@ -210,14 +210,23 @@ class JellyfinScanner {
         return;
       }
 
-      if (metadata.ProviderIds.Tvdb) {
-        tvShow = await this.tmdb.getShowByTvdbId({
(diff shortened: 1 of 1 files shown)


server/lib/scanners/jellyfin/index.ts
```diff
@@ -210,14 +210,23 @@ class JellyfinScanner {
...
-        tvShow = await this.tmdb.getTvShow({
-          tvId: Number(metadata.ProviderIds.Tmdb),
-        });
+      if (metadata.ProviderIds.Tmdb) {
+        try {
+          tvShow = await this.tmdb.getTvShow({
+            tvId: Number(metadata.ProviderIds.Tmdb),
+          });
+        } catch {
+          /* empty */
```
Shadowghost: You migth still want to log it

gauthier-th: Makes sense

fallenbagel: Debug log

server/lib/scanners/jellyfin/index.ts
```diff
@@ -210,14 +210,25 @@ class JellyfinScanner {
...
+          });
+        }
+      }
+      if (!tvShow && metadata.ProviderIds.Tvdb) {
+        try {
+          tvShow = await this.tmdb.getShowByTvdbId({
+            tvdbId: Number(metadata.ProviderIds.Tvdb),
+          });
+        } catch {
+          /* empty */
```
Shadowghost: Same here I guess 😅

gauthier-th: I added it [here](https://github.com/Fallenbagel/jellyseerr/pull/1138/commits/c164d97dde8b4b69891ea482e7b20cdcaaa3a553#diff-799b112b3397343d66cfa5140349c7551e63de0bebbf9bf628b7de831f916691R505-R507)

gauthier-th: Wait. It's not consistent. For TMDB it's logged only if there is an error while retrieving the tv show from jellyfin, while it's logged if there is a missing ID or and error for TVDB.
Do you think it makes sense to also log when there is no error but just a missing ID or not?

Shadowghost: I'd say we should log both? Not sure how you guys handle logging though

server/lib/scanners/jellyfin/index.ts
```diff
@@ -491,7 +504,9 @@ class JellyfinScanner {
           }
         });
       } else {
-        this.log(`failed show: ${metadata.Name}`);
+        this.log('No information found for this show', 'debug', {
```
fallenbagel: Why aren't we adding the `metadata.name`?

Fallenbagel (approved):

devanteweary:
tldr; Sonarr uses TVDB so let Jellyseerr use it as well.

---

Hey brother. Why not use TVDB first? 
Sonarr uses it exclusively so sometimes there are mismatches with TMDB and TVDB.

For instance, the series called "Dirty Jobs" shows as 11 seasons in TMDB but only 10 in TVDB (they consider one of the seasons a spin-off).

Whether they are correct or not, because Jellyseerr is using TMDB for this series and thinks there is a season 11, but Sonarr only sees 10 seasons, there is a mixmatch. 

I noticed this when someone requested season 11 and it added in Sonarr, but none of the seasons were monitored.  
Because season 11 does not exist in TVDB and therefore Sonarr.

Honestly, I'd love the option to turn TMDB off for series altogether. Or at least choose which one is used first.

Thank you!

Fallenbagel:
@devanteweary this pr is not at all what you think it is. That's not what this pr is for at all.

We don't use TVDB AT ALL. This PR just changes the order at which series are scanned FROM JELLYFIN. To get its external ID. Because we use tmdb. To match with tmdb.

PR #1145, opened 2024-12-12
feat(notifications): added telegram thread id's

### LINKED ISSUE #674: Add support for Telegram Topics
### Description

Currently it is only possible to send Telegram notifications to user chats and group chats. 
However, there is a third kind of group chat: [supergroups](https://core.telegram.org/api/channel#supergroups)

Supergroups allow having multiple channels (or topics) inside the same group.

Other *arr apps such as Sonarr, Radarr or Readarr support Telegram notifications for supergroups topics.

**Use case:**
I might have a group with multiple channels that might be used for different bots and people:
- A channel for Jellyseer requests notifications.
- A channel for for Radarr/Sonarr/Readarr notifications.
- A channel for health/status notifications.
- Other common channels for people to talk

**How are topic messages sent:**

Sending a notification to a supergroup is very similar to sending it to a group, using the [sendMessage](https://core.telegram.org/bots/api#sendmessage) endpoint, it just requires one additional field on the payload: 

- `message_thread_id`: Integer _(Optional)_ Unique identifier for the target message thread (topic) of the forum; for forum supergroups only

### Desired Behavior

In the Telegram notification setup section, I would like to have an additional optional field to set the Topic/Channel/Thread id (pick the desired name) just below the chat id field.

See Radarr/Sonarr Telegram notification setup as an example.

### Additional Context

_No response_

### Code of Conduct

bytebone:
#### Description

This PR adds support for Telegram Topics / Threads for both system and user notifications. I've tried my hardest to keep new code in line with existing syntax and apparent naming conventions. Should anything have slipped through, let me know. 

Following scenarios were tested for both user and system notifications:
- [x] Send to supergroup topic when both IDs are specified
- [x] Send to "general" topic when supergroup but no topic specified
- [x] Send to regular chat / channel when no thread specified (current behaviour)

As a bonus, I've already included german translations for the new strings.  

***Note:** as per `server/lib/notifications/agents/telegram.ts#L218`, when a user specifies the same group as is set for the system, the user will not receive notifications. In the supergroup construct, this may no longer be ideal.*
https://github.com/Fallenbagel/jellyseerr/blob/89831f70909df0a76dfa8a027702e4e5f9b57be8/server/lib/notifications/agents/telegram.ts#L218
<hr>

**System Settings**
| Empty | Input Validation |
| --- | --- |
| image | image |

**User Settings**
| Empty | Input Validation |
| --- | --- |
| image | image |

#### To-Dos

- [x] Database migration

#### Issues Fixed or Closed

- Fixes #674


src/components/Settings/Notifications/NotificationsTelegram.tsx
```diff
@@ -64,6 +67,18 @@ const NotificationsTelegram = () => {
         /^-?\d+$/,
         intl.formatMessage(messages.validationChatIdRequired)
       ),
+    messageThreadId: Yup.string()
+      .when(['types'], {
```
gauthier-th, line 73: Didn't you forget a `'enabled'` here? (cf. line 59)

src/i18n/locale/de.json
gauthier-th, line 1: You shouldn't write the translations here. Translations are handled by our weblate: https://translate.jellyseerr.dev/

gauthier-th (changes requested):
Just a couple thing.
And please fix the cypress tests.

src/i18n/locale/de.json
bytebone: My bad. Do I remove the lines only from the german file, or also from the english file? I am unsure what gets auto generated and what needs to be supplied in this PR.

src/components/Settings/Notifications/NotificationsTelegram.tsx
bytebone: I didn't as this field is optional. When `'enabled'` is added, the validation warning will show up after focusing the field once and not satisfying the regex pattern (i.e. leaving it empty). Right now the warning will only show up when the field is not empty and does not satisfy the pattern.

src/i18n/locale/de.json
fallenbagel: When you do `pnpm i18n:extract` what gets auto extracted is English ones. Then that's what is used by weblate

bytebone:
All lights should be green now. I've left open one conversation while waiting for a final decision. If there's anything else to do, let me know.

server/migration/1734180643080-AddTelegramMessageThreadId.ts
gauthier-th, line 1: This migration is wrong. There shouldn't be so many changes.
To generate a migration:
1. Delete your database
2. Checkout the develop branch
3. Run Jellyseerr to create a fresh database
4. Checkout your branch
5. Let TypeORM generates the migration for you with `pnpm migration:generate server/migration/YourMigrationName`

gauthier-th (changes requested):

server/migration/1734180643080-AddTelegramMessageThreadId.ts
bytebone: Very sorry for the trouble, it's my first time dealing with migrations, I didn't think of resetting the database before generating the script. The new file looks more sensical to me too, hopefully it is to your liking.

bytebone:
If there's anything else that needs doing, let me know. Otherwise I'd love to see this merged as Jellyseer is the last app on my servers that doesn't support topics yet, so I'd be very happy to get this done sometime soon.

gauthier-th (approved):
LGTM

Fallenbagel (approved):

PR #1211, opened 2024-12-31
fix(setup): fix continue button disabled on refresh in setup 3

Fallenbagel:
#### Description
This pr resolves an issue where the continue button in setup step 3 remained disabled after a page refresh even when libraries are toggled. This was happening because `mediaServerSettingsComplete` state was reset on refresh and not correctly re-initialized.

- The state is properly initialized from localStorage when the component mounts.
- Validation for enabled libraries is done when the state is hydrated so the continue button state is correct
- The continue button remains enabled if a library has already been toggled.

**Previously** (there is a library enabled after refresh but continue button disabled until i re-toggle):
image

After this PR the continue button will always stay enabled if there is a library setup. 

#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX

### DIFF
--- a/src/components/Setup/index.tsx
+++ b/src/components/Setup/index.tsx
@@ -14,10 +14,12 @@ import useLocale from '@app/hooks/useLocale';
 import useSettings from '@app/hooks/useSettings';
 import defineMessages from '@app/utils/defineMessages';
 import { MediaServerType } from '@server/constants/server';
+import type { Library } from '@server/lib/settings';
 import Image from 'next/image';
 import { useRouter } from 'next/router';
 import { useEffect, useState } from 'react';
 import { useIntl } from 'react-intl';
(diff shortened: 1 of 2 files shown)


gauthier-th (commented):
Shouldn't you also do `localStorage.removeItem('mediaServerSettingsComplete');` when the setup is done?
Could it cause some issues when you do the setup process again (i.e. when resetting the settings.json)

Fallenbagel:
I actually have not tested that. I'll take a look at it

gauthier-th:
> Shouldn't you also do `localStorage.removeItem('mediaServerSettingsComplete');` when the setup is done? Could it cause some issues when you do the setup process again (i.e. when resetting the settings.json)

Just tested and it works properly.

gauthier-th (approved):

PR #840, opened 2024-06-25
refactor: switch from Axios for Fetch API

gauthier-th:
#### Description

This PR switch our codebase from the Axios library to the native Fetch API instead.

#### To-Dos

- [ ] Successful build `yarn build`
- [ ] Translation keys `yarn i18n:extract`
- [ ] Database migration (if required)

#### Issues Fixed or Closed

- Fixes #XXXX

PR #1242, opened 2025-01-11
fix: resolve plex user mismatch due to caching issues

### LINKED ISSUE #1227: Jellyseer will log into the wrong account after a first login with plex is done
### Description

You get logged into the wrong account using Plex accounts

### Version

2.2.3

### Steps to Reproduce

Log into another account:
1. Container started
2. Log into an account with Plex
3. Log with another account → Result = Logged with admin account

Access Denied Error:
1. Set up jellyseer
2. Import the 2nd user
3. Stop the container
4. Start the container
5. Try to log in with the 2nd account → Result = Access Denied

### Screenshots

image

### Logs

https://discord.com/channels/952656177924300932/952656177924300935/1325586287146831922
https://discord.com/channels/952656177924300932/952656177924300935/1325511600286990367
https://discord.com/channels/952656177924300932/952656177924300935/1325498349742456842
https://discord.com/channels/952656177924300932/952656177924300935/1325497521203843082
https://discord.com/channels/952656177924300932/952656177924300935/1325496852467945504

### Platform

desktop

### Database

SQLite (default)

### Device

PC & Android

### Operating System

W11 & andorid

### Browser

Edge & Chrome

### Additional Context

_No response_

### Code of Conduct

Fallenbagel:
#### Description
This PR addresses an issue introduced with #840 where cached responses for PlexTV API requests returned incorrect user data when multiple users were logged in (when previous cache existed). This was caused by a shared cache key that did not account for differences in auth tokens, i.e `X-Plex-Token`. The `serializeCacheKey` method should now include headers in the cache key generation to ensure unique cache keys. This should fix the plex user mismatch that occurred due to the same cache key being used without accounting for the difference in auth token.

**_(Note)_**
_The reason this did not occur on axios was because it was creating a new axios instance in the constructor with its own base config for each user/token, while with fetch it was reusing the same instance with shared cache._

PR can be tested with the tag `preview-fix-plex-user-mismatch`

#### To-Dos

#### Issues Fixed or Closed

- Fixes #1227

### DIFF
--- a/server/api/externalapi.ts
+++ b/server/api/externalapi.ts
@@ -16,10 +16,6 @@ interface ExternalAPIOptions {
   rateLimit?: RateLimitOptions;
 }
 
-interface CustomRequestConfig extends RequestInit {
-  params?: Record<string, unknown>;
-}
-
 class ExternalAPI {
   protected fetch: typeof fetch;
   protected params: Record<string, string>;
@@ -71,10 +67,14 @@ class ExternalAPI {
     endpoint: string,
     params?: Record<string, string>,
     ttl?: number,
-    config?: CustomRequestConfig
+    config?: RequestInit
   ): Promise<T> {
     const headers = { ...this.defaultHeaders, ...config?.headers };
-    const cacheKey = this.serializeCacheKey(endpoint, config?.params, headers);
+    const cacheKey = this.serializeCacheKey(
+      endpoint,
+      { ...this.params, ...params },
+      headers
+    );
 
     const cachedItem = this.cache?.get<T>(cacheKey);
     if (cachedItem) {
@@ -114,10 +114,13 @@ class ExternalAPI {
     ttl?: number,
     config?: RequestInit
   ): Promise<T> {
-    const cacheKey = this.serializeCacheKey(endpoint, {
-      config: { ...this.params, ...params },
-      data,
-    });
+    const headers = { ...this.defaultHeaders, ...config?.headers };
+    const cacheKey = this.serializeCacheKey(
+      endpoint,
+      { ...this.params, ...params },
+      headers
+    );
+
     const cachedItem = this.cache?.get<T>(cacheKey);
     if (cachedItem) {
       return cachedItem;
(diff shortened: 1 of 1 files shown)


gauthier-th (approved):

server/api/externalapi.ts
```diff
@@ -67,12 +71,11 @@ class ExternalAPI {
...
     ttl?: number,
-    config?: RequestInit
+    config?: CustomRequestConfig
   ): Promise<T> {
-    const cacheKey = this.serializeCacheKey(endpoint, {
-      ...this.params,
-      ...params,
-    });
+    const headers = { ...this.defaultHeaders, ...config?.headers };
+    const cacheKey = this.serializeCacheKey(endpoint, config?.params, headers);
```
Fallenbagel:
```suggestion
    const cacheKey = this.serializeCacheKey(endpoint, {...this.params, ...params}, headers);
```

server/api/externalapi.ts
```diff
@@ -69,10 +69,13 @@ class ExternalAPI {
     ttl?: number,
     config?: RequestInit
   ): Promise<T> {
-    const cacheKey = this.serializeCacheKey(endpoint, {
-      ...this.params,
-      ...params,
-    });
+    const headers = { ...this.defaultHeaders, ...config?.headers };
```
gauthier-th, line 72: You should use this headers variable below, in the fetch call, instead of rewriting it
(same for the other methods)

server/api/externalapi.ts
```diff
@@ -111,10 +114,13 @@ class ExternalAPI {
...
-    const cacheKey = this.serializeCacheKey(endpoint, {
-      config: { ...this.params, ...params },
-      data,
-    });
+    const headers = { ...this.defaultHeaders, ...config?.headers };
+    const cacheKey = this.serializeCacheKey(
+      endpoint,
+      { ...this.params, ...params },
+      headers
+    );
```
gauthier-th: You forgot to include the data

server/api/externalapi.ts
```diff
@@ -155,10 +161,13 @@ class ExternalAPI {
...
```
gauthier-th: You forgot the data

server/api/externalapi.ts
```diff
@@ -325,13 +337,15 @@ class ExternalAPI {
...
+    const key = `${this.baseUrl}${endpoint}`;
+    if (!params && !headers) {
+      return key;
     }
 
-    return `${this.baseUrl}${endpoint}${JSON.stringify(params)}`;
+    return `${key}${JSON.stringify({ params, headers })}`;
   }
 
   private async getDataFromResponse(response: Response) {
```
gauthier-th: IMO the previous implementation was good, but the 2nd parameter of the function should probably not be called "params", but something like "data", as this parameter can contains URL params, headers or data (see the missing "data" you forgot above)

gauthier-th (changes requested):

server/api/externalapi.ts
fallenbagel: renamed it to options

gauthier-th (approved):

PR #1244, opened 2025-01-11
fix: Add missing parameter definition for PUT /overrideRule/{ruleId} in API spec

Wunderharke:
#### Description

While building an API client using openapi-generator-cli, I encountered an issue in overseerr-api.yml. The PUT /overrideRule/{ruleId} operation was missing a parameter definition, which I have now added.

I am not sure how Swagger UI is built or if it needs to be updated as well. Please advise if further changes are required.

#### To-Dos

I did not perform a build, as I believe it is unnecessary for this small fix. If building is required, please let me know.

### DIFF
--- a/overseerr-api.yml
+++ b/overseerr-api.yml
@@ -7018,6 +7018,12 @@ paths:
       description: Updates an Override Rule from the request body.
       tags:
         - overriderule
+      parameters:
+        - in: path
+          name: ruleId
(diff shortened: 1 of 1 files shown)


gauthier-th (approved):

Fallenbagel:
@allcontributors please add @Wunderharke  for docs

PR #1250, opened 2025-01-12
feat: possibility to set custom tmdb api key by env var

Im-a-Train:
#### Description

Make it possible to use your own tmdb api key.

Could resolve tmdb api problems on having to much requests with the same api key.

#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX

### DIFF
--- a/server/api/themoviedb/index.ts
+++ b/server/api/themoviedb/index.ts
@@ -108,7 +108,9 @@ class TheMovieDb extends ExternalAPI {
     super(
       'https://api.themoviedb.org/3',
       {
-        api_key: 'db55323b8d3e4154498498a75642b381',
+        api_key: process.env.TMDB_API_KEY
+          ? process.env.TMDB_API_KEY
+          : 'db55323b8d3e4154498498a75642b381',
       },
       {
         nodeCache: cacheManager.getCache('tmdb').data,


Fallenbagel:
Please run prettier on the file

Fallenbagel (changes requested):
Please run prettier on the file

gauthier-th:
This setting with the env variable should probably be written somewhere in the docs, but I don't know where would it be the best.
@Fallenbagel any idea?

Shadowghost:
Just as a note: Once you add this possibility and Jellyseer will have an issue (e.g. making insane amount of API calls because of a bug) it is impossible for TMDB to realize this if people use their own API keys and also to remedy it by blocking the API key.

Fallenbagel:
> Just as a note: Once you add this possibility and Jellyseer will have an issue (e.g. making insane amount of API calls because of a bug) it is impossible for TMDB to realize this if people use their own API keys and also to remedy it by blocking the API key.

Oh so it's a bad idea? 🤔

Im-a-Train:
I wonder why TMDB not already blocked the hardcoded API Key. Because it's the same, also Overseerr uses and I guess a huge amount of requests with the same key are made. We don't know who's key it's acctually is...

@Shadowghost: TMDB has a Rate Limit: https://developer.themoviedb.org/docs/rate-limiting (50 requests/s ), and I guess this is a reason, why sometimes the movies can be loaded and sometimes not. If there would be a bug of execeeding requests, they would automatically return a 429 , I hope.

@gauthier-th It would be a good idea to document it and write a how to order your own key.

Even better than an env var would be a configuration property in the ui. I can also try to add this.

Fallenbagel:
> I wonder why TMDB not already blocked the hardcoded API Key. Because it's the same, also Overseerr uses and I guess a huge amount of requests with the same key are made. We don't know who's key it's acctually is...

We didn't get blocked because there hasn't been such a case yet? You should always prepare for such a case. There was an issue with plex watchlist with overseerr and iirc jellyfin trakt also had similar issue once (v20 i think). We should always be prepared for worst case scenario.

> @Shadowghost: TMDB has a Rate Limit: https://developer.themoviedb.org/docs/rate-limiting (50 requests/s ), and I guess this is a reason, why sometimes the movies can be loaded and sometimes not. If there would be a bug of execeeding requests, they would automatically return a 429 , I hope.
>

We already ratelimit it on our side. So you won't hit the tmdb api limit because our own ratelimiter will limit before that happens. The reason why sometimes a movie doesn't load is probably related to the fetch failure undici timeout bug we have been dealing with since we migrated to fetch not tmdb ratelimiting you. And changing api key isn't gonna fix that.

Fallenbagel:
> Because it's the same, also Overseerr uses

Ideally, we should not. And this will be changed to use our own key. The volume of api calls from jellyfin would be way higher than both of our projects combined. We are not hammering the api (a malfunction that @Shadowghost cautioned) yet. Doesnt mean it might not happen in the future.

Im-a-Train:
Sounds fine to me.
I guess there is already e bug for the fetch problem, and it's not easy to solve?

I dont't know how to handle third party secrets in open source projects.
As a plaintext string, seems to be bad, because it's no secret any more... Any ideas?

Fallenbagel:
> Sounds fine to me.
> I guess there is already e bug for the fetch problem, and it's not easy to solve?
> 

It's a problem coming from undici. Not us. Even they can't tell the exact reason why and it's dependent on user network as well

gauthier-th:
After careful consideration, we don't think it would be a good idea to implement this.

As @Fallenbagel said, rate limit shouldn't be an issue: we have a rate limiter implemented inside Jellyseerr, and the limitation of 50 rps is per IP.
And as stated above, having the same API key for the app may help TMDB folks if our app has an issue and starts hammering their API.
A new API key dedicated to Jellyseerr will be used in the next release.

Thanks @Shadowghost for pointing that out :)

adam-the:
@gauthier-th I don't think this is a good justification for using a single API key across all instances, though to be honest, I'm only offended that this is closed because it's seemingly blocking the TVDB indexer feature. 


Why not simply use the User-Agent header to identify the application to TMDB, which is standard in other fields? From a quick glance, it seems to be proxied via your API anyway, so the user-agent approach is possible. At any rate, why not include this anyway? It was already written to fallback to the global key anyway.


Not really relevant to this issue but it seems bizarre to me that TMDB is willing to provide free access to their service. Purely speculative, it's likely they will switch to a model similar to TVDB in the future, if only to better understand usage patterns per user, so it would likely be wise to have this feature available pre-emptively.

gauthier-th:
> this is closed because it's seemingly blocking the TVDB indexer feature

It doesn't? I don't get why it would block it.

> Why not simply use the User-Agent header to identify the application to TMDB, which is standard in other fields?

Because it's easier for TMDB to block an API key instead of filter and block an user-agent.

> Purely speculative, it's likely they will switch to a model similar to TVDB in the future

As you said, that's purely speculative. If that day comes in the future, we'll do the necessary to make it work again.
For now, TMDB is totally fine with us using 1 API key for the whole application. And that's what Overseerr, Jellyfin, etc. are doing.

fallenbagel:
> @gauthier-th I don't think this is a good justification for using a single API key across all instances, though to be honest, I'm only offended that this is closed because it's seemingly blocking the TVDB indexer feature. 
> 
This is not what's blocking the tvdb indexer feature. There are some issues we want it to address, + that's a big change that requires proper reviewing and testing before merging, so nevertheless that pr will be blocked until it's battle-tested.

As for the single apikey justification, this is usually how it's handled in most open source apps (at least for tmdb, i.e., jellyfin is one case), so this decision is final.

adam-the:
Sorry, I misinterpreted your comment on the other thread - I see this isn't a blocker.


> As you said, that's purely speculative. If that day comes in the future, we'll do the necessary to make it work again.

Ultimately it doesn't really matter, as you say, this issue can be addressed when it arises. It seems redundant to not just include this env variable now?

On that, I'd also be concerned that the cancellation of 1 key could render numerous instances inoperable, even if those instances weren't causing any issues, causing an influx of support/spam and potential distrust.

> Because it's easier for TMDB to block an API key instead of filter and block an user-agent.

Blocking an API key may be easier due to the implementation, but providing a user-agent which contains the Jellyseerr version and perhaps other details would allow them to filter requests based on problematic versions, rather than just killing all active Jellyseerr instances.

> this is usually how it's handled in most open source apps (at least for tmdb, i.e., jellyfin is one case), so this decision is final.

Finally, and I apologise if I'm coming across argumentative, I would not say this is how it's handled in most open source apps - I rarely see applications outside of media centre & kodi add-ons/apps which don't explicitly expose their client name and version to the upstream services.

fallenbagel:
> On that, I'd also be concerned that the cancellation of 1 key could render numerous instances inoperable, even if those instances weren't causing any issues, causing an influx of support/spam and potential distrust.

If we have a broken version that hammered the api, we would definitely want that blocked. Even if it caused distrust or support/spam, you wouldn't want users to continue using a broken version that ultimately might be harder for tmdb to block. Recently, something similar happened to the trakt plugin on jellyfin where trakt blocked due to a similar.. That makes it sure that no one else can continue using the broken version. That is what we are talking about here when we say causing issues. That's not gonna be a single instance. That's going to be all of them. And blocking it would be more important than the risk of support/spam/distrust.

Also, this is a completely open source project, if anyone disagrees with our decision to keep one apikey (like how it's handled by jellyfin, overseerr etc) they could just simply go into the source code and change the apikey. Or maintain a fork. There's nothing stopping them

adam-the:
> If we have a broken version that hammered the api, we would definitely want that blocked.

Cool, they have no means of doing this right now, so maybe the project should include a user-agent to make the service aware so they don't have to block all instances without prejudice...

fallenbagel:
> > If we have a broken version that hammered the api, we would definitely want that blocked.
> 
> Cool, they have no means of doing this right now, so maybe the project should include a user-agent to make the service aware of the source, then they don't have to block all instances without prejudice... 

Blocking by user agents is more complicated than just revoking one API key. And it's not like you can't circumvent a user agent block. Just change it and recompile.

I do agree we should use a unique user agent, ideally also including the version string and a hardcoded API key.

PR #1254, opened 2025-01-13
chore(docs): add documentation about which files/data should be back up

### LINKED ISSUE #1100: Add explaination for backups in docs
Inexperienced users may find it difficult to understand how to make a Jellyseerr backup. Documentation should explain that all you need to do is make a copy of the configuration folder (once Jellyseerr is off, to avoid database corruption).

M0NsTeRRR:
#### Description

Add documentation on which files need to be backed up and which ones should be excluded.
The only missing part is the inclusion of specific instructions to help users with certain setups (e.g., AUR, Nix, Kubernetes, etc.).

#### To-Dos

#### Issues Fixed or Closed

- Fixes #1100

### DIFF
--- a/docs/using-jellyseerr/backups.md
+++ b/docs/using-jellyseerr/backups.md
@@ -1,7 +1,7 @@
 ---
 title: Backups
 description: Understand which data you should back up.
-sidebar_position: 1
+sidebar_position: 4
 ---
 
 # Which data does Jellyseerr save and where?
--- a/docs/using-jellyseerr/backups.md
+++ b/docs/using-jellyseerr/backups.md
@@ -60,7 +60,6 @@ After restoring your `db/db.sqlite3` file and, optionally, the `settings.json` f
 ├── db
 │   └── db.sqlite3
 ├── logs             <-- Optional
-├── lost+found       <-- Optional
 └── settings.json    <-- Optional (required if you want to avoid reconfiguring Jellyseerr)
 ```
 
@@ -88,7 +87,6 @@ Optionally, restore the `settings.json` file. The `config` folder structure shou
 .
 ├── cache            <-- Optional
 ├── logs             <-- Optional
-├── lost+found       <-- Optional
 └── settings.json    <-- Optional (required if you want to avoid reconfiguring Jellyseerr)
 ```
 
--- a/docs/using-jellyseerr/backups.md
+++ b/docs/using-jellyseerr/backups.md
@@ -39,9 +39,9 @@ Install [postgresql-client](https://www.postgresql.org/download/) and run the fo
 :::info
 Depending on how your PostgreSQL instance is configured, you may need to add these options to the command below.
 
-  -d, --dbname=DBNAME      database to dump
-
   -h, --host=HOSTNAME      database server host or socket directory
+
+  -p, --port=PORT          database server port number
 :::
 
 ```bash
@@ -73,9 +73,9 @@ Install the [PostgreSQL client](https://www.postgresql.org/download/) and restor
 :::info
 Depending on how your PostgreSQL instance is configured, you may need to add these options to the command below.
 
-  -d, --dbname=DBNAME      database to dump
-
   -h, --host=HOSTNAME      database server host or socket directory
+
+  -p, --port=PORT          database server port number
 :::
 
 ```bash
--- a/docs/using-jellyseerr/backups/_category_.json
+++ b/docs/using-jellyseerr/backups/_category_.json
@@ -1,6 +1,6 @@
 {
   "label": "Backups",
-  "position": 1,
+  "position": 4,
   "link": {
     "type": "generated-index",
     "title": "Backups",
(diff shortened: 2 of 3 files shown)


docs/using-jellyseerr/backups/_category_.json
```diff
@@ -0,0 +1,9 @@
+{
+  "label": "Backups",
+  "position": 1,
+  "link": {
+    "type": "generated-index",
+    "title": "Backups",
+    "description": "Understand which data you should back up"
+  }
+}
```
gauthier-th: Missing newline at the end of the file

docs/using-jellyseerr/backups/index.md
```diff
@@ -0,0 +1,67 @@
...
+
+```bash
+sqlite3 db/db.sqlite3 ".backup '/tmp/db/db.sqlite3.bak'"
+```  
+
+Then, copy the `/tmp/db/db.sqlite3.bak` file to your desired backup location.
+
+### PostgreSQL
+
+You can back up the `config` folder and dump the PostgreSQL database without stopping the Jellyseerr application.
```
gauthier-th: Could you add an example command on how to dump the database?

docs/using-jellyseerr/backups/index.md
```diff
@@ -0,0 +1,67 @@
...
+
+After restoring your `db/db.sqlite3` file and, optionally, the `settings.json` file, the `config` folder structure should look like this:
+
+```
+.
+├── cache            <-- Optional
+├── db
+│   └── db.sqlite3
+├── logs             <-- Optional
+├── lost+found       <-- Optional
```
gauthier-th: Do you think it's necessary to specify the `lost+found` directory here? It might confuse users rather than help them

docs/using-jellyseerr/backups/index.md
```diff
@@ -0,0 +1,67 @@
...
+
+### PostgreSQL
+
+After restoring the PostgreSQL database, the `config` folder structure should look like this:
+
+```
+.
+├── cache            <-- Optional
+├── logs             <-- Optional
+├── lost+found       <-- Optional
```
gauthier-th: Same as above

gauthier-th (changes requested):
Why is the Backup page inside of a Backup category? The category for only one item seems unnecessary.
And I think the position of the Backup page should be at the end of the "Using Jellyseerr" category, not at the beginning

Fallenbagel:
> Why is the Backup page inside of a Backup category? The category for only one item seems unnecessary. And I think the position of the Backup page should be at the end of the "Using Jellyseerr" category, not at the beginning

I would say it's better to add to `extending-jellyseerr` since backing up is not a core functionality but a way to extend your jellyseerr usage. Same as using an additional database or reverse proxy

M0NsTeRRR:
> Why is the Backup page inside of a Backup category? The category for only one item seems unnecessary. And I think the position of the Backup page should be at the end of the "Using Jellyseerr" category, not at the beginning

I would add a page for each deployment type with specific information.
> The only missing part is the inclusion of specific instructions to help users with certain setups (e.g., AUR, Nix, Kubernetes, etc.).

> I would argue it's better to add to extending-jellyseerr since backing up is not a core functionality but a way to extend your jellyseerr usage. Same as using an additional database or reverse proxy
Not a core functionality, but an important part of managing your application, as you prefer 👍

docs/using-jellyseerr/backups/index.md
M0NsTeRRR: It's not necessary, but I think it's better to tell the user that these files are useless, so they don't ask about them. However, I can remove it if you think it would be better the other way.

gauthier-th (dismissed):

docs/using-jellyseerr/backups.md
```diff
@@ -0,0 +1,95 @@
...
+
+After restoring your `db/db.sqlite3` file and, optionally, the `settings.json` file, the `config` folder structure should look like this:
+
+```
+.
+├── cache            <-- Optional
+├── db
+│   └── db.sqlite3
+├── logs             <-- Optional
+├── lost+found       <-- Optional
```
fallenbagel: I would recommend removing this as it's created by system and not us. I'm sure most users would've already seen this in other places. If we want to be sure, we can just add an admonition as a `note` talking about this but I would recommend removing this.

docs/using-jellyseerr/backups.md
```diff
@@ -0,0 +1,95 @@
...
+pg_restore -U <database_user> -d <database_name> /tmp/jellyseerr_db.sql
+```
+
+Optionally, restore the `settings.json` file. The `config` folder structure should look like this:
+
+```
+.
+├── cache            <-- Optional
+├── logs             <-- Optional
+├── lost+found       <-- Optional
```
fallenbagel: Same as above

Fallenbagel (changes requested):

Fallenbagel (approved):

PR #1265, opened 2025-01-15
feat(api): make rottentomatoes matching more robust

### LINKED ISSUE #1249: RottenTomatoes title match failure
### Description

Sorry in advance for this being super long for a relatively small problem.

I ran across another instance of a movie getting matched to the wrong RT ratings and have been thinking about how to improve the RT search logic and make it generally more robust. I'd like to have a discussion about some of these ideas and whether or not they're worthwhile, and then I'll be happy to open a PR implementing whatever is decided on.

For starters, the specific mismatch that prompted this involves the movie "The OctoGames" (`/movie/1028703` in Jellyseerr) which ends up showing the RT ratings for the unrelated movie "The Stranger." The reason for this failure has two parts:
1. TMDB lists the title for this movie as "The OctoGames" while RT has the title as just "Octogames" (note both the lack of "The" and the lowercase "g") with no alternative titles. Searching the RT algolia API for "The OctoGames" doesn't return the correct movie anywhere in the list of results. The movie is on RT, but you need to search without "The" to find it.
2. The Jellyseerr RT search result matching logic does a desperate pass through the results looking for any result with a matching release year, and will use that result even if the titles don't match at all. In this case, the RT results are just a pile of random popular movies that have "The" in the title, and one of them happens to have the same release year as Octogames, so it gets selected.

There are, roughly, two separate issues to address here. One is normalizing search queries before sending them to the RT algolia API, which may be as simple as stripping "the" from search queries, since the API seems to be sensitive to it. The other issue is Jellyseerr's result selection logic lacking robustness when presented with low quality search results, which is my main focus here.

For context, the [current result selection logic](https://github.com/Fallenbagel/jellyseerr/blob/99e2e17b8ada9dafa82c6f5293af5c842a5fb3c9/server/api/rating/rottentomatoes.ts#L109-L129) for movies performs 4 passes over the search results, with each pass trying to grab the first movie that satisfies the condition of the pass, and moves on to the next pass if no results match that condition. The conditions of the 4 passes are:
1. Exact title and exact year match
2. Partial title (via `String.includes`) and exact year match
3. Exact year match only **(!)**
4. Exact title match only

With the third pass being the most problematic one that caused the mismatch in this specific case.

Here is a rough (and by no means complete) collection of potential data mismatches that should be accounted for:
1. Capitalization differences in titles
2. Minor article and punctuation differences in titles, such as missing "The" or "A", or ":" vs "-"
3. Alternate titles being treated as primary titles
4. Off-by-one years, for smaller movies with complicated release stories that happened near the end/beginning of a year

An interesting observation here is that both titles and years can have three potential match states, which I'll refer to as "exact," "partial," and "none." This makes me think that the best approach would be to scan the list of search results once, assign a "score" to each result based on multiplying together subscores from how closely the titles and years match, and then picking the result with the highest score above a minimum threshold. An exact match would contribute a subscore of 1, a non-match would contribute a subscore of 0 (forcing the final score to 0 as well), and a partial match would contribute a subscore between 0 and 1 potentially based on how close the match is.

Here is my proposed scoring logic so far:
- Year:
  - exact match: 1
  - off by one: 0.5 (?)
  - anything else: 0
- Title:
  - exact match between _normalized_ (?) titles: 1
  - partial match via any sensible string similarity algo (?): the score of that algo
  - Sufficiently low (?) similarity: 0

Under this scheme, exact title and year matches would have a score of 1 and automatically win. Results that have very different titles or years that are more than 1 off would have a score of 0 and automatically lose. Anything in between will compete for the highest title*year score.

I put `(?)`s in places of the scoring logic above where I would particularly like feedback.
- The 0.5 score for off-by-one years is arbitrary. I don't think the exact value matters too much, but I could see arguments existing for lowering (0.3?) or raising (0.8?) it.
- "Normalizing" is pretty open ended. I envision it as lowercasing, stripping non-alphanumeric characters like ":", and removing minor articles like "the," but I'm not married to any of those details.
- A string similarity algo could be as simple as the one already being used (ie, string-contains) or as complicated as something like jaro-distance. I don't think this needs to be particularly robust, and we'd probably be taking on an additional library dependency if we decide to go the jaro-distance route, so I think it's reasonable to stick with a simpler check if desired.
- Depending on the string similarity algorithm, a cutoff score should be determined where values below that should be forced to 0, so we don't accidentally end up with cases where very different titles and similar years end up with a very low but non-zero total score and still win because everything else is 0. If we go with jaro, that might be "scores under 0.3 become 0." If we go with string-contains, it might just be "substring match -> 0.5, no match -> 0"

Also note that the RT algolia API returns alternate titles, but we don't currently use them. I intend to check all of them in the updated logic, and not just the primary title.

If this approach sounds good then I'll get started on a PR. Please let me know about any thoughts you have or tweaks that should be made.

### Version

latest develop

### Steps to Reproduce

1. Go to `/movie/1028703`
2. Click RT ratings
3. Observe being linked to the wrong RT movie page

### Screenshots

Screen Shot 2025-01-11 at 10 32 01 PM
Screen Shot 2025-01-11 at 10 32 26 PM
Screen Shot 2025-01-11 at 10 32 45 PM

### Logs

_No response_

### Platform

desktop

### Database

SQLite (default)

### Device

Any

### Operating System

Any

### Browser

Any

### Additional Context

_No response_

### Code of Conduct

benhaney:
#### Description

Replaces the RT search result matching logic with a ranking system that should be more robust. Also changes RT search queries to only search for the desired result type (`movie` or `tv`) and strips "The" from movie search queries.

From my testing so far, this fixes a handful of mismatched ratings, but there are some matches that this fails on in the same way as the old logic. Usually the examples that still fail are just absurd data, like results with release years that disagree by more than 1 year (eg, _Terrifier_ which finished production in 2016 but wasn't widely released until 2018) or just have bad RT entries (eg, _Nightmare Before Christmas_ which has two RT entries: one with all the ratings called _**Tim Burton's** Nightmare Before Christmas_, and another dummy entry with no ratings called just _Nightmare Before Christmas_ which always matches better).

I want to make sure this doesn't make any matches _worse_, so I'm leaving it as a draft until I have time to do some more comprehensive testing. I've scraped the RT search results for the first ~1000 movies in Jellyseerr's Movies tab and fed them through the old and new logic, but I still need to test TV show matching and changes to the search query.

Gotta name all the magic numbers too.

#### To-Dos

#### Issues Fixed or Closed

- Fixes #1249


Fallenbagel:
Changed it to feat as I would classify this as an enhancement that makes an existing feature better

benhaney:
I've done some more testing and I'm pretty happy with the results. I've diff'd the behavior of the new and old logic against the first 998 movies and 327 tv shows on jellyseerr's respective discover pages, and hand-checked the differences. The new logic found and corrected 6 movies out of my sample set that either failed to match or were matched to the wrong movie by the old logic, for an absolute improvement of about 0.6% - small, but the old logic was already pretty close to 100%, and I think this represents the majority of the room there was to improve. For TV shows on the other hand, the new logic found and corrected 15 previously incorrect results in my samples, including high profile shows like _House (2004)_, for an improvement of nearly 5% - much higher than I originally expected, especially since the logic was meant primarily for movies with tv shows as an afterthought.

I've not yet found a single regression where the new logic fails to match a movie or tv show that the old logic correctly matched, so I'm comfortable marking it as ready for review.

benhaney:
In case anyone cares, here's a list of media from my test set that's fixed by this PR, broken down by media type and then failure type.

### Movies
#### No valid RT scores exist, but matched to something else's score
- The OctoGames (2022)
- Making Squid Game: The Challenge (2023)
#### Valid RT scores exist, but failed to match to anything
- AfrAId (2024)
- Double Blind (2024)
- Lee (2024)
- The Convert (2024)
### TV Shows
#### No valid RT scores exist, but matched to something else's score
- Alice (1976)
- The City (1995)
- Smile (2002)
- A Girl's Guide to 21st Century Sex (2006)
- Masterpiece (2017)
- Kin (2018)
- Hudson & Rex (2019)
- Tamron Hall (2019)
- Legacy (2020)
#### Valid RT scores exist, but failed to match to anything
- Angel (1999)
- House (2004)
- Shameless (2011)
- Girls (2012)
#### Valid RT scores exist, but matched to something else's score
- The Hills (2006)
- Dune: Prophecy (2024)

gauthier-th (approved):
LGTM.
Thanks for the extensive tests

fallenbagel (approved):

PR #1266, opened 2025-01-15
feat(settings): add settings for custom DNS servers and IPv4 resolution first

gauthier-th:
#### Description

This PR adds settings to change the DNS servers Jellyseerr uses and to force Jellyseerr to resolve DNS queries using IPv4 first. These settings aim to make it easier for less experienced users to fix network errors related to DNS resolution.

image

#### To-Dos


Shadowghost:
Out of intereste: Doesn't the used request framework support something like Happy Eyeballs for DNS resolution?

gauthier-th:
> Out of intereste: Doesn't the used request framework support something like Happy Eyeballs for DNS resolution?

Node.js has a [loosely implementation](https://github.com/nodejs/node/pull/44731) of Happy Eyeballs, that is enabled by default.

It doesn't appear to solve issues on broken IPv6 stacks. In fact, we disable it with when forcing to IPv4 first (`net.setDefaultAutoSelectFamily(false);`).
From the tests we have done with users having this kind of issue, it seems that this solves their problem.

If you have any idea regarding this, I'd be glad to hear it, because we struggled quite a lot on this as we almost never manage to replicate the issue.

Fallenbagel (approved):

onedr0p:
@gauthier-th @Fallenbagel 

I don't want to get in an argument but applications should not support changing the network stack (DNS) at this low of a level, this feature should be reverted IMO.

DatGuy1:
This PR introduced a bug where dnsServers does not exist in settings.main, leading to `TypeError: Cannot read properties of undefined (reading 'trim')` at https://github.com/Fallenbagel/jellyseerr/blob/develop/server/index.ts#L83

Fallenbagel:
> This PR introduced a bug where dnsServers does not exist in settings.main, leading to `TypeError: Cannot read properties of undefined (reading 'trim')` at https://github.com/Fallenbagel/jellyseerr/blob/develop/server/index.ts#L83

The issue seems to be that your settings.json did not get migrated to add those keys (I am running 2.3.0 with no issues)

You might need to add those keys in manually into the main block

gauthier-th:
@DatGuy1 we can't reproduce your issue. It may be related to a misconfiguration on your side.
Can you please hop onto Discord so we can figure this out?

Fallenbagel:
> This PR introduced a bug where dnsServers does not exist in settings.main, leading to `TypeError: Cannot read properties of undefined (reading 'trim')` at https://github.com/Fallenbagel/jellyseerr/blob/develop/server/index.ts#L83

If you're installing from source, you might need to delete `.dist` and `.next` folder before rebuilding.

gauthier-th:
> > @gauthier-th @Fallenbagel.
> > I don't want to get in an argument but applications should not support changing the network stack (DNS) at this low of a level, this feature should be reverted IMO. You can take a look around at self-hosted apps and I am willing to bet the majority (if not all) of them do not support this feature nor should they.
> 
> We understand this shouldn't belong to the application land, but we had (and still have) issues everyday of users complaining because of the network error. This setting came as a convenient way for us and for users (especially for non-technical users) to fix the issue no matter how Jellyseerr has been installed.
> 
> We're using the Fetch API of the latest LTS version of Node.js, which is nowadays the "standard" way of doing requests in Node.js and JavaScript. From our point of view, our choices came down to: reverting to the Axios (a third party networking library, with a different DNS and request mechanism) which we had previously migrated to Fetch, or adding this option to make our lives easier.
> 
> As I said earlier, we're open to suggestions as we're not network guys, so it's hard for us to figure out what can be done here.

I dug a bit about this, and came across this interesting issue: https://github.com/nodejs/node/issues/54359

It says that:
> this is a known technical limitation of Node and its implementation. The reason is that there is no way for us to cancel successful connection if another goes through so we chose to only have one attempt running at any time.

So Node.js doesn't properly implement the Happy Eyeballs mechanism, but does 2 requests, one at a time.
I'm confident enough to say that this is the source of many of our problems with request timeouts.

Issue #129, opened 2022-06-07
[New Feature] Combined login


NicolaiVdS:
### Description

Adding a new way to handle logins.

### Desired Behavior

- Combined Login: login with either a local user or a Jellyfin user in the same login form
- enable/disable each login method

### Code of Conduct

- [X] I agree to follow Overseerr's Code of Conduct

NicolaiVdS:
New Settings Preview:
<img width="874" alt="image" src="https://user-images.githubusercontent.com/5257222/172459036-0e075346-34c9-466f-9ae2-3b1a2a59ffb5.png">

S0N0FBISCUIT:
Can we get an ETA on this?  It would significantly help my less tech savvy users.

Fallenbagel:
> Can we get an ETA on this?  It would significantly help my less tech savvy users.

No eta. For your less tech savvy users, just disable one of the forms so there's only one form left so they don't get confused?

PR #1286, opened 2025-01-18
feat: revamp login page and support disabling media server login

michaelhthomas:
#### Description

This PR includes enhancements from my original OpenID Connect PR which allowed disabling media server login, as well as additional enhancements to the login page to make it more usable, particularly with Plex and OpenID Connect setups.  These changes enable future work on #183, while also addressing some of the concerns from #129. Below is a quick summary of the changes

- Adds a `mediaServerLogin` setting, which can be disabled to remove the option to log in with Plex/Jellyfin/Emby from the login screen.
- Updates the user settings page layout to add a section for login options, and includes the option to enable/disable media server login within that section
- Removes the "instead of Plex OAuth" text from the local login tip, since it is no longer fully accurate and unnecessary.
- Updates the design of the login page, removing the accordion design and making it easier to access Plex / OpenID connect login options (see screenshots below)
  - This updated login page also hides the media server login options when the `mediaServerLogin` option is disabled.

| Plex and Local Login | Plex Login only |
|--------|--------|
| Plex and Local Login | Plex login only |

| Local Login only | Jellyfin/Emby Login |
| ------------ | ------------ |
| Local Login only | Jellyfin/Emby login  |

https://github.com/user-attachments/assets/c6cbb0c4-7b3b-4d66-b143-0c0e38b184fb

#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX


src/components/Common/Button/index.tsx
```diff
@@ -97,7 +98,7 @@ function Button<P extends ElementTypes = 'button'>(
   if (as === 'a') {
     return (
       <a
-        className={buttonStyle.join(' ')}
+        className={twMerge(buttonStyle)}
```
michaelhthomas, line 101: [`tailwind-merge`](https://github.com/dcastil/tailwind-merge) makes it possible to apply style overrides (like a different background color) without conflicts. I've added it as a dependency and would recommend adopting it generally, since it makes stuff like this a lot more pleasant. In this case, the Plex button would end up with the wrong background color otherwise.

src/components/Login/JellyfinLogin.tsx
```diff
@@ -1,63 +1,39 @@
 import Button from '@app/components/Common/Button';
```
michaelhthomas, line 1: I moved the Jellyfin setup to its own component (`JellyfinSetup.tsx`), since previously the entire `JellyfinLogin` component was just an if statement over the `initial` flag.

Fallenbagel:
> I have attempted to update all of the translations as part of this PR

I would suggest not to make these changes and only do `en.json` and let weblare handle it. Otherwise, it will mess up weblate

michaelhthomas:
> I would suggest not to make these changes and only do `en.json` and let weblare handle it. Otherwise, it will mess up weblate

Gotcha. I've gone through and verified most of the translations, so those I'm certain about I would be happy to just submit on weblate instead if that would be better.

Fallenbagel:
> Gotcha. I've gone through and verified most of the translations, so those I'm certain about I would be happy to just submit on weblate instead if that would be better.

That would be better, yes

Gauvino:
The design is gorgeous, good job really!

gauthier-th (commented):
What's the point of the `react-transition-group` library?
Can't this be done with [TailwindCSS `group`](https://tailwindcss.com/docs/hover-focus-and-other-states#styling-based-on-parent-state)?

gauthier-th (changes requested):
image
You should update the tests too, as they are all failing :sweat_smile:

michaelhthomas:
> What's the point of the `react-transition-group` library? Can't this be done with [TailwindCSS `group`](https://tailwindcss.com/docs/hover-focus-and-other-states#styling-based-on-parent-state)?

Unfortunately, no. `react-transition-group` allows CSS transitions between two different elements (i.e. one fades out then the other fades in). This is used to have a nice transition between the different login forms. I was unable to get the same working with the `Transition` component from `@headlessui/react`, since it only supports transitioning a single element, not between elements. Luckily, `react-transition-group` was already being used as an indirect dependency (it seems to be pretty prevalent in react), so this change just makes it direct.

src/components/Login/LocalLogin.tsx
```diff
@@ -78,19 +77,24 @@ const LocalLogin = ({ revalidate }: LocalLoginProps) => {
...
               <div>
-                <label htmlFor="email" className="text-label">
-                  {intl.formatMessage(messages.email) +
-                    ' / ' +
-                    intl.formatMessage(messages.username)}
-                </label>
-                <div className="mt-1 mb-2 sm:col-span-2 sm:mt-0">
+                <h2 className="mb-6 -mt-1 text-center text-lg font-bold text-neutral-200">
+                  {intl.formatMessage(messages.loginwithapp, {
+                    appName: 'Jellyseerr',
```
gauthier-th: Shouldn't you use the app name from the main settings here?

src/components/Login/index.tsx
```diff
@@ -86,14 +95,67 @@ const Login = () => {
...
+        settings.currentSettings.localLogin &&
+        (mediaServerLogin ? (
+          <Button
+            key="jellyseerr"
+            className="flex-1 bg-transparent"
+            onClick={() => setMediaServerLogin(false)}
+          >
+            {/* eslint-disable-next-line @next/next/no-img-element */}
+            <img src="/os_icon.svg" alt="Jellyseerr" className="mr-2 h-5" />
+            <span>Jellyseerr</span>
```
gauthier-th: Same, shouldn't you use the app name from the main settings here?

src/components/Settings/SettingsUsers/index.tsx
```diff
@@ -32,6 +40,29 @@ const messages = defineMessages('components.Settings.SettingsUsers', {
...
+
+        if (isValid) return true;
+        return this.createError({
+          path: 'localLogin | mediaServerLogin',
+          message: intl.formatMessage(messages.atLeastOneAuth),
+        });
+      },
+    });
+};
+
```
gauthier-th: Why a function for this? Instead of putting it directly inside the Component.

gauthier-th (changes requested):
Overall looks good, just a few comments :)

src/components/Login/LocalLogin.tsx
michaelhthomas: Yeah that would definitely be better. Will do.

src/components/Settings/SettingsUsers/index.tsx
michaelhthomas: I did this as part of the original OIDC PR so honestly I don't remember. I'll do that, it would be cleaner and probably more performant.

src/components/Settings/SettingsUsers/index.tsx
```diff
@@ -42,6 +51,31 @@ const SettingsUsers = () => {
...
+            if (isValid) return true;
+            return this.createError({
+              path: 'localLogin | mediaServerLogin',
+              message: intl.formatMessage(messages.atLeastOneAuth),
+            });
+          },
+        }),
+    [intl]
+  );
```
gauthier-th: We a useMemo? It feels a bit unnecessary to memoize this.
```suggestion
  const schema = yup
    .object()
    .shape({
      localLogin: yup.boolean(),
      mediaServerLogin: yup.boolean(),
    })
    .test({
      name: 'atLeastOneAuth',
      test: function (values) {
        const isValid = ['localLogin', 'mediaServerLogin'].some(
          (field) => !!values[field]
        );

        if (isValid) return true;
        return this.createError({
          path: 'localLogin | mediaServerLogin',
          message: intl.formatMessage(messages.atLeastOneAuth),
        });
      },
    });
```

michaelhthomas: It feels a bit wasteful to rebuild the schema each time the component re-renders, since it's only dependent on the intl so really only needs to be built once. But it would probably be fine without.

src/components/Login/LocalLogin.tsx
michaelhthomas: This broke the Cypress tests, whoops. (they set the applicationTitle to Overseerr)

src/components/Settings/SettingsUsers/index.tsx
gauthier-th: I used to use memo a lot some time ago, but I've since read several articles that stated that abusing memoization is in fact counterproductive, hence my comment.
And these optimizations should (hopefully) be automatically achieved soon, with React Compiler.

gauthier-th: But you can also leave it like that, it's not really a problem, i mainly said that because `yup.schema()` are also not memoized in the rest of the app

michaelhthomas: I went ahead and removed it so I think I'll just leave it like that. Hopefully React compiler will come soon 🙏

gauthier-th (dismissed):
LGTM.
I'll do a bit more tests this week-end before we can merge.
@fallenbagel do you want to review?

fallenbagel:
> LGTM.
> I'll do a bit more tests this week-end before we can merge.
> @fallenbagel do you want to review?

I can review tonight or tomorrow

gauthier-th:
@michaelhthomas can you please rebase on develop? I can't create a preview tag since there is an issue with our CI that has been fixed recently.

fallenbagel (changes requested):
@michaelhthomas could you rebase develop

gauthier-th (approved):

fallenbagel (approved):

fallenbagel (approved):

PR #1287, opened 2025-01-18
refactor(settings): move network settings to their own settings tab

gauthier-th:
#### Description

This PR moves the network settings out of the General Settings section to a new Netowrk Settings tab.

image
image

#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX


src/components/Settings/SettingsNetwork/index.tsx
```diff
@@ -0,0 +1,461 @@
...
+});
+
+const SettingsMain = () => {
+  const { addToast } = useToasts();
+  const intl = useIntl();
+  const {
+    data,
+    error,
+    mutate: revalidate,
+  } = useSWR<MainSettings>('/api/v1/settings/main');
```
fallenbagel: Shouldn't we also migrate the settings to a dedicated network block and also backend route to call /settings/network instead of main?

Fallenbagel (changes requested):

src/components/Settings/SettingsNetwork/index.tsx
gauthier-th: I don't know. Sure we can move it, but what would it be useful for? We'd also have to create a migration.
And that would be a block for only 4 settings.

As you wish.

fallenbagel: Mainly incase we need to expand. And it'd make more sense for consistency sake I meant. But I'm open to suggestions

gauthier-th: My main argument against it is the need of yet another migration script, even if it can perfectly be done. IMO the less migration the better.
But yes it would be more consistent to have it separated.

fallenbagel (approved):
LGTM

PR #1306, opened 2025-01-25
feat(blacklist): Automatically add media with blacklisted tags to the blacklist

### LINKED ISSUE #1076: Add Tag-Based Content Filtering System in Blacklist Settings
### Description

The current adult content filtering system relies solely on TMDB's **"adult"** tag, which often misses content that should be filtered, especially for family-friendly environments. While the recently introduced blacklist feature allows blocking individual titles, a more comprehensive solution is needed for proactive filtering.

### Desired Behavior

- Add a configurable tag blacklist system in the blacklist settings, fe: **blacktags** ?

- Allow administrators to specify TMDB tags to filter (e.g., "sexploitation", "erotic", "softcore")
- Apply these filters across all discovery sections (Movies, TV Shows, Trending, etc.)
- Option to enable/disable the tag filter system globally
- Ability to export/import tag lists for easy sharing of curated filters

### Additional Context

This would build upon the existing single-item blacklist feature, offering a more scalable solution for content filtering. The TMDB API already exposes these tags, as evidenced in the example: https://www.themoviedb.org/movie/37265 where tags like "erotic" are present but not caught by the current adult filter.
see #460 for more.

              **parallel topic / feature request @ overseer:** https://github.com/sct/overseerr/issues/3533

From this discussion it seems like TMDB does have something for that in their api:

https://www.themoviedb.org/talk/606da6c698f1f10029694d78

Quickly looking up the API reference for `/discover-movie` there seem to be parameters for that too ... e.g. `include_adult` or `certification_country`

https://developer.themoviedb.org/reference/discover-movie

Certifications for all Countries can be gotten here

https://developer.themoviedb.org/reference/certification-movie-list

**Response:** [certifications.json](https://github.com/Fallenbagel/jellyseerr/files/15144396/certifications.json)

----

So all in all ... shouldn't this allow for a proper filter?

Not by tags but actually by **country & age ratings (=certifications) for both movies and TV-Shows** ...

----

As for the **UI/Settings in regard to the Limits - you could basically take the values from the Parental Control settings in Jellyfin** for that - or?

image

(Sorry for the German Screenshot ...)

_Originally posted by @thiscantbeserious in https://github.com/Fallenbagel/jellyseerr/issues/460#issuecomment-2081715306_
            

### Code of Conduct

benbeauchamp7:
#### Description
This pull request introduces blacktags, which automatically blacklist media based on their tags (keywords). Blacktags are configured on the settings page and processed with the Process Blacktags job, which is configured to run weekly by default.

The job queries discover with each sort option round-robin style, with each sort being queries up to "Limit" times, where limit is configured in the settings just under where tags are entered. This is done so that media is blacklisted as much as possible no matter the sort used on the discover page (I don't want media with blacktags to appear when I sort popularity ascending for example). The default limit is 50 pages, resulting in a maximum of 32,000 blacklist entries per tag (2 media types * 50 pages * 20 entries per page * 16 sort options). The maximum limit is 250, which would cap out at 160,000 entries per tag.

##### UI
chrome_XVWinkTagi
chrome_ymiZi0kj1E

##### Settings:
chrome_CFCZpmFnuH
chrome_pSSdNveEPp
chrome_Q6d25aKuCx

##### Job:
chrome_Wzg9UWzyce

#### To-Dos

- [x] Add copy and paste functionality to keyword setting to allow sharing of blacktag configurations
- [x] Add blacklist filter to hide media added by blacktag
- [x] More bugtesting before ready for review

#### Issues Fixed or Closed

- Fixes #1076


benbeauchamp7:
That covers the functionality I was planning to implement. The only request I didn't cover from the original issue is an easy way to enable or disable the feature from the settings with a checkbox. I kind-of implemented a way to easily mass-unblock media by running the job without any blacktags configured, which will drop the blacktagged media.

Hoping to open everything up for review tomorrow night, want to do one last pass for bugs and code review

benbeauchamp7:
TMDB doesn't return keywords on their endpoints, it actually looks like
keywords are one of the only things they *don't* return. In order to not
stuff the blacklist, I would have needed to hit the API again *for each
movie* to get their keywords before actually showing the media. That felt
slow and bad, so I went with the job.

fallenbagel:
@benbeauchamp7 please rename it to `blacklistTags`

benbeauchamp7:
Changed all instances to some form of blacklistedTag. Rebased on top of develop (force push did not alter any previous commits)

docs/using-jellyseerr/settings/general.md
```diff
@@ -62,6 +62,12 @@ Set the default display language for Jellyseerr. Users can override this setting
 
 These settings filter content shown on the "Discover" home page based on regional availability and original language, respectively. The Streaming Region filters the available streaming providers on the media page. Users can override these global settings by configuring these same options in their user settings.
 
+## Blacklist Content with Tags and Limit Content Blacklisted per Tag
+
+These settings blacklist any TV shows or movies that have one of the entered tags. Entries are added to the blacklist using the "Process Blacklisted Tags" job. Removing blacklisted tags will remove media from the blacklist if it was blacklisted for that tag. To clear all blacklist entries created with tag, remove all tags in the setting and run the "Process Blacklisted Tags" job.
```
gauthier-th: You say that "Removing blacktags will remove media from the blacklist if it was blacklisted for that blacktag".
But after you say that you need to run the "Process Blacktags" job to remove all entries.
Does the Process Blacktags job also have to run to remove a single entry? It's not really clear.

benbeauchamp7:
The blacklisted tags are processed only when the job runs, so any updates
you make to the list will only go into effect the next time the job runs.
Entries for that tag aren’t removed immediately.

Will update/rephrase doc when I get home from work

docs/using-jellyseerr/settings/general.md
benbeauchamp7: Updated, lmk what you think!

benbeauchamp7:
Rebased on develop, no code was altered in the rebase

server/job/blacklistedTagsProcessor.ts
```diff
@@ -0,0 +1,186 @@
...
+  private total = 0;
+
+  public async run() {
+    this.running = true;
+
+    try {
+      await dataSource.transaction(async (em) => {
+        await this.cleanBlacklist(em);
+        await this.createBlacklistEntries(em);
+      });
```
gauthier-th, line 35: Why do you need to use a transaction here (and everywhere else in the PR)?
Won't this block all database queries until the 2 functions have been executed?

server/job/blacklistedTagsProcessor.ts
```diff
@@ -0,0 +1,186 @@
...
+  private running = false;
+  private progress = 0;
+  private total = 0;
+
+  public async run() {
+    this.running = true;
+
+    try {
+      await dataSource.transaction(async (em) => {
+        await this.cleanBlacklist(em);
```
gauthier-th, line 33: Why do you remove all previous blacklisted items before creating new entries?
The job to create blacklist entries takes a bit of time, so it means that in the meantime no tag will be blacklisted anymore?

server/job/blacklistedTagsProcessor.ts
```diff
@@ -0,0 +1,186 @@
...
+        ) {
+          await blacklistRepository.update(blacklistEntry.id, {
+            blacklistedTags: `${blacklistEntry.blacklistedTags}${keywordId},`,
+          });
+        }
+
+        continue;
+      }
+
+      // Media wasn't previously blacklisted, add it to the blacklist
```
gauthier-th: IMHO it's more clear to have an `else` block instead of the `continue`.

src/components/BlacklistedTagsBadge/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+  data: BlacklistItem;
+}
+
+const BlacklistedTagsBadge = ({ data }: BlacklistedTagsBadgeProps) => {
+  const [tagNamesBlacklistedFor, setTagNamesBlacklistedFor] =
+    useState<string>('Loading...');
+  const intl = useIntl();
+
+  useEffect(() => {
+    if (data.blacklistedTags == null) {
```
gauthier-th: Please use `===`

server/job/blacklistedTagsProcessor.ts
```diff
@@ -0,0 +1,186 @@
...
+      return;
+    }
+
+    // The maximum number of queries we're expected to execute
+    this.total =
+      2 * blacklistedTagsArr.length * pageLimit * SortOptionsIterable.length;
+
+    for (const type of [MediaType.MOVIE, MediaType.TV]) {
+      const getDiscover =
+        type == MediaType.MOVIE ? tmdb.getDiscoverMovies : tmdb.getDiscoverTv;
```
gauthier-th: Please use `===`

server/job/blacklistedTagsProcessor.ts
```diff
@@ -0,0 +1,186 @@
...
+          const response = await getDiscover({
+            page,
+            sortBy,
+            keywords: tag,
+          });
+          await this.processResults(response, tag, type, em);
+          await new Promise((res) => setTimeout(res, TMDB_API_DELAY_MS));
+
+          this.progress++;
+          if (page == 1 && response.total_pages <= queryMax) {
```
gauthier-th: Please use `===`

server/job/blacklistedTagsProcessor.ts
```diff
@@ -0,0 +1,186 @@
...
+    em: EntityManager
+  ) {
+    const blacklistRepository = em.getRepository(Blacklist);
+
+    for (const entry of response.results) {
+      const blacklistEntry = await blacklistRepository.findOne({
+        where: { tmdbId: entry.id },
+      });
+
+      if (blacklistEntry != null) {
```
gauthier-th: Please use `!==`

src/components/Blacklist/index.tsx
```diff
@@ -353,24 +410,33 @@ const BlacklistedItem = ({ item, revalidateList }: BlacklistedItemProps) => {
...
-                          alt=""
-                          className="avatar-sm ml-1.5"
-                          width={20}
-                          height={20}
-                          style={{ objectFit: 'cover' }}
-                        />
-                        <span className="ml-1 truncate text-sm font-semibold group-hover:text-white group-hover:underline">
-                          {item.user.displayName}
+                  user:
+                    item.user != null ? (
```
gauthier-th: Please use `!==`

server/job/blacklistedTagsProcessor.ts
```diff
@@ -0,0 +1,186 @@
...
+        // Don't mark manual blacklists with tags
```
gauthier-th, line 139: Why?

src/components/BlacklistBlock/index.tsx
```diff
@@ -78,22 +79,33 @@ const BlacklistBlock = ({
...
-              <Link
-                href={
-                  data.user.id === user?.id
-                    ? '/profile'
-                    : `/users/${data.user.id}`
-                }
-              >
-                <span className="font-semibold text-gray-100 transition duration-300 hover:text-white hover:underline">
-                  {data.user.displayName}
+            {data.user != null ? (
```
gauthier-th: Please use `!==`

src/components/BlacklistBlock/index.tsx
```diff
@@ -78,22 +79,33 @@ const BlacklistBlock = ({
...
+                  >
+                    <span className="font-semibold text-gray-100 transition duration-300 hover:text-white hover:underline">
+                      {data.user.displayName}
+                    </span>
+                  </Link>
                 </span>
-              </Link>
-            </span>
+              </>
+            ) : data.blacklistedTags != null && data.blacklistedTags != '' ? (
```
gauthier-th: Please use `!==`

src/components/BlacklistedTagsSelector/index.tsx
```diff
@@ -0,0 +1,438 @@
...
+          setCopied();
+        }}
+        className="input-action"
+        type="button"
+      >
+        <ClipboardDocumentIcon />
+      </button>
+    </Tooltip>
+  );
+};
```
gauthier-th: Why don't you use the already existing CopyButton instead of duplicating it here?

src/components/BlacklistedTagsSelector/index.tsx
```diff
@@ -0,0 +1,438 @@
...
+        onChange={update}
+        defaultValue={defaultValue}
+        components={{
+          DropdownIndicator: undefined,
+          IndicatorSeparator: undefined,
+          ClearIndicator: VerifyClearIndicator,
+        }}
+      />
+
+      <BlacklistedTagsCopyButton value={value ?? ''} />
```
gauthier-th: You could maybe disable the button if the value is empty?

gauthier-th (changes requested):

gauthier-th (commented):
Currently, this will not work when a user search a media that have not been added with the job, i.e. for older movies not present in the discovery sections.
I didn't check if it's possible, but can you think of a convenient way of blacklisting it / hiding it in the search page too?

server/job/blacklistedTagsProcessor.ts
benbeauchamp7: The transaction lets me clean the blacklist and create the new entries at the same moment so there isn't a time when the blacklist is empty. It also prevents the blacklist from becoming in a bad state if something strange goes wrong during the job since it would automatically rollback the transaction.

The transaction should only block queries within it that use the provided entity manager, everything else should execute on demand.

server/job/blacklistedTagsProcessor.ts
benbeauchamp7: Since it's in a transaction, there isn't a time where things are unexpectedly missing from the blacklist, both the clean and the new entries are executed at the same time when the transaction commits.

I remove all entries blacklisted by tag because the job's goal is to keep the first X pages of every search order for each tag in the blacklist. Particularly for new releases, I felt it was important to keep entries up to date without continuing the dump things into the blacklist.

server/job/blacklistedTagsProcessor.ts
benbeauchamp7: Because they weren't blacklisted for a tag, they were blacklisted manually. If I added a tag to the db, it would appear in the blacklist as blacklisted by a tag rather than a person. It would also mean it would get dropped at the start of the job, but not re-added if the job didn't find it in one of the tags. It could also get removed if a tag is removed from the setting.

Since it was blacklisted by a human, I felt it should remain as it did before and the job shouldn't touch it.

src/components/BlacklistBlock/index.tsx
benbeauchamp7: I use != here to check for both null and undefined, instead of only null. Things would similarly break if the user was undefined.

src/components/BlacklistedTagsBadge/index.tsx
benbeauchamp7: I use == here to check for both null and undefined, instead of only null. It shouldn't matter, but it keeps things safe in case it's undefined for some reason

src/components/Blacklist/index.tsx
benbeauchamp7: I use != here to check for both null and undefined, instead of only null. Things would similarly break if the user was undefined.

gauthier-th: It's maybe better to do `!item.user` then. (same for others `!=` and `==`)
I find it less confusing as you don't compare to `null` and implicitly expecting it to compare to undefined too. With `!`, you know it will be negative for any falsy value.
But it's a detail and just my opinion. @fallenbagel wdyt?

server/job/blacklistedTagsProcessor.ts
gauthier-th: Makes sense, thanks for the explanation.

src/components/Blacklist/index.tsx
benbeauchamp7: I changed this one to `!== undefined`, typescript didn't see `!item.user` as a guard against undefined for some reason. item.user has type `User | undefined`, so the compiler should throw a sufficient fit if someone tries to put null there.

Imo, I find `!== undefined` uglier than `!= null`, but I don't mind using `!==` here if it's preferred, it's y'alls repo after all.

benbeauchamp7: I'll use `item.user` here then. I prefer `!= null` personally, but it's y'alls repo and they both work the same :)

benbeauchamp7:
It looks like Jellyseerr's search uses the /search/multi endpoint [here](https://developer.themoviedb.org/reference/search-multi), which unfortunately doesn't have any keyword filtering capabilities. The only way I can think to resolve this is by
1. Querying every shown tv show/movie's keywords (feels slow and bad)
2. Increasing the limit setting to hopefully catch those shows

Since the original issue was "my feed's full of the kinda shows I don't want on my TV" I think this sufficiently solves that issue. Is probably worth writing up another issue for the search problem.

src/components/BlacklistedTagsSelector/index.tsx
benbeauchamp7: Done in [edc1802](https://github.com/fallenbagel/jellyseerr/pull/1306/commits/edc1802b3d793d6c4d795bc7190b231b41c550f0)

src/components/BlacklistedTagsSelector/index.tsx
benbeauchamp7: I was messing around with the button initially and forgot to merge my changes back into CopyButton. Fixed!

src/components/BlacklistedTagsSelector/index.tsx
benbeauchamp7: Good idea, added general CSS support for disabled button.input-actions too

src/components/BlacklistBlock/index.tsx
benbeauchamp7: Became `) : data.blacklistedTags ? (`

src/components/BlacklistBlock/index.tsx
benbeauchamp7: Updated: became `{data.user ? (`

src/components/Blacklist/index.tsx
benbeauchamp7: Done!

server/job/blacklistedTagsProcessor.ts
benbeauchamp7: Became `if (blacklistEntry) {`

server/job/blacklistedTagsProcessor.ts
benbeauchamp7: Done!

server/job/blacklistedTagsProcessor.ts
benbeauchamp7: Done!

src/components/BlacklistedTagsBadge/index.tsx
benbeauchamp7: Done! Became `if (!data.blacklistedTags) {`

server/job/blacklistedTagsProcessor.ts
benbeauchamp7: Done!

gauthier-th:
> It looks like Jellyseerr's search uses the /search/multi endpoint [here](https://developer.themoviedb.org/reference/search-multi), which unfortunately doesn't have any keyword filtering capabilities.

Yep, that's what I was afraid of. I'll look at TMDB forums to see if there is any open issue to add tags in the search response (or create one if there isn't), it would solve our issue.

gauthier-th:
Just created a ticket on TMDB forum to ask for the `without_keyword` filter in the `/search` endpoint and in the `/trending` endpoint: https://www.themoviedb.org/talk/67abc9c85bd9fe5057b0a872
The `without_keyword` filter already exists in the `/discover` endpoint.

gauthier-th (approved):

gauthier-th:
Tested and worked well for what I tried. A few more tests are welcome though.

benbeauchamp7:
Force push did not alter any code, only fixed the merge conflict. Unfortunately, doing that seems to reset review status.
@gauthier-th and @fallenbagel could you re-approve this merge request?

gauthier-th:
A preview tag will be available in 30min: `preview-blacklist-tags`

benbeauchamp7:
Hello! I was wondering if I needed to do anything or if I missed something to get this merged in. Thanks!

gauthier-th:
Hi! Nothing missed, we just haven't had time to look into it lately.
I'd also like a few people to test this before we merge it, I'll ask for testers in the Jellyseerr Discord.

gauthier-th (approved):
LGTM. Tested and it works well.

fallenbagel (dismissed):
LGTM. Did not test it though

fallenbagel:
> Hello! I was wondering if I needed to do anything or if I missed something to get this merged in. Thanks!

@benbeauchamp7 Hello! Could you rebase with develop branch? and also migrate `fetch` to `axios? We want to merge this in as soon as possible!

benbeauchamp7:
Everything is unchanged except the first commit (conflict resolution to use axios in the settings page) and the last commit, which changes the fetch calls I added to axios.get calls.

Thank you for your review!

src/components/BlacklistedTagsBadge/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+  useEffect(() => {
+    if (!data.blacklistedTags) {
+      return;
+    }
+
+    const keywordIds = data.blacklistedTags.slice(1, -1).split(',');
+    Promise.all(
+      keywordIds.map(async (keywordId) => {
+        try {
+          const { data }: { data: Keyword } = await axios.get(
```
gauthier-th:
```suggestion
          const { data } = await axios.get<Keyword>(
```

src/components/BlacklistedTagsSelector/index.tsx
```diff
@@ -0,0 +1,408 @@
...
+        }))
+      );
+    };
+
+    loadDefaultKeywords();
+  }, [defaultValue, onChange]);
+
+  const loadKeywordOptions = async (inputValue: string) => {
+    try {
+      const { data }: { data: TmdbKeywordSearchResponse } = await axios.get(
```
gauthier-th:
```suggestion
      const { data } = await axios.get<TmdbKeywordSearchResponse>(
```

src/components/BlacklistedTagsSelector/index.tsx
```diff
@@ -0,0 +1,408 @@
...
+
+    if (!/^(?:\d+,)*\d+$/.test(formValue)) {
+      setErrors([intl.formatMessage(messages.noSpecialCharacters)]);
+      return false;
+    }
+
+    const keywords = await Promise.allSettled(
+      formValue.split(',').map(async (keywordId) => {
+        try {
+          const { data }: { data: Keyword } = await axios.get(
```
gauthier-th:
```suggestion
          const { data } = await axios.get<Keyword>(
```

src/components/BlacklistedTagsSelector/index.tsx
```diff
@@ -0,0 +1,408 @@
...
+      }
+
+      const keywords = await Promise.all(
+        defaultValue.split(',').map(async (keywordId) => {
+          try {
+            const { data } = await axios.get(`/api/v1/keyword/${keywordId}`);
+            return data;
+          } catch (err) {
+            throw new Error('Network repsonse was not ok');
+          }
```
gauthier-th: You can remove the try/catch since it's just catching the error from axios to throw another less explicit error

src/components/BlacklistedTagsSelector/index.tsx
```diff
@@ -0,0 +1,408 @@
...
+        `/api/v1/search/keyword?query=${encodeURIExtraParams(inputValue)}`
+      );
+
+      return data.results.map((result) => ({
+        label: result.name,
+        value: result.id,
+      }));
+    } catch (err) {
+      throw new Error('Network repsonse was not ok');
+    }
```
gauthier-th: Same for this try/catch

gauthier-th (changes requested):

gauthier-th (approved):

fallenbagel:
@all-contributors please add @benbeauchamp7 for code

PR #1324, opened 2025-02-02
feat(settings): overridable settings config

jetjinser:
#### Description
Load the `settings.override.json` file from the config directory, if it is set. This is useful for automating deployment tools and server setup.

##### addition
I haven't read the migration related code and hope this PR won't break it.

As well as I opened the PR blindly, I'm not sure whether it would be better to open an issue first for this feature, as this might lead to a lot of documentation maintenance work.
Apologies in advance for any confusion I may cause, I'm opening this PR because I really want to be able to use this mechanism to set it persistently on the server.

Thank you for your work and this wonderful software anyway!

#### To-Dos

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@ yarn-error.log*
 config/db/*.sqlite3*
 config/settings.json
 config/settings.old.json
+config/settings.override.json
 
 # logs
 config/logs/*.log*
--- a/server/lib/settings/index.ts
+++ b/server/lib/settings/index.ts
@@ -319,6 +319,10 @@ const SETTINGS_PATH = process.env.CONFIG_DIRECTORY
   ? `${process.env.CONFIG_DIRECTORY}/settings.json`
   : path.join(__dirname, '../../../config/settings.json');
 
+const SETTINGS_OVERRIDE_PATH = process.env.CONFIG_DIRECTORY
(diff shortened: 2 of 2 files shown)


gauthier-th:
I don't understand how is this useful for automating deployments.
If you want to have default settings before the container is created, just create a `settings.json` file with the value you want?

jetjinser:
> I don't understand how is this useful for automating deployments. If you want to have default settings before the container is created, just create a `settings.json` file with the value you want?

I didn't notice that jellyseerr can load incomplete `settings.json`! Considering that I try to avoid configuring in webui, this should be enough for me!

However I guess a separate override settings also has a benefit:
All fields configured in the webui will be retained instead of being destroyed with server updates. Specifically, if you deploy with `settings.json` that only configures the required fields, all configurations made in webui will be lost every time you redeploy.

But this is a small thing, if you feel it is not necessary, feel free to close this PR :)

gauthier-th:
> However I guess a separate override settings also has a benefit:
> All fields configured in the webui will be retained instead of being destroyed with server updates. Specifically, if you deploy with settings.json that only configures the required fields, all configurations made in webui will be lost every time you redeploy.

If you manage your settings.json correctly, you shouldn't have settings losses.
I'll close this as it seems to have very little / no interest. Thanks anyway :)

PR #1333, opened 2025-02-05
fix(overriderules): allows every user to be added to the override rules

### LINKED ISSUE #1332: Override Rules only allow you to select from first ten users
### Description

Even if I make a new rule with just user 11, although it appears in the 'main' page with their name, it does not add them to the actual rules. It's hard for me to give images without giving away some of my users PII. If unable to reproduce, I can share in a private channel.

Feature request on top - The use case for this specifically is that I want to add all 15 of my jellyseerr users to a rule (except for me) - a nicer option would be to select all or invert selection.

### Version

2.3.0

### Steps to Reproduce

1. Have 11+ users
2. Go to Services, choose any service
3. Attempt to add a rule
4. Notice that only the first ten users appear on this list. You can search for the users beyond this, and they will appear on the screen until you hit save. But if you go back to that rule, they are not there.

Expected behavior: I should be able to at the very least search and add these users, but would prefer to see them all in a list

### Screenshots

This shows that user 11 does appear on the 'main' page -
Image

But back in the rule they still do not - 

Image

### Logs

```shell

```

### Platform

desktop

### Database

SQLite (default)

### Device

All, but this was an macbook pro

### Operating System

All, but this was macOS. The server was running in a k3s cluster in Ubuntu 22.04

### Browser

All, but this was Safari. Have tested on firefox as well.

### Additional Context

I am unable to get debug logs working but I will update if I can.

### Code of Conduct

RankWeis:
#### Description

* Updates /users API to allow efficient retrieving of multiple users at once.
* Uses this API to ensure that all users selected in an overridden rule are selected.

image
image

#### To-Dos

#### Issues Fixed or Closed

- Fixes #1332

### DIFF
--- a/server/routes/user/index.ts
+++ b/server/routes/user/index.ts
@@ -32,9 +32,11 @@ const router = Router();
 
 router.get('/', async (req, res, next) => {
   try {
-    const includeIds = req.query.includeIds
-      ? req.query.includeIds.toString().split(',')
-      : [];
+    const includeIds = [
+      ...new Set(
+        req.query.includeIds ? req.query.includeIds.toString().split(',') : []
+      ),
+    ];
     const pageSize = req.query.take
       ? Number(req.query.take)
       : Math.max(10, includeIds.length);
--- a/server/routes/user/index.ts
+++ b/server/routes/user/index.ts
@@ -32,12 +32,14 @@ const router = Router();
 
 router.get('/', async (req, res, next) => {
   try {
-    const pageSize = req.query.take ? Number(req.query.take) : 10;
-    const skip = req.query.skip ? Number(req.query.skip) : 0;
-    const q = req.query.q ? req.query.q.toString().toLowerCase() : '';
     const includeIds = req.query.includeIds
       ? req.query.includeIds.toString().split(',')
       : [];
+    const pageSize = req.query.take
+      ? Number(req.query.take)
+      : Math.max(10, includeIds.length);
+    const skip = req.query.skip ? Number(req.query.skip) : 0;
+    const q = req.query.q ? req.query.q.toString().toLowerCase() : '';
     let query = getRepository(User).createQueryBuilder('user');
 
     if (q) {
--- a/server/routes/user/index.ts
+++ b/server/routes/user/index.ts
@@ -42,6 +42,9 @@ router.get('/', async (req, res, next) => {
       : Math.max(10, includeIds.length);
     const skip = req.query.skip ? Number(req.query.skip) : 0;
     const q = req.query.q ? req.query.q.toString().toLowerCase() : '';
+    const includeIds = req.query.includeIds
+      ? req.query.includeIds.toString().split(',')
+      : '';
     let query = getRepository(User).createQueryBuilder('user');
 
     if (q) {
--- a/server/routes/user/index.ts
+++ b/server/routes/user/index.ts
@@ -37,7 +37,7 @@ router.get('/', async (req, res, next) => {
     const q = req.query.q ? req.query.q.toString().toLowerCase() : '';
     const includeIds = req.query.includeIds
       ? req.query.includeIds.toString().split(',')
-      : '';
+      : [];
     let query = getRepository(User).createQueryBuilder('user');
 
     if (q) {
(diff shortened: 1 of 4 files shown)


src/components/Settings/OverrideRule/OverrideRuleTile.tsx
```diff
@@ -73,18 +73,20 @@ const OverrideRuleTile = ({
...
-          .map((rule) => rule.users?.split(','))
-          .flat()
-          .filter((userId) => userId)
-          .map(async (userId) => {
-            const res = await fetch(`/api/v1/user/${userId}`);
-            if (!res.ok) throw new Error();
-            const user: User = await res.json();
-            return user;
-          })
+      const distinctUsers = new Array(
```
RankWeis: Set to distinct, back to array to rejoin. Probably unnecessary as the repository query is set to distinct already, but it doesn't hurt.

gauthier-th: It doesn't hurt but it's not necessary at all, because the User Select will not allow you to add an user twice, and because the /user endpoint will not return an user twice even if specified twice.
It would be more clean without it.

src/components/Settings/OverrideRule/OverrideRuleTile.tsx
```diff
@@ -73,18 +73,23 @@ const OverrideRuleTile = ({
...
+            .filter((users) => users)
+            .flat()
+            .flatMap((userIds) => userIds?.split(','))
+        )
+      ).join(',');
+      const res = await fetch(
+        '/api/v1/user' +
+          (distinctUsers
+            ? `?includeIds=${encodeURIComponent(distinctUsers)}`
+            : '')
```
gauthier-th: You can do a `if (rule.users)` before the `const res = ...` instead of this ternary. It would be more clear and avoid an unnecessary api call if the override rule has no rule with an user.

gauthier-th (changes requested):

src/components/Settings/OverrideRule/OverrideRuleTile.tsx
RankWeis: This is getting all the rules on the page, not just a single rule. I've implemented the fix, but the request will be making duplicates:

image

server/routes/user/index.ts
```diff
@@ -32,7 +32,12 @@ const router = Router();
 
 router.get('/', async (req, res, next) => {
   try {
-    const pageSize = req.query.take ? Number(req.query.take) : 10;
+    const includeIds = req.query.includeIds
+      ? req.query.includeIds.toString().split(',')
+      : [];
+    const pageSize = req.query.take
```
RankWeis, line 40: Without this, if you had more than ten users in a single rule, it would only display the first ten. There may be better ways of doing this (i.e. from the Tile rather than in here), but this felt cleanest to me.

ETA - this does introduce the small issue where if you pass in duplicates, your pageSize is actually too large. i.e. I pass in eleven userIds but they're all the same, this will pageSize to 11. This seems minor but wanted to call it out.

gauthier-th: Then you could make the ids unique to make sure this doesn't happen? It would easily remove the issue

server/routes/user/index.ts
```diff
@@ -32,7 +32,12 @@ const router = Router();
 
 router.get('/', async (req, res, next) => {
   try {
-    const pageSize = req.query.take ? Number(req.query.take) : 10;
+    const includeIds = req.query.includeIds
+      ? req.query.includeIds.toString().split(',')
+      : [];
```
gauthier-th:
```suggestion
    const includeIds = [
      ...new Set(
        req.query.includeIds ? req.query.includeIds.toString().split(',') : []
      ),
    ];
```

gauthier-th (changes requested):

gauthier-th (dismissed):

gauthier-th:
@RankWeis can you please rebase on develop?

RankWeis:
I’m out of the country without my computer right now; I can do this in two days, if that’s okay

RankWeis:
This rebase is complete

gauthier-th (approved):

fallenbagel (approved):

gauthier-th:
Thanks for you contribution!

gauthier-th:
@all-contributors please add @RankWeis for code

PR #1390, opened 2025-02-22
feat(airdatebadge): convert airDate from UTC to local timezone

### LINKED ISSUE #1370: It would be nice to be able to change the profile photo for a local user
### Description

Just started using Suggestarr and it uses local Jellyseerr users to automate placing redommendations within Jellyseer. It does this via local users within Jellyseer though, so it would be cool if I could change the profile picture.

### Desired Behavior

Ability to change profile picture for local users in Jellyseerr.

### Additional Context

_No response_

### Code of Conduct

fallenbagel:
#### Description
This PR will ensure that the airdate is in the user's local timezone so that its more user friendly and the relative time calculation would be consistent

#### To-Dos

#### Issues Fixed or Closed

- Fixes #1370


gauthier-th (commented):
Can't we just add the offset of the timezone to the timestamp of the date instead of using this library?

fallenbagel:
> Can't we just add the offset of the timezone to the timestamp of the date instead of using this library?

We could. But we were already using date-fns so I thought it would be better? :thinking:

gauthier-th:
> > Can't we just add the offset of the timezone to the timestamp of the date instead of using this library?
> 
> We could. But we were already using date-fns so I thought it would be better? 🤔

Well you installed `date-fns-tz`. IMO it's better to do an simple addition ourselves than installing yet another lib.

fallenbagel:
> > > Can't we just add the offset of the timezone to the timestamp of the date instead of using this library?
> > 
> > 
> > We could. But we were already using date-fns so I thought it would be better? 🤔
> 
> Well you installed `date-fns-tz`. 

yes I meant like that is still related to `date-fns`.

> IMO it's better to do an simple addition ourselves than installing yet another lib.

alright let me offset it instead

src/components/AirDateBadge/index.tsx
```diff
@@ -38,7 +39,7 @@ const AirDateBadge = ({ airDate }: AirDateBadgeProps) => {
           year: 'numeric',
           month: 'long',
           day: 'numeric',
-          timeZone: 'UTC',
+          timeZone,
```
fallenbagel, line 42: @gauthier-th No need to offset even. The relative time is already calculated correctly. Just have to pass in the proper timezone

gauthier-th (approved):

znidaxp:
I have the same problem still in 2.7.2, and I cant understand if this was implemented or not.

gauthier-th:
> I have the same problem still in 2.7.2, and I cant understand if this was implemented or not.

No, this has been reverted. See #1467

znidaxp:
> > I have the same problem still in 2.7.2, and I cant understand if this was implemented or not.
> 
> No, this has been reverted. See #1467

So if I understand this correctly this will never be implemented? Could it be an option to select if we want to convert the airtime? It really bothers me to see the wrong airtime here but not in Sonarr.

gauthier-th:
> > > I have the same problem still in 2.7.2, and I cant understand if this was implemented or not.
> > 
> > 
> > No, this has been reverted. See #1467
> 
> So if I understand this correctly this will never be implemented? Could it be an option to select if we want to convert the airtime? It really bothers me to see the wrong airtime here but not in Sonarr.

No, it's not planned because it's currently not possible.
The issue here is that we don't use the same metadata provider as Sonarr. Sonarr uses TheTVDB and we use TMDB. TMDB does not provide any timezone for the airtime, and the timezone is not the same depending on who is adding the information on TMDB.

znidaxp:
Ok I see. Would it make sense to add an option for manual delay (+-1day)?

gauthier-th:
> Ok I see. Would it make sense to add an option for manual delay (+-1day)?

I don't think so. Sometime the timezone will have a positive offset, sometime it's going to be a negative offset. A manual delay will just add more errors.

PR #1391, opened 2025-02-22
Update fr.json

ebzao:
Fix fr.json errors.


fallenbagel:
Errors for locales should be fixed in the [weblate](https://translate.jellyseerr.dev)

PR #1413, opened 2025-02-28
fix(api): make item endpoints user-independent

fallenbagel:
#### Description
Fix Jellyfin/Emby item retrieval to work properly with api tokens by using user-independent item endpoints. This will resolve the issue where disabled libraries for jellyseerr owner's would fail to scan even when we migrated to API tokens due to using the user-dependent item view endpoints.

This PR can be tested using `preview-user-independent-items`

#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX


gauthier-th (approved):
LGTM

PR #1433, opened 2025-03-09
Add an option to specify no PVC or a preexisting one

Gylesie:
#### Description

Adds an option to specify no PVC to be created for the persistence in which case an emptyDir is used as a volume. Also, adds an option to use an existing PVC instead of defining a new one.

I am new to helm templating, my changes were inspired by MoJo2600/pihole-kubernetes chart. Please do check for any mistakes.

Also, I did not know how to regenerate the readme. Maybe it is regenerated as part of CI/CD pipeline? Be free to commit some fixes if need be.

N/A

#### To-Dos

- [ ] Regenerate chart's readme

#### Issues Fixed or Closed

N/A

### DIFF
--- a/charts/jellyseerr-chart/templates/deployment.yaml
+++ b/charts/jellyseerr-chart/templates/deployment.yaml
@@ -70,8 +70,12 @@ spec:
           {{- end }}
       volumes:
         - name: config
+          {{- if .Values.config.persistence.enabled }}
           persistentVolumeClaim:
-            claimName: {{ include "jellyseerr.configPersistenceName" . }}
+            claimName: {{ if .Values.config.persistence.existingClaim }}{{ .Values.config.persistence.existingClaim }}{{- else }}{{ include "jellyseerr.configPersistenceName" . }}{{- end }}
+          {{- else }}
+          emptyDir: {}
+          {{- end }}
       {{- with .Values.volumes }}
         {{- toYaml . | nindent 8 }}
       {{- end }}
--- a/charts/jellyseerr-chart/values.yaml
+++ b/charts/jellyseerr-chart/values.yaml
@@ -53,6 +53,8 @@ service:
 # -- Creating PVC to store configuration
 config:
   persistence:
+    # -- set to true to use pvc
+    enabled: true
     # -- Size of persistent disk
     size: 5Gi
     # -- Annotations for PVCs
@@ -63,8 +65,9 @@ config:
     # -- Config name
     name: ""
     # -- Name of the permanent volume to reference in the claim.
-    # Can be used to bind to existing volumes.
     volumeName: ""
+    # -- specify an existing `PersistentVolumeClaim` to use
+    # existingClaim: ""
 
 ingress:
   enabled: false
--- a/charts/jellyseerr-chart/templates/persistentvolumeclaim.yaml
+++ b/charts/jellyseerr-chart/templates/persistentvolumeclaim.yaml
@@ -1,3 +1,5 @@
+{{- if .Values.config.persistence.enabled -}}
+{{- if not .Values.config.persistence.existingClaim -}}
 apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
@@ -18,3 +20,5 @@ spec:
   resources:
     requests:
       storage: "{{ .Values.config.persistence.size }}"
+{{- end -}}
+{{- end -}}


Gylesie:
cc @M0NsTeRRR

charts/jellyseerr-chart/templates/deployment.yaml
```diff
@@ -70,8 +70,12 @@ spec:
           {{- end }}
       volumes:
         - name: config
+          {{- if .Values.config.persistence.enabled }}
           persistentVolumeClaim:
-            claimName: {{ include "jellyseerr.configPersistenceName" . }}
+            claimName: {{ if .Values.config.persistence.existingClaim }}{{ .Values.config.persistence.existingClaim }}{{- else }}{{ include "jellyseerr.configPersistenceName" . }}{{- end }}
+          {{- else }}
+          emptyDir: {}
```
M0NsTeRRR, line 77: Why add an empty dir if there is no persistence ?

M0NsTeRRR:
Hello,

Thanks for your PR! Since it's a new feature, you need to bump the `version` in `Chart.yml` to `2.2.0`.  

To generate the README, use [helm-docs](https://github.com/norwoodj/helm-docs) by running the `helm-docs` command from the `charts` folder. I forgot to add these instructions earlier, but I'll include them later, thanks for the reminder!  

Also, if you want to use an existing PVC, as you mentioned, set `.Values.config.persistence.enabled` to `false` (the only thing needed in this PR), then configure your existing PVC as a volume and volume mount.

charts/jellyseerr-chart/templates/deployment.yaml
Gylesie: I saw it having it done like this in the chart I got inspired from. I think it might be slightly more performant rather than relying on the container's writeable layer that utilizes Copy on Write, since emptyDir is just a folder on a node and would bypass this overhead.

Gylesie:
> Hello,
> 
> Thanks for your PR! Since it's a new feature, you need to bump the `version` in `Chart.yml` to `2.2.0`.
> 
> To generate the README, use [helm-docs](https://github.com/norwoodj/helm-docs) by running the `helm-docs` command from the `charts` folder. I forgot to add these instructions earlier, but I'll include them later, thanks for the reminder!
> 
> Also, if you want to use an existing PVC, as you mentioned, set `.Values.config.persistence.enabled` to `false` (the only thing needed in this PR), then configure your existing PVC as a volume and volume mount.

Thanks for the tip! I'll do that.

While you are absolutely right, I think having an option akin to `existingClaim` would be more up to helm chart standards and would relieve the users from having to fiddle with mountpoints themselves, which might very well change in the future.

charts/jellyseerr-chart/templates/deployment.yaml
M0NsTeRRR: Okay, but it's for the configuration, so performance is not really needed. So we can keep it as it exists.

charts/jellyseerr-chart/values.yaml
```diff
@@ -63,8 +65,9 @@ config:
     # -- Config name
     name: ""
     # -- Name of the permanent volume to reference in the claim.
-    # Can be used to bind to existing volumes.
     volumeName: ""
+    # -- specify an existing `PersistentVolumeClaim` to use
```
M0NsTeRRR, line 69: `specify` => `Specify`

charts/jellyseerr-chart/values.yaml
```diff
@@ -53,6 +53,8 @@ service:
 # -- Creating PVC to store configuration
 config:
   persistence:
+    # -- set to true to use pvc
```
M0NsTeRRR, line 56: `set to true to use pvc` => `Set to true to use PVC `

M0NsTeRRR:
> > Hello,
> > Thanks for your PR! Since it's a new feature, you need to bump the `version` in `Chart.yml` to `2.2.0`.
> > To generate the README, use [helm-docs](https://github.com/norwoodj/helm-docs) by running the `helm-docs` command from the `charts` folder. I forgot to add these instructions earlier, but I'll include them later, thanks for the reminder!
> > Also, if you want to use an existing PVC, as you mentioned, set `.Values.config.persistence.enabled` to `false` (the only thing needed in this PR), then configure your existing PVC as a volume and volume mount.
> 
> Thanks for the tip! I'll do that.
> 
> While you are absolutely right, I think having an option akin to `existingClaim` would be more up to helm chart standards and would relieve the users from having to fiddle with mountpoints themselves, which might very well change in the future.

Okay, I didn't see many Helm charts doing this, but I don't disagree with your comment, so we can do it this way.

charts/jellyseerr-chart/values.yaml
M0NsTeRRR, line 56:
```suggestion
    # -- Set to true to use PVC
```

charts/jellyseerr-chart/values.yaml
M0NsTeRRR, line 69:
```suggestion
    # -- Specify an existing `PersistentVolumeClaim` to use
```

M0NsTeRRR (commented):
Except for minor typos and the suggested changes we discussed, it looks good to me :)

gauthier-th:
@Gylesie can you please fix the requested changes?

gauthier-th:
@M0NsTeRRR do you want to fix the pending comments yourself?

PR #1467, opened 2025-03-13
revert(airdate): reverts airdate offset & changes relative time to only display date (not time)

fallenbagel:
#### Description
This reverts #1390 as it created more confusion when we offsetted the air date in relevance to the timezone. It also changes the relative time to use date instead of time (so it will say `aired yesterday` `today` `5 days ago` instead of `aired x hours ago` since we dont really the airtime data.

#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX

### DIFF
--- a/src/components/AirDateBadge/index.tsx
+++ b/src/components/AirDateBadge/index.tsx
@@ -28,20 +28,8 @@ const AirDateBadge = ({ airDate }: AirDateBadgeProps) => {
     showRelative = true;
   }
 
-  const airDateStartOfDay = new Date(
-    dAirDate.getFullYear(),
-    dAirDate.getMonth(),
-    dAirDate.getDate()
-  ).getTime();
-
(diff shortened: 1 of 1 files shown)


gauthier-th (changes requested):
image
Looks like it's still using relative time

fallenbagel:
> image
> Looks like it's still using relative time

Oh yeah I made the change and forgot to push it 💀

gauthier-th (approved):

flan7:
Is it not possible to get an adjusted time from TMDB itself per this thread?

https://www.themoviedb.org/talk/54c554679251416eae010d6d

fallenbagel:
> Is it not possible to get an adjusted time from TMDB itself per this thread?
> 
> https://www.themoviedb.org/talk/54c554679251416eae010d6d
 
Thats for /airing_today endpoint only. Not really relevant here.

Issue #163, opened 2022-06-23
[Feature Request] Ability to hide all movies or all tv shows from users


n8detar:
### Description

It would be nice to be able to hide all movies or all shows if the user doesn't have permission to request that type of content. Right now even if a user is not allowed to request movies, for example, they still see movies on the discover page and can search for them and they still come up.

### Desired Behavior

If the user doesn't have permission to request or view a certain type of content (Movie or TV Show) they won't see any of that content on the Discover page or under search. I could see this being tied directly to the request permission or as its own permission.

### Code of Conduct

- [X] I agree to follow Overseerr's Code of Conduct

Fallenbagel:
Do you mean like filter certain age rated ones or genres from a specific user?

n8detar:
That would also be nice, but I would like it to be all movies or all shows. I have some users who don't have permission to request movies using the current permission system and it would be nice if they didn't see any movies at all, whether it is in the search results or on the discover page. I think it is confusing for users to see movies in their results but aren't able to actually request them. It would make it easier for them to find new shows on the discover page because they would know everything is a TV show and something they are able to request.

Fallenbagel:
> That would also be nice, but I would like it to be all movies or all shows. I have some users who don't have permission to request movies using the current permission system and it would be nice if they didn't see any movies at all, whether it is in the search results or on the discover page. I think it is confusing for users to see movies in their results but aren't able to actually request them. It would make it easier for them to find new shows on the discover page because they would know everything is a TV show and something they are able to request.

Oh so like if the person is only given permission for TV show request then movies should be hidden?

n8detar:
> Oh so like if the person is only given permission for TV show request then movies should be hidden?

Yes, exactly.

AetherVoyager:
any update regarding this ?

PR #1479, opened 2025-03-16
feat: Movies only toggle

Akhithaa-4:
#### Description
This is a feature which let admins to hide shows/tv from all the users and making jellyseerr exclusive for movies only
image
image

#### Issues Fixed or Closed

- Ref #163


gauthier-th:
My 2c on this: if we add this feature, we should add it for both movies and series, not just movies.

Akhithaa-4:
How about a radio selection like 

- All contents
- Movies Only
- TV Only

gauthier-th:
> How about a radio selection like
> 
>     * All contents
> 
>     * Movies Only
> 
>     * TV Only

Sounds good to me.

Akhithaa-4:
image

fallenbagel:
Why was this closed?

PR #1487, opened 2025-03-17
fix(auth): Bitwarden autofill fix on local/Jellyfin login (2)

### LINKED ISSUE #1452: Bitwarden Autofill not working anymore.
### Description

Hello Together,

since 2.4.0 Bitwarden is not autofilling anymore. It seems to be a false implementation of the field attribute:

https://github.com/bitwarden/clients/blob/e1415af407fef139fb69126349e5cc6d286f4474/apps/browser/src/autofill/services/collect-autofill-content.service.ts#L884-L889

Please be aware that bitwarden does not autofill if the attribute "data-bwignore" is "just" present, the value does not matter, so just deleting the "data-bwignore" in the attributes for fields it should autofill in should do the trick.

Bug introduced with: https://github.com/fallenbagel/jellyseerr/pull/1396

Hope this helps 

### Version

2.5.0

### Steps to Reproduce

Try to autofill the local login with Bitwarden.

### Screenshots

_No response_

### Logs

```shell

```

### Platform

desktop

### Database

SQLite (default)

### Device

Windows PC

### Operating System

Win11

### Browser

Chrome

### Additional Context

_No response_

### Code of Conduct

Kugelstift:
#### Description

Fixes the bwignore attribute field for wanted autofill fields.

#### To-Dos

#### Issues Fixed or Closed

- Fixes #1452

Fully Tested, maybe check if the implementation in the function is in "the state of art" im no typescript guy ^^.

### DIFF
--- a/src/components/Common/SensitiveInput/index.tsx
+++ b/src/components/Common/SensitiveInput/index.tsx
@@ -4,7 +4,6 @@ import { useState } from 'react';
 
 interface CustomInputProps extends React.ComponentProps<'input'> {
   as?: 'input';
-  'data-bwignore'?: boolean | string; // Add data-bwignore to the props type
 }
 
 interface CustomFieldProps extends React.ComponentProps<typeof Field> {
@@ -16,18 +15,13 @@ type SensitiveInputProps = CustomInputProps | CustomFieldProps;
 const SensitiveInput = ({ as = 'input', ...props }: SensitiveInputProps) => {
   const [isHidden, setHidden] = useState(true);
   const Component = as === 'input' ? 'input' : Field;
-
-  // Deconstruct to filter
-  const { 'data-bwignore': bwignore, ...filteredProps } = props;
-
   const componentProps =
     as === 'input'
-      ? filteredProps
+      ? props
       : {
-          ...filteredProps,
+          ...props,
           as: props.type === 'textarea' && !isHidden ? 'textarea' : undefined,
         };
-
   return (
     <>
       <Component
@@ -35,8 +29,6 @@ const SensitiveInput = ({ as = 'input', ...props }: SensitiveInputProps) => {
         data-form-type="other"
         data-1pignore="true"
         data-lpignore="true"
-        // Only include data-bwignore if it's not "false" always with the value "true"
-        {...(bwignore !== 'false' ? { 'data-bwignore': 'true' } : {})}
         {...componentProps}
         className={`rounded-l-only ${componentProps.className ?? ''}`}
         type={
--- a/src/components/Common/SensitiveInput/index.tsx
+++ b/src/components/Common/SensitiveInput/index.tsx
@@ -4,7 +4,7 @@ import { useState } from 'react';
 
 interface CustomInputProps extends React.ComponentProps<'input'> {
   as?: 'input';
-  'data-bwignore'?: boolean | string;  // Add data-bwignore to the props type
+  'data-bwignore'?: boolean | string; // Add data-bwignore to the props type
 }
 
 interface CustomFieldProps extends React.ComponentProps<typeof Field> {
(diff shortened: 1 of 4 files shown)


src/components/Common/SensitiveInput/index.tsx
gauthier-th, line 1: I don't understand why you changed this file.
I don't think all of this is necessary, since the `data-bwignore` prop will be passed to the `<Component>` with `...componentProps`, which already contains the `props` passed to `<SensitiveInput>`.

gauthier-th (changes requested):

src/components/Common/SensitiveInput/index.tsx
Kugelstift: Hey @gauthier-th 

it is necessary to filter out `data-bwignore` IF it is set to false, but set it to true and pass it to `<Component>` (Line 39) if it is true or not there at all (in E.g. the JellyfinLogin.tsx), to archieve the same behavior as the other things under `<Component>`.

It will not be passed to `{...componentProps}` because the bwignore attribute is filtered out (Line 21)(`Props` is decontsructed), then the rest is passed to the `...filteredProps` which will be used to fill the `{...componentProps}` (Line 23).

All that to archieve that the other files (E.g. JellyfinLogin.tsx) can be handled the same way as with any other Password Manager. (No thinking about this in the future, attribute X=false will allow and attribute X=true will ignore)

The Clue is, that the attribute `data-bwignore` is not allowed to be there at all for bitwarden to fill out this field.

I hope i was able to describe this, if you need we can communicate via discord :).

Thank you !

Edited for readability :)

gauthier-th: Then why can't it be just removed when you use a `<SensitiveInput>` instead of setting it to false and then adding a condition to filter out these false values?

Kugelstift: To simulate the same behavior as the other passwort manager attributes.

If i just remove the attribute `data-bwignore` from `sensitiveinput` the standard behavior is to allow autofill for Bitwarden, not the same as the other Managers.

For example if you remove the attribute `data-1pignore` in the `JellyfinLogin.tsx` the function will add `data-1pignore="true"` to the attributes, so the standard behavior of the function is to ignore the password Managers.

This is only possible if it can be explicitely allowed to autofill. Which in bitwardens case needs the attribute removed if it is set to false in the for example `JellyfinLogin.tsx`.

gauthier-th: Ah ok. Then you could also specify `data-lpignore="true"` everywhere you use a `<SensitiveInput>` instead?

gauthier-th: @fallenbagel WDYT?

fallenbagel: I am quite confused. Correct me if I am wrong but:
```ts
  const nodeHasBwIgnoreAttribute = node.hasAttribute("data-bwignore");
    const nodeIsValidInputElement =
      nodeTagName === "input" && !this.ignoredInputTypes.has((node as HTMLInputElement).type);
    if (nodeIsValidInputElement && !nodeHasBwIgnoreAttribute) {
      return true;
    }
```

This function returns true if the node is an <input> element (with a type that isn't in the ignored list) and it hasn't been explicitly marked to be ignored via the "data-bwignore" attribute.

So I don't understand why we cant just remove the `data-bwignore` from the inputs that need auto-fill to work on? Doesn't matter if its not consistent with other managers, as it would keep the code more simple, less complex, and more maintainable (maintainability with less complex and less ambiguous code/simpler code is always better)

Kugelstift:
No Problem, i made the easy changes :) . Just wanted to provide a way which is the same for all Managers even in the future.

Please also keep in Mind that in Order to get rid of the `data-bwignore` attribute i need to remove it in sensitive input, otherwise it gets added through this function.

Thanks for the Learning ! :)

gauthier-th (approved):

PR #1491, opened 2025-03-17
feat(wehbook): add support for dynamic placeholders in webhook URL

### LINKED ISSUE #1472: Support for dynamic notification URLs
### Description

I would love to be able to send notifications to individual users when their request becomes available. My selfhosted ntfy supports creating topics on the fly and I could use a webhook notification for that. 
Unfortunately, Jellyseer's interface won't let me add a dynamic url.

Because a picture says more than a thousand words, see attached screenshot:

Image

### Desired Behavior

Being able to add a field to the webhook URL.

### Additional Context

_No response_

### Code of Conduct

0xSysR3ll:
#### Description

This PR introduces support for **dynamic placeholders in webhook URLs**, starting with `{{requestedBy_username}}`.  
Users can enable this feature in the **notification settings**, and available placeholders are displayed as a list.

Additionally, the **UI now includes a tooltip** explaining that the test webhook URL is set to `/test`.

- Should we rename **"placeholders"** to **"variables"** or **"tokens"** for better clarity?
- Is the **UI placement** of the placeholders list intuitive, or should it be adjusted? *(Open to feedback!)*
- Should we **extend support for additional placeholders** in future updates?

*Marked this feature as **experimental** to gather feedback before finalization.*

image
image

#### To-Dos

#### Issues Fixed or Closed

- Fixes #1472


gauthier-th:
> Should we rename "placeholders" to "variables" or "tokens" for better clarity?

IMO "variables" is more clear, but "placeholders" is fine too.

> Is the UI placement of the placeholders list intuitive, or should it be adjusted? (Open to feedback!)

It may quickly be a bit messy if the list is too long, don't you think so?
We could just add a link to the documentation, like here: https://docs.jellyseerr.dev/using-jellyseerr/notifications/webhook#template-variables ?

> Should we extend support for additional placeholders in future updates?

What's wrong with using the same placeholder system as current one in the webhook content itself?

0xSysR3ll:
> IMO "variables" is more clear, but "placeholders" is fine too.

Yep, totally makes sense !

> It may quickly be a bit messy if the list is too long, don't you think so? We could just add a link to the documentation, like here: https://docs.jellyseerr.dev/using-jellyseerr/notifications/webhook#template-variables ?

As there was only one variable available, it wasn't necessary. But since we include them all, then yes I will do this.

> What's wrong with using the same placeholder system as current one in the webhook content itself?

I did not think about making all variables available at first, but I guess you're right, it could be interesting for some users.

src/components/Settings/Notifications/NotificationsWebhook/index.tsx
```diff
@@ -111,8 +117,25 @@ const NotificationsWebhook = () => {
...
-        isValidURL
+        function (value) {
+          const { supportVariables } = this.parent;
+
+          // allow variable syntax through
+          if (
+            supportVariables &&
+            typeof value === 'string' &&
+            value.includes('{{')
+          ) {
```
gauthier-th: Shouldn't we just always skip validation if there is `supportVariables` enabled?

0xSysR3ll: TBH, I wasn't sure if this was over-engineered or not.
But you're right, since we can't really validate the URL, this check is meaningless after all.

src/components/Settings/Notifications/NotificationsWebhook/index.tsx
```diff
@@ -249,10 +264,59 @@ const NotificationsWebhook = () => {
...
+                  onChange={(e: React.ChangeEvent<HTMLInputElement>) =>
+                    setFieldValue('supportVariables', e.target.checked)
+                  }
+                />
+              </div>
+            </div>
+            {values.supportVariables && (
+              <div className="mt-2">
+                <Link
+                  href="https://docs.overseerr.dev/using-overseerr/notifications/webhooks#template-variables"
```
gauthier-th:
```suggestion
                  href="https://docs.jellyseerr.dev/using-jellyseerr/notifications/webhook#template-variables"
```
And can you also update the other link below?

gauthier-th: @0xSysR3ll can you please also update the link line 378? 🙏

fallenbagel (approved):

PR #1493, opened 2025-03-18
feat: Radio selection for content filtering

Akhithaa-4:

image

#### Issues Fixed or Closed

- Ref #163

### DIFF
--- a/server/entity/MediaRequest.ts
+++ b/server/entity/MediaRequest.ts
@@ -17,6 +17,7 @@ import { DbAwareColumn } from '@server/utils/DbColumnHelper';
 import { truncate } from 'lodash';
 import {
   AfterInsert,
+  AfterLoad,
   AfterUpdate,
   Column,
   Entity,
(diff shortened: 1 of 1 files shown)


server/routes/discover.ts
```diff
@@ -595,18 +638,20 @@ discoverRoutes.get<{ networkId: string }>(
...
+      message: 'TV series are disabled in movies only mode',
+    });
+  }
 
   try {
     const data = await tmdb.getDiscoverTv({
       page: Number(req.query.page),
       language: (req.query.language as string) ?? req.locale,
-      firstAirDateGte: date,
+      firstAirDateGte: new Date().toISOString().split('T')[0],
```
fallenbagel: Why was the offset removed?

fallenbagel (changes requested):

0xSysR3ll:
Hello,

Just a visual thing but the radio button group is not aligned with both title and tip

gauthier-th:
> Hello,
> 
> Just a visual thing but the radio button group is not aligned with both title and tip

It's the indented behavior. It's the same for other settings too.

PR #1500, opened 2025-03-18
Merge jellyseerr tags with existing *arr tags instead of overwriting

andrew-kennedy:
I recently got this same PR merged into overseerr (https://github.com/sct/overseerr/pull/4019), but jellyseerr seems more alive, so I am submitting it here too.

By default, jellyseerr clobbers any existing tags within sonarr/radarr when it adds it's own tags. This change merges them instead.


gauthier-th:
Thanks for you PR, but we already merge changes from upstream, this is included in #1466 that will be merged soon.

PR #184, opened 2022-07-11
feat: support OIDC authentication

michaelhthomas:
#### Description

Adds support for logging in using an OIDC Identity Provider (Auth0, Authentik, Keycloak, etc.).

Refactors Plex OAuth login to be in its own component, like the other identity providers.

#### Screenshot (if UI-related)

Login Page

Settings Page

#### To-Dos

- [x] Successful build `yarn build`
- [x] Translation keys `yarn i18n:extract`
- [ ] Database migration (if required)

#### Issues Fixed or Closed

- Fixes #183
- Fixes #86

PR #1505, opened 2025-03-20
feat: support OpenID Connect login

### LINKED ISSUE #86: [Feature request] SSO
#### Description

Jellyfin has SSO with this plugin https://github.com/Pfuenzle/jellyfin-plugin-sso and with https://github.com/goauthentik/authentik you are able to use SSO as the provider.

#### Desired Behavior

Automatically sign in if the auth provider/Jellyfin is signed into.

#### Additional Context

Provide any additional information or screenshots that may be relevant or helpful.

### LINKED ISSUE #183: Support login with OIDC

michaelhthomas:
#### Description

Adds long-awaited support for OpenID Connect as a means of logging in to Jellyseerr. This implementation supports adding several third-party OpenID Connect providers to your instance, and allows existing users to link and unlink OpenID Connect accounts from their user accounts. For details on how to configure OpenID Connect, documentation updates are included in this PR. 

image

#### Checklist

#### Issues Fixed or Closed

- Fixes #86
- Fixes #183


cordlord:
It would be an improvement to have the ability to set the id field to something other than email. Jellyfin, which I am also using with OIDC, does not have an email field, so I use usernames for user ids.

I would like for users to be able to have their existing accounts link automatically rather than them needing to go add them or me have to update all their emails.

Zegorax:
Since the migration to the new preview-OIDC Docker tag this night, I had to re-add the OIDC provider. After that, it seems that the migrations might not be running correctly or included. When performing an OIDC login, it fails and the following error message can be found : 

` Failed OIDC login attempt {"cause":"Unknown error","ip":"18.192.118.164","errorMessage":"SQLITE_ERROR: no such table: linked_accounts"}`

I've looked in `/app/dist/migration/sqlite`, and it seems that the migration file `1742858617989-AddLinkedAccount.ts` is not there.

fallenbagel:
> Since the migration to the new preview-OIDC Docker tag this night, I had to re-add the OIDC provider. After that, it seems that the migrations might not be running correctly or included. When performing an OIDC login, it fails and the following error message can be found : 
> 
> ` Failed OIDC login attempt {"cause":"Unknown error","ip":"18.192.118.164","errorMessage":"SQLITE_ERROR: no such table: linked_accounts"}`
> 
> I've looked in `/app/dist/migration/sqlite`, and it seems that the migration file `1742858617989-AddLinkedAccount.ts` is not there. 

We did not update the preview tag yet

Zegorax:
> > Since the migration to the new preview-OIDC Docker tag this night, I had to re-add the OIDC provider. After that, it seems that the migrations might not be running correctly or included. When performing an OIDC login, it fails and the following error message can be found :
> > ` Failed OIDC login attempt {"cause":"Unknown error","ip":"18.192.118.164","errorMessage":"SQLITE_ERROR: no such table: linked_accounts"}`
> > I've looked in `/app/dist/migration/sqlite`, and it seems that the migration file `1742858617989-AddLinkedAccount.ts` is not there.
> 
> We did not update the preview tag yet

I'm not sure what happened, but the tag on Docker hub was updated 12 hours ago, and this comment : https://github.com/Fallenbagel/jellyseerr/pull/184#issuecomment-2749269478 says the tag has been updated to this PR

fallenbagel:
> > > Since the migration to the new preview-OIDC Docker tag this night, I had to re-add the OIDC provider. After that, it seems that the migrations might not be running correctly or included. When performing an OIDC login, it fails and the following error message can be found :
> > > ` Failed OIDC login attempt {"cause":"Unknown error","ip":"18.192.118.164","errorMessage":"SQLITE_ERROR: no such table: linked_accounts"}`
> > > I've looked in `/app/dist/migration/sqlite`, and it seems that the migration file `1742858617989-AddLinkedAccount.ts` is not there.
> > 
> > 
> > We did not update the preview tag yet
> 
> I'm not sure what happened, but the tag on Docker hub was updated 12 hours ago, and this comment : [#184 (comment)](https://github.com/fallenbagel/jellyseerr/pull/184#issuecomment-2749269478) says the tag has been updated to this PR

I was wrong. I thought we didn't but apparently @gauthier-th did

DuvelCorp:
Lets us know when you fix the `CREATE TABLE  linked_accounts`  issue. Once done I will test the preview on my real database (I understand it should work with it)

fallenbagel:
> Lets us know when you fix the `CREATE TABLE  linked_accounts`  issue. Once done I will test the preview on my real database (I understand it should work with it) and several users having an Authentik account

https://github.com/fallenbagel/jellyseerr/issues/183#issuecomment-2749716648
You could try that too ^

DuvelCorp:
> > Lets us know when you fix the `CREATE TABLE  linked_accounts`  issue. Once done I will test the preview on my real database (I understand it should work with it) and several users having an Authentik account
> 
> [#183 (comment)](https://github.com/fallenbagel/jellyseerr/issues/183#issuecomment-2749716648) You could try that too ^

Yes but no :-D 
We know the manual CREATE works anyway.
I am gonna wait the definitive automatic creation as its the most meaningful to test (and coz I am a bit lazy also...)

fallenbagel:
> > > Lets us know when you fix the `CREATE TABLE  linked_accounts`  issue. Once done I will test the preview on my real database (I understand it should work with it) and several users having an Authentik account
> > 
> > [#183 (comment)](https://github.com/fallenbagel/jellyseerr/issues/183#issuecomment-2749716648) You could try that too ^
> 
> Yes but no :-D 
> We know the manual CREATE works anyway.
> I am gonna wait the definitive automatic creation as its the most meaningful to test (and coz I am a bit lazy also...)
> 

I was wrong again. @gauthier-th did **not** update the Image after migrations was added. So i was right the first time. We didn't update image yet. **Database migrations has been added.**

gauthier-th:
@Zegorax I just updated the preview tag: `preview-OIDC` (it's building, done in 30min), it didn't include the last commit with the migrations, my bad.

DuvelCorp:
Just tested.
STopped my :lasted Jellyseer container.
Backuped my config (including db) directory
>CHanged the tag and restarted the stack.

Jellyseer start fine and the db migration work

Configured everything in Authentik
Configured everything in Jellyseer 

Got this 
image

Triple checked my OID config on both sides and cant see what I did wrong (I also compared with my Komga OID config that works and its just the same)
According to me, the redirect-uri I specified in Authentik is correct : 
image

The last part is the slug in Jellyseer OID config.

image

Sorry for all the red hidden stuff but I dont want to show any hint about my domain name in a public place.

Any hint ??

RobertRvB:
> Just tested the last preview image Stopped my :latest Jellyseer container. Backuped my config (including db) directory Changed the tag to :preview-OIDC and restarted the stack.
> 
> Jellyseer starts fine and the db migration seems to work
> 
> Configured everything in Authentik Configured everything in Jellyseer
> 
> At first tries I had this message message from Authentik after pushing the new button on jellyseer login screen image
> 
> Because I had tried this as redirect uri : https://jellyseer.domain.com/login/oidc/callback https://jellyseer.domain.com/login/oidc/callback/slug-in-jellyseer
> 
> and nothing worked.
> 
> Then I found this [#183 (comment)](https://github.com/fallenbagel/jellyseerr/issues/183#issuecomment-2752872025)
> 
> I thus made another try without specifying a redirect URI in Authentik, and there it worked with my test user in an incognito window. The test user was not existing in Jellyseer and it created it, and logged me in
> 
> I then made another test with my own admin user, and I got this : image
> 
> Those are the logs (nothing more)
> 
> ```
> 2025-03-26T07:19:39.198Z [info]: Creating user for tom@xxxxxxx.tld {"ip":"::ffff:192.168.2.10","email":"tom@xxxxxxx.tld"}
> 
> 2025-03-26T07:19:39.210Z [error]: Failed OIDC login attempt {"cause":"Unknown error","ip":"::ffff:192.168.2.10","errorMessage":"SQLITE_CONSTRAINT: UNIQUE constraint failed: user.email"}
> ```
> 
> So the issue is that it tries to create my user. It fails because it already exists, with that email address image
> 
> The question is, why does it fails (or doesn't do) the lookup for my user email, before deciding to create the user??
> 
> Additional question : I am having hard to figure out how this will work with my existing user base. All of them are imported from Jellyfin, and thus none of them have an email address in Jellyseer. If the auth works only with emails and not username, I dont see how it can there be any match for existing users. If no username login is possible, I suppose that to use Authentik login, all of them will have to first login normally in Jellyseer and edit their profile to add their Authentik email. Its a bit annoying when you have several dozens of users.

I had the same issue, but was able to resolve it by linking my Authentik account to the automatically created Plex user.

image

Once these were linked, I did not get the error anymore.

DuvelCorp:
> > Just tested the last preview image Stopped my :latest Jellyseer container. Backuped my config (including db) directory Changed the tag to :preview-OIDC and restarted the stack.
> > Jellyseer starts fine and the db migration seems to work
> > Configured everything in Authentik Configured everything in Jellyseer
> > At first tries I had this message message from Authentik after pushing the new button on jellyseer login screen image
> > Because I had tried this as redirect uri : https://jellyseer.domain.com/login/oidc/callback https://jellyseer.domain.com/login/oidc/callback/slug-in-jellyseer
> > and nothing worked.
> > Then I found this [#183 (comment)](https://github.com/fallenbagel/jellyseerr/issues/183#issuecomment-2752872025)
> > I thus made another try without specifying a redirect URI in Authentik, and there it worked with my test user in an incognito window. The test user was not existing in Jellyseer and it created it, and logged me in
> > I then made another test with my own admin user, and I got this : image
> > Those are the logs (nothing more)
> > ```
> > 2025-03-26T07:19:39.198Z [info]: Creating user for tom@xxxxxxx.tld {"ip":"::ffff:192.168.2.10","email":"tom@xxxxxxx.tld"}
> > 
> > 2025-03-26T07:19:39.210Z [error]: Failed OIDC login attempt {"cause":"Unknown error","ip":"::ffff:192.168.2.10","errorMessage":"SQLITE_CONSTRAINT: UNIQUE constraint failed: user.email"}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So the issue is that it tries to create my user. It fails because it already exists, with that email address image
> > The question is, why does it fails (or doesn't do) the lookup for my user email, before deciding to create the user??
> > Additional question : I am having hard to figure out how this will work with my existing user base. All of them are imported from Jellyfin, and thus none of them have an email address in Jellyseer. If the auth works only with emails and not username, I dont see how it can there be any match for existing users. If no username login is possible, I suppose that to use Authentik login, all of them will have to first login normally in Jellyseer and edit their profile to add their Authentik email. Its a bit annoying when you have several dozens of users.
> 
> I had the same issue, but was able to resolve it by linking my Authentik account to the automatically created Plex user.
> 
> image
> 
> Once these were linked, I did not get the error anymore.

ok this worked indeed !

RobertRvB:
@michaelhthomas is there a way to have it automatically go to the OIDC login page (Authentik in my case) whenever that is the only available login provider? I could not find a setting for this

image

Ruakij:
> @michaelhthomas is there a way to have it automatically go to the OIDC login page (Authentik in my case) whenever that is the only available login provider? I could not find a setting for this

Such a setting doesnt exist (yet) as far as i have seen. Give them some time.

voc0der:
As someone who is clinging on to the old `preview-OIDC` from the last PR, can anyone chime in on if the upgrade process is relatively straightforward? I really appreciate the work and want to jump back to the branch with the new stuff.

michaelhthomas:
FYI, the ability to link to an existing account (with the same email) while logging in is planned. It may end up being a separate PR though, this one is quite large as it is.

michaelhthomas:
Also, I've created [a discussion](https://github.com/fallenbagel/jellyseerr/discussions/1529) for support/issues with the OpenID Connect preview. Please use this discussion for any questions or issues encountered with the OIDC preview and refrain from commenting them here, allowing this PR to be more effective for code review.

denisgabriel5:
I think I may have found a bug while setting up my homelab.

If I already have an account (don't know if it matters, but it was a Jellyfin account) and I try to login via my OIDC provider then I will get an error. My theory is that it tries to create another account and it fails because my email is the same in both the OIDC provider and Jellyseerr. Below is a log with that error. 
```
2025-04-11T22:53:59.143Z [error]: Failed OIDC login attempt {"cause":"Unknown error","ip":"::ffff:10.10.10.7","errorMessage":"SQLITE_CONSTRAINT: UNIQUE constraint failed: user.email"}
```

FirbyKirby:
> I think I may have found a bug while setting up my homelab.
> 
> If I already have an account (don't know if it matters, but it was a Jellyfin account) and I try to login via my OIDC provider then I will get an error. My theory is that it tries to create another account and it fails because my email is the same in both the OIDC provider and Jellyseerr. Below is a log with that error.
> 
> ```
> 2025-04-11T22:53:59.143Z [error]: Failed OIDC login attempt {"cause":"Unknown error","ip":"::ffff:10.10.10.7","errorMessage":"SQLITE_CONSTRAINT: UNIQUE constraint failed: user.email"}
> ```

I believe you are facing the same issue that I and others are facing right now. I [raised this same problem](https://github.com/fallenbagel/jellyseerr/discussions/1529#discussioncomment-12682751) in the support discussion thread where someone else pointed me to another discussion. Essentially, your old account isn't linked to your OpenID account yet, so it's throwing an error. It's unique to those who used the previous OIDC-preview PR I believe. I'm personally just trying to be patient and wait for the mentioned "automatic linking" feature to be added to the PR.

kernelb00t:
Hey, it's impossible to load the UI.

I replaced the official image with the image built from the PR Dockerfile. I used prod data (yes, backups) but there's a big problem here :

https://github.com/user-attachments/assets/e01343c2-a725-433f-9455-f2f9193977b9

The text is saying "Jellyseerr has been updated; please click on the button below to reload Jellyseerr."

Should I start off a clean database?

fallenbagel:
> Hey, it's impossible to load the UI.
> 
> I replaced the official image with the image built from the PR Dockerfile. I used prod data (yes, backups) but there's a big problem here :
> 
> https://github.com/user-attachments/assets/e01343c2-a725-433f-9455-f2f9193977b9
> 
> The text is saying "Jellyseerr has been updated; please click on the button below to reload Jellyseerr."
> 
> Should I start off a clean database?
> 
> Edit: this issue happened also in a previously logged in session (logged in with the stable image)

Because you built with a missing arg. When you docker build you have to pass in `--build-arg COMMIT_TAG=xxxxx`

Where `xxxxx` stands for the commit hash

ados8:
@kernelb00t you've leaked your domain.

ados8:
Be careful updating from the old OIDC build as it will break things and there's feature loss.
Summary:
- Login won't work after update (resolvable)
- OIDC won't work for existing accounts (resolvable)
- OIDC cannot be the only login item on the login screen (unresolvable)
- Redirect URI change can break authentication (this is also an older issue but related) (resolvable)

_Disclaimer: the OIDC pertains to Authentik as it's what I use, maybe unrelated to those with Authelia etc._

After update it took a bit of time to go through the issues and conduct tests to find where problems were.
Some can be resolved and others cannot, hopefully the following helps any others updating.

### Login won't work after update
When updating there's a decent change to the configuration file whereby new has parts not in the old and vice versa.
This was the first issue and while this might not be necessary it's part of the process I did in getting things going.
1. Spin up a second new instance of the docker (configure minimum needed to get through setup)

2. You'll need to configure your SSO in this new docker due to the changes

3. Manually copy sections of the settings config from the old to the new. The main difference is the OIDC, so copy the matching parts from the old to the new config. With your settings copied that should be all you need to have a now matching setup.

4. Now copy the new docker settings file and replace the settings in your original docker, this is assuming you already updated. If you haven't updated do that and then copy the settings from the new docker to your original.

5. Enable all the other login methods in the config file e.g. Jellyfin/Emby and local because you'll likely have issues. Expect that OIDC will not work, will talk about that later, try local login and/or Jellyfin/Emby to get back in.

### OIDC won't work for existing accounts
As of v2.5.2 the OIDC logins won't work for existing accounts, even if all the previous logins were OIDC.
I messed around with scopes, claims and nothing works, you'll get the following error.
image
This seems to be because there's no account matching, even with scopes and claims configured.
Usually the account anchor is the email, UPN or CN but this doesn't seem to be the case.
However, there's 2 workarounds each have their pros and cons.
1. Go into the account settings and linked accounts (/profile/settings/linked-accounts) whereby you can link your OIDC/Jellyfin/Emby. You'll need to do this for your admin account so you can log back in using OIDC after the update. Note you cannot link other user accounts to external providers on their behalf. Without a matching login method enabled they'll get stuck and won't be able to login to even link their account. You as an admin can see the account type on the users page. It's also very unlikely users will know how or be bothered to go into their settings and link the OIDC provider; in this case look at option

2. Delete each user account, ideally not admin, and ensure OIDC 'allow new users' is enabled in advance settings. This allows the user to login using OIDC without linking as the account is made on login.

As you can see there's some draw backs with both.
Keep the account and keep the users data but unless you want to reset all their passwords, login as them and link OIDC for them it's not practical.
The second option seems the best, what I went with, lose the request data and user settings but they can just login.

### OIDC cannot be the only login item on the login screen
This is unfortunate, many people maintaining the system are likely the most technical in their family so this in my opinion is the biggest hiccup for me.
The previous setup I had OIDC as the only login option, one button, nothing else.
The new system takes that away, even if you turn everything off or mess with the config there will always be 2 login options.
This wouldn't matter as much if you could force a priority but while that has been talked about its not possible yet.
What I opted for was renaming the button to 'one click login' and set the secondary login to Emby with allow account creation off.
This means a user gets a failure on attempt to login and the button name should hopefully draw them to it.
If you remove the Emby/Jellyfin option then local takes its place so you need to chose which is alongside OIDC.

### Redirect URI change can break authentication
This has been around for a while, not sure if it started with OIDC build or not, possibly.
Either way I see it causing issues for people in this thread so this should help.
I've always used /login with Authentik at the app.domain.com for NGINX.
By default sites with Authentik use / for the location.
This ensures all traffic is passed to Authentik but this has always caused redirect issues for me.
Switching to /login means, page land pushes to /login which pushes to Authentik which pushes to home page.
With / you can loop but /login the redirect is only on the sub path.
Jellyseerr will automatically push the /login page when a user is landing unauthenticated which pushes to Authentik.
Technically this reduces the security as Authentik won't kick in on other parts of the Jellyseerr site so if there's an exploit it could make you more vulnerable. You could use this only if needed.
The other part pertaining more to this update is the redirect URI has changed.
Many people still likely have `https://domain.com/login/oidc/callback/` but the URL is now `https://domain.com/login/oidc/callback/authentik`.
You can confirm this with Authentik, clear the URI field and on login attempt it should fill this address.
If you see another URL fill the field and using NGINX try replacing the aforementioned / with /login. 

Thanks team for the amazing work.

kernelb00t:
> @kernelb00t you've leaked your domain.

I don't even care, you can get my name on the Internet in 10min chrono. With a little bit of search you can find my domain name, so I don't consider it private 😉

(It's bad security if you have to keep your domain name private to remain secure)

simcop2387:
> Be careful updating from the old OIDC build as it will break things and there's feature loss. Summary:
> 
...
> 
> ### OIDC won't work for existing accounts
> 
> As of v2.5.2 the OIDC logins won't work for existing accounts, even if all the previous logins were OIDC. I messed around with scopes, claims and nothing works, you'll get the following error. 
>     1. Go into the account settings and linked accounts (/profile/settings/linked-accounts) whereby you can link your OIDC/Jellyfin/Emby. You'll need to do this for your admin account so you can log back in using OIDC after the update. Note you cannot link other user accounts to external providers on their behalf. Without a matching login method enabled they'll get stuck and won't be able to login to even link their account. You as an admin can see the account type on the users page. It's also very unlikely users will know how or be bothered to go into their settings and link the OIDC provider; in this case look at option 2.

Just a heads up, there is a way to do this but it's not available in the UI at all.

you'll need to know a few things about the users to do it, particularly whatever ID is mapped to the `sub` field of the OIDC provider.  For keycloak this is the ID field by default, an example of what the DB looks like after linking an account is like this:

```
1|keycloak|d975849e-e8c3-43bf-9be4-6fc424d7b10b|simcop2387|1
```

You'll also need to look up the user id from the database for each user and then create the mappings for them all like the following:

```
> SELECT id, username FROM user;
1|simcop2387
...
```

which would result in doing this to create the link:

```
> INSERT INTO linked_accounts (provider, sub, username, uspserId) VALUES ('keycloak', 'd975849e-e8c3-43bf-9be4-6fc424d7b10b', 'simcop2387', 1);
```

Automating this for a large setup might be difficult but if you're like me and only have a few users it's not too hard to write up all the insert statements in a text editor and then run them.

The database can be found for the docker container as part of the config volume, `db/db.sqlite` not sure about locations for any other fancier setups.

jribal:
@simcop2387 
Can confirm it works for me.

`INSERT INTO linked_accounts (provider, sub, username, userId) VALUES ("PROVIDER", "YOUR_USER_OIDC_ID", "YOUR_USER_NAME",  YOUR_JELLYSEERR_USER_ID); `

in my case it looked like this:
`INSERT INTO linked_accounts (provider, sub, username, userId) VALUES ("authentik", "subnumber123", "TESTUSER123", 123);`

I updated my container and found my OIDC users unable to login via authentik. It is their only way to login so I wanted to unblock them without deleting their accounts.

YOUR_JELLYSEERR_USER_ID: found on your jellyseerr user profile
YOUR_USER_NAME: found on you jellyseerrr user profile
YOUR_USER_OIDC_ID:  This would be your sub. On authentik go to your jellyseerr provider -> preview -> test your user and you should see a sub.

FirbyKirby:
@jribal I used your post with my Authentik installation and it worked well. I do have one quick edit for anyone following along. The values in the insert command need to be single quote, not double quote if you're using the sqlite3 command line program.

So, the template should look like this:
`INSERT INTO linked_accounts (provider, sub, username, userId) VALUES ('PROVIDER', 'YOUR_USER_OIDC_ID', 'YOUR_USER_NAME',  YOUR_JELLYSEERR_USER_ID);`

ados8:
I've been unable to get iFrame working which my users have gotten used to, can you please support this.
The problem is only with OIDC, other logins work, the former results in X-Frame-Options sameorigin failure.
Seems Jellyseerr doesn't support CSP (Content-Security-Policy) possibly the phased out allow-from?
I have other sites working with NGINX with HSTS and CSP, interestingly disabling HSTS also doesn't work but I don't want to do that anyway.
The older build got around this by opening the authentication process in a new window, as this is outside the iFrame there's no issue but this appears to have been changed.

gauthier-th:
> Also, I've created [a discussion](https://github.com/fallenbagel/jellyseerr/discussions/1529) for support/issues with the OpenID Connect preview. Please use this discussion for any questions or issues encountered with the OIDC preview and refrain from commenting them here, allowing this PR to be more effective for code review.

As stated below, please use the dedicated discussion, your messages are not related to the code review.

fabioteixei:
Hello devs, how are you?

Any idea if this PR will be merged?

Thanks for your great work.

jon4hz:
I just tested this PR with a new jellyseerr instance and for some reason the "Login with Jellyfin" page shows up, even though it's disabled in the settings. The login form doesn't actually work tho and if I try to log in with a jellyfin or jellyseerr account, I'm getting an error 500.

image


image

tom-nom-nom:
Any news on this? When it will be merged or anything...? Because this is a really cool and useful feature :]

spiritedsnowcat:
@michaelhthomas, are you able to resolve merge conflicts with this PR? I know it's potentially a tall order, but it'd be really nice to get this change reviewed and in.

michaelhthomas:
> @michaelhthomas, are you able to resolve merge conflicts with this PR? I know it's potentially a tall order, but it'd be really nice to get this change reviewed and in.

Agreed. I've been pretty busy lately but should have some time soon to get this caught up and ready to be reviewed

michaelhthomas:
@fallenbagel @gauthier-th Provided initial testing goes ok, this should be ready to review & merge. If one of you guys would update the preview image that would be great.

gauthier-th:
> @fallenbagel @gauthier-th Provided initial testing goes ok, this should be ready to review & merge. If one of you guys would update the preview image that would be great.

Nice!
Preview image is updating: `preview-OIDC`

v3DJG6GL:
Here is a summary of the issues since the last update:

- It is not possible to edit/remove an existing OIDC configuration, or add a new OIDC configuration. Jellyseerr always shows the error `Failed to save OpenID Connect provider configuration`. Beyond that there are no further Jellyseerr logs.
  - Workaround: Manually editing the `settings.json` (https://github.com/fallenbagel/jellyseerr/discussions/1529#discussioncomment-13674726)
- After updating to the latest version of `preview-OIDC` the field `scopes` in my case was changed to `openid,profile,groups,email` instead of `openid profile groups email` which resulted in a `401` error.
  - Workaround: Manually edit the `scopes` field in the `settings.json` file. (https://github.com/fallenbagel/jellyseerr/discussions/1529#discussioncomment-13680857)
- The `redirect_uri` was changed from `https://jellyseerr.DOMAIN.TLD/login/oidc/callback/authelia` to `https://jellyseerr.DOMAIN.TLD/login?provider=authelia&callback=true`.
  - Workaround: Change the field `redirect_uri` at your OIDC provider (in my case that is Authelia) (https://github.com/fallenbagel/jellyseerr/discussions/1529#discussioncomment-13680857)

gauthier-th (commented):
In the case of an user attempting to connect with an email already associated with an existing user (not having oidc), what is the expected behavior? Merge or error?
If it should throw an error, could you add a more explicit message?

docs/using-jellyseerr/settings/users/index.md
```diff
@@ -22,6 +22,14 @@ When disabled, users will only be able to sign in using their email address. Use
 
 This setting is **enabled** by default.
 
+## Enable OpenID Connect Sign-In
+
+When enabled, users will be able to sign in to Jellyseerr using their OpenID Connect credentials, provided they have linked their OpenID Connect accounts. Once enabled, the [OpenID Connect settings](./oidc.md) can be accessed using the settings cog to the right of this option, and OpenID Connect providers can be configured.
+
+When disabled, users will only be able to sign in using their email address. Users without a password set will not be able to sign in to Jellyseerr.
```
gauthier-th: > using their email address

Or their username for Jellyfin?

src/utils/oidc.ts
gauthier-th, line 1: Is it necessary to have a dedicated utils file for oidc containing only 1 function? Especially when this function is only used once.

server/routes/auth.ts
```diff
@@ -721,6 +735,274 @@ authRoutes.post('/local', async (req, res, next) => {
...
+        gravatarUrl(fullUserInfo.email, { default: 'mm', size: 200 });
+      user = new User({
+        avatar: avatar,
+        username: fullUserInfo.preferred_username,
+        email: fullUserInfo.email,
+        permissions: settings.main.defaultPermissions,
+        plexToken: '',
+        userType: UserType.LOCAL,
+      });
+      await userRepository.save(user);
```
gauthier-th, line 971: What if an user already exists with this email?

src/components/Login/ErrorCallout.tsx
gauthier-th, line 1: This file is not used anywhere?

src/components/Settings/EditOidcModal/index.tsx
```diff
@@ -0,0 +1,374 @@
...
+              <label htmlFor="oidcName" className="text-label">
+                {intl.formatMessage(messages.oidcName)}
+                <span className="label-required">*</span>
+                <span className="label-tip">
+                  {intl.formatMessage(messages.oidcNameTip)}
+                </span>
+              </label>
+              <div className="form-input-area">
+                <Field id="oidcName" name="name" type="text" />
+                <ErrorMessage className="error" component="span" name="name" />
```
gauthier-th: We are not using `ErrorMessage` in the other forms, but a `<div className="error">`, like this:
```jsx
{errors.name &&
  touched.name &&
  typeof errors.name === 'string' && (
    <div className="error">{errors.name}</div>
  )}
```
Can you please refactor to this for the sake of consistency?

gauthier-th (changes requested):
Can you please refactor the Fetch API calls to Axios?

server/utils/oidc.ts
```diff
@@ -0,0 +1,203 @@
...
+    },
+  }).then((r) => r.json());
+
+  return wellKnownInfo;
+}
+
+function getOpenIdCallbackUrl(req: Request, provider: OidcProvider) {
+  const callbackUrl = new URL(
+    `/login`,
+    `${req.protocol}://${req.headers.host}`
```
yoyecros, line 30: **question:** with reverse-proxy (also with the Network > Enable Proxy Support option enabled and the App base URL set), it stay with http://localhost:8097. Should it be replaced with the App baseUrl if set ?

docs/using-jellyseerr/settings/users/index.md
kernelb00t: Would be better to make it generic,  "identifier" or "username" is great

src/utils/oidc.ts
michaelhthomas: It's been so long I can't remember, but iirc this used to have a lot more stuff. Definitely not very useful now, I'll relocate it.

src/components/Login/ErrorCallout.tsx
michaelhthomas: Indeed it is not. I believe it was used before the login page redesign, I'll remove it.

michaelhthomas:
> In the case of an user attempting to connect with an email already associated with an existing user (not having oidc), what is the expected behavior? Merge or error? If it should throw an error, could you add a more explicit message?

For now, an error, which should be more specific with the latest changes. The end goal would be to support linking your account by logging in with a password the first time, but to get this completed I chose to leave that for a future PR.

docs/using-jellyseerr/settings/users/oidc.md
```diff
@@ -0,0 +1,82 @@
...
+Unique identifier for the provider
+
+### Scopes
+
+Comma-separated list of scopes to request from the provider
+
+### Required Claims
+
+Comma-separated list of boolean claims that are required to log in
+
```
Gylesie: These are all **space**-separated lists as far as I know from the source code.

Translation strings need to be changed accordingly as well.

Casuallynoted:
Any update on this? It seems to be in a really good place sans the change to the info about oidc scopes - but not sure if there are any other major blockers.

src/components/Login/OidcLoginButton.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const intl = useIntl();
+  const searchParams = useSearchParams();
+  const router = useRouter();
+
+  const [loading, setLoading] = useState(false);
+
+  const redirectToLogin = useCallback(async () => {
+    try {
+      const res = await axios.get<{ redirectUrl: string }>(
+        `/api/v1/auth/oidc/login/${provider.slug}`
```
Jappie3, line 57: this will query the api using `http://` (*not* `https://` [1]), returning `redirectUrl = await getOpenIdRedirectUrl(req, provider, state);`, which calls `getOpenIdCallbackUrl` to construct the callback URL.

The latter uses the protocol of `req` in the returned callback URL, which is HTTP - not what you want when it comes to authentication.

```
function getOpenIdCallbackUrl(req: Request, provider: OidcProvider) {
  const callbackUrl = new URL(
    `/login`,
    `${req.protocol}://${req.headers.host}`
  );
  callbackUrl.searchParams.set('provider', provider.slug);
  callbackUrl.searchParams.set('callback', 'true');
  return callbackUrl.toString();
}
```

There's a jellyfin plugin for SSO that has an option to override the scheme used in the callback URL (https://github.com/9p4/jellyfin-plugin-sso/blob/71bb9a2f14b74f98542f5536dfc64e258fb07cf3/SSO-Auth/Config/PluginConfiguration.cs#L132), I suggest something similar to be implemented here - I can whip up a patch tomorrow if you want.


[1] not actually sure if this is always the case, this is just what I observed on my instance. Might be different if you configure jellyseerr to use HTTPS directly?

BrenekH: > [1] not actually sure if this is always the case, this is just what I observed on my instance. Might be different if you configure jellyseerr to use HTTPS directly?

I assume you're referring to how the client reaches out to the server here, correct? Since the URL passed to axios doesn't have any protocol or domain, the browser will use whatever protocol you're currently connected to the website with.

> The latter uses the protocol of req in the returned callback URL, which is HTTP - not what you want when it comes to authentication.

`req.protocol` does return HTTPS if you enable the "Enable Proxy Support" setting in the Network tab. This setting turns on Express's `trust proxy` option which allows `req.protocol` to be set by the `X-Forwarded-Proto` header ([Express docs link](https://expressjs.com/en/4x/api.html#req.protocol)). So if you don't have "Enable Proxy Support" on, or your reverse proxy isn't setting `X-Forwarded-Proto`, `req.protocol` will always be HTTP.

That being said, I do support adding an override just in case someone happens to need it. However, I think being more clear about what "Enable Proxy Support" does in the UI would be helpful as well. The current prose only mentions client IPs, but in reality the option touches more than just IPs.

If it was a choice between ship OIDC or add the Protocol override, I would ship it, but I think the override is simple enough to add. A patch would be a good idea, perhaps as a PR to Michael's fork ([https://github.com/michaelhthomas/jellyseerr/tree/feat/oidc-login](https://github.com/michaelhthomas/jellyseerr/tree/feat/oidc-login)).

AlexTryHarder:
What is the latest :preview-OIDC branch version?
It seems to be affected https://github.com/fallenbagel/jellyseerr/issues/1787

src/components/Login/OidcLoginButton.tsx
Jappie3: > I assume you're referring to how the client reaches out to the server here, correct? Since the URL passed to axios doesn't have any protocol or domain, the browser will use whatever protocol you're currently connected to the website with.

yes, sorry - typed my comment right before going to bed :p

I have a proxy in front of Jellyseerr, so Jellyseerr sees all incoming traffic as HTTP - just tried enabling proxy support under the network tab in settings and that indeed causes the redirect URI to be `https://[...]`.

> That being said, I do support adding an override just in case someone happens to need it.

agreed, might be nice to have this option closer to the other OIDC settings too

Jappie3: see https://github.com/michaelhthomas/jellyseerr/pull/6

tested on my jellyseerr instance, seems to be working fine

stumbled upon something else while doing this though... the current implementation has no support for PKCE - I'm not an OIDC expert, but maybe that's something to consider here

https://www.rfc-editor.org/rfc/rfc7636
https://kanidm.github.io/kanidm/master/frequently_asked_questions.html#why-is-disabling-pkce-considered-insecure

Jappie3: implemented PKCE in https://github.com/michaelhthomas/jellyseerr/pull/6

Tested with Kanidm on my server - it refused to do auth without PKCE, I had to configure `warning-insecure-client-disable-pkce` for the jellyseerr client. Tested with that option removed seems to work with the latest commit.

itsmejoeeey:
> What is the latest :preview-OIDC branch version?
> It seems to be affected https://github.com/fallenbagel/jellyseerr/issues/1787

@AlexTryHarder This branch is older than the `v2.7.3` release that fixed that bug. I believe the preview branch version is near `2.7.0`/`2.7.1`.

src/components/Login/OidcLoginButton.tsx
michaelhthomas: I'm not totally convinced about adding an override for this specific case, since having Jellyseerr configured correctly with proxy support should be sufficient. Better documentation may be good, though. For the PKCE support, I think that's definitely worthwhile, but want to hold off for now since it's a relatively significant change and would require further testing. Once this PR is merged, I would be happy to test and review a follow-up to add PKCE support.

voc0der:
I just have CRON restart the container every 4 hours silently for now..hope this is merged soon !!

v3DJG6GL:
@fallenbagel @gauthier-th
Could you update the preview image? That would be great :)

gauthier-th:
Preview updated with the tag `preview-OIDC`

src/components/Login/OidcLoginButton.tsx
Jappie3: > I'm not totally convinced about adding an override for this specific case, since having Jellyseerr configured correctly with proxy support should be sufficient. Better documentation may be good, though.

Fair enough, and yeah better documenting this option (maybe backlinking from the OIDC documentation) sounds like a good idea

> For the PKCE support, I think that's definitely worthwhile, but want to hold off for now since it's a relatively significant change and would require further testing. Once this PR is merged, I would be happy to test and review a follow-up to add PKCE support.

Jellyseerr isn't the most high-value target for attackers, but one of the main attack vectors regarding OIDC is the use of redirects (https://www.rfc-editor.org/rfc/rfc6819) and PKCE mitigates this. While imo the change isn't that significant (the protocol is explained in 3 pages in https://www.rfc-editor.org/rfc/rfc7636), I wouldn't mind more testing - though I hope that this won't fall through the cracks because of that

hydrian:
There is an issue if the Owner user (Plex Auth)  overlaps with your OIDC username and email. The OIDC user gets a failed login message from the WEBUI but there are no log entries in the jellyseerr logs at the debug log level. 

I suspect it is when Jellyseerr tries to link or does not try to link the OIDC account to the existing account. 

Also see #1915 and https://github.com/fallenbagel/jellyseerr/discussions/1529#discussioncomment-14464627

rajsinghtech:
LGTM!

Anexgohan:
> > @michaelhthomas is there a way to have it automatically go to the OIDC login page (Authentik in my case) whenever that is the only available login provider? I could not find a setting for this
> 
> Such a setting doesnt exist (yet) as far as i have seen. Give them some time.

if you are going through a reverse proxy, then just manually redirect the login page to the OIDC login page.

james-d-elliott:
One of the Authelia maintainers here. Our users have helped contribute a [guide](https://www.authelia.com/integration/openid-connect/clients/jellyseerr/) for this PR.

Was just made aware of this and would love to include a common issues checklist. If you'd like an explanation on any of the items please let me know. I have not interrogated this PR to see if any of these are issues but let me know if you want a formal assessment of either the implementation or if these common issues may need attention.

These are backed by the specification and explicitly stated as a "must":

- All Standard Claims granted via scopes should be retrieved from the UserInfo endpoint (scope affects access tokens, not ID Tokens). The only exceptions to this rule are when using the `claims` parameter, or when using discouraged flows (ID Token Implicit Flow). _Fairly certain_ this PR handles this correctly based on the guide.
- User accounts should never be "linked" using standard claims, they should be expressly and intentionally linked by end-users using the `iss` and `sub` ID Token claims. This is particularly important because the `email` and `preferred_username` claims are entirely mutable by the end-user or provider. These additional claims are useful as hints when logging in for the first time or registering, or otherwise updating accounts.


These are good general practices:

- PKCE in addition to client authentication is a good idea.

Jappie3:
thanks for the input @james-d-elliott 

a PKCE implementation can be found here: https://github.com/michaelhthomas/jellyseerr/pull/6https://github.com/michaelhthomas/jellyseerr/pull/6 (still need to drop that other commit because admittedly it's kinda unnecessary)

been running that on my server ever since writing the code, works fine using Kanidm

thimplicity:
This was not merged yet, correct? Can I still use it with the  preview-OIDC label?

baudneo:
> This was not merged yet, correct? Can I still use it with the preview-OIDC label?

Not yet, I'm in the same boat, care about OIDC more though so im using that preview image. I imagine this will all get merged into the next couple of releases. They seem quite busy with merging codebases, so give them some time :)

kernelb00t:
Hey @michaelhthomas, what do you think about a "Auto-connect" feature? I was thinking about some kind of URL, or option that starts the OIDC flow automatically, like Mailcow's `?iam-sso=1` flag that starts SSO directly. It's useful for Authentik/Keycloak and other homepages. 

1. The admin sets up the homepage element
2. He sets the url to https://seerr.example.com/login?autologin=authentik
3. When the user loads the webpage, it directly executes the login flow associated to the SSO provider.
4. The user is then logged in, without having to push any button.

This is a great compromise between those who do not like individual login pages and want to skip the login page entirely, and those who do want the possibility for the user to login by username/password; while minimizing the complexity of the login process in the codebase.

kaaax0815:
From a security standpoint and to ease implementation of e.g. pkce, wouldn't it make more sense to use a "battle-tested" and certified OIDC implementation like https://www.npmjs.com/package/openid-client?

I am in no way affiliated with https://www.npmjs.com/package/openid-client

Khaaz:
I am so much waiting for this!! This is a requirement for me to be able to properly use Seerr, can't wait :)

jorgecastillodp:
is this kind of on track to be added?

CoenWarmer:
Would love to see this merged!

michaelhthomas:
> From a security standpoint and to ease implementation of e.g. pkce, wouldn't it make more sense to use a "battle-tested" and certified OIDC implementation like https://www.npmjs.com/package/openid-client?
> 
> I am in no way affiliated with https://www.npmjs.com/package/openid-client

When I began this work (almost 4 years ago!), it was based on a similar Overseerr PR, which implemented everything "by hand." I evaluated migrating everything to use a library then, and IIRC considered the Auth0 client library and found it was somewhat coupled to their service and not mature enough for the use-case. Looking at that library now, it seems to have improved significantly, so I may give that another shot. While I did my best to implement within the spec, I would also be more comfortable leveraging a "battle-tested" library than a bespoke implementation. 

As an update on the status of this PR generally, I am considering closing this and splitting it further into pieces, in hopes that it will be more likely to be reviewed in a timely manner (and also easier for me to keep up-to-date). As part of that work, I may explore a new server implementation using such a library instead of a hand-rolled and likely flawed implementation.

michaelhthomas:
> Hey @michaelhthomas, what do you think about a "Auto-connect" feature? I was thinking about some kind of URL, or option that starts the OIDC flow automatically, like Mailcow's `?iam-sso=1` flag that starts SSO directly. It's useful for Authentik/Keycloak and other homepages.
> 
>     1. The admin sets up the homepage element
> 
>     2. He sets the url to https://seerr.example.com/login?autologin=authentik
> 
>     3. When the user loads the webpage, it directly executes the login flow associated to the SSO provider.
> 
>     4. The user is then logged in, without having to push any button.
> 
> 
> This is a great compromise between those who do not like individual login pages and want to skip the login page entirely, and those who do want the possibility for the user to login by username/password; while minimizing the complexity of the login process in the codebase.

This is in fact already possible, though admittedly should be better documented. Your example was already almost correct, but `https://seerr.example.com/login?provider=authentik` should work as described.

coolguy1771:
I'd be willing to take up this work, I have a working prototype I'll push in the next couple of days

adiojoe627:
Is this implemented in seerr?

edbourque0:
> Is this implemented in seerr? 

No, this PR is still a draft.

adiojoe627:
> > Is this implemented in seerr?
> 
> No, this PR is still a draft.

Gotcha thank you for the hard work you guys put in.

cordlord:
I'm willing to donate/pay to get this feature merged/closed by the end of the month. March 20th is the 1 year mark...for this PR. The original issue was opened Jul 2022...

0xSysR3ll:
> I'm willing to donate/pay to get this feature merged/closed by the end of the month. March 20th is the 1 year mark...for this PR. The original issue was opened Jul 2022...
> 

Offering payment or donations does not affect whether or when a PR is merged. This project is volunteer-maintained, and reviews are based on technical merit and available time.

Also, as noted by the original author, this PR is likely to be reworked or split into smaller pieces to improve reviewability and long-term maintainability. Any next steps will follow from that.

Please be respectful of the maintainers' time.

zcraftelite9495:
What are the current plans for integrating OIDC into the new master Seerr project? 

I run my own media service for fam and friends and have used preview-OIDC for quite a while now, I have doubts that migration would be easy.

michaelhthomas:
Closed in favor of #2715

PR #1510, opened 2025-03-21
fix(ui): handle import-from-plex response as array

0xSysR3ll:
#### Description

This PR fixes an issue in the `Plex User Import Modal` where the import process raises the error message 'Something went wrong while importing Plex users.' even when the import is successful.
This happens because the frontend expected the response to be an `object` with a `data` key, but the API returns an `array` directly.
Adjusted the response handling logic to correctly process imported users.

#### To-Dos


gauthier-th (approved):
LGTM

fallenbagel (approved):

PR #1533, opened 2025-03-26
feat: add anidb support to jellyfin scanner

ale183:
#### Description
This PR adds AniDB support in the Jellyfin scanner.

The AniDB ID is used only if there's no other provider id available to ensure it doesn't change the way it works now with TMDB/TVDB.

In some cases it doesn't work correctly, for example if the show is splitted in multiple season while on anidb it's a single season.
I know it's not optimal, but these are quite rare cases and i think it's better than nothing.
#### To-Dos

### DIFF
--- a/server/lib/scanners/jellyfin/index.ts
+++ b/server/lib/scanners/jellyfin/index.ts
@@ -116,7 +116,10 @@ class JellyfinScanner {
           });
           return;
         }
-        const episodes = this.jfClient.getEpisodes(jellyfinitem.Id, season.Id);
+        const episodes = await this.jfClient.getEpisodes(
+          jellyfinitem.Id,
+          season.Id
+        );
         if (!episodes[0]) {
           this.log('No episode found for anidb movie', 'debug', {
             jellyfinitem,
--- a/server/lib/scanners/jellyfin/index.ts
+++ b/server/lib/scanners/jellyfin/index.ts
@@ -116,16 +116,14 @@ class JellyfinScanner {
           });
           return;
         }
-        const episode = (
-          await this.jfClient.getEpisodes(jellyfinitem.Id, season.Id)
-        ).at(0);
-        if (!episode) {
+        const episodes = this.jfClient.getEpisodes(jellyfinitem.Id, season.Id);
+        if (!episodes[0]) {
           this.log('No episode found for anidb movie', 'debug', {
             jellyfinitem,
           });
           return;
         }
-        metadata = await this.jfClient.getItemData(episode.Id);
+        metadata = await this.jfClient.getItemData(episodes[0].Id);
         if (!metadata) {
           this.log('No metadata found for anidb movie', 'debug', {
             jellyfinitem,
--- a/server/lib/scanners/jellyfin/index.ts
+++ b/server/lib/scanners/jellyfin/index.ts
@@ -68,7 +68,7 @@ class JellyfinScanner {
     const mediaRepository = getRepository(Media);
 
     try {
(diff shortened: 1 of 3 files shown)


src/i18n/locale/en.json
gauthier-th, line 1: Please remove these changes, this is not related to this PR.
There are pending i18n that have been forgotten but this is out of scope of this PR, I'll make a PR to fix it.

server/lib/scanners/jellyfin/index.ts
```diff
@@ -91,14 +103,20 @@ class JellyfinScanner {
...
         ).some((MediaStream) => {
           return (MediaStream.Width ?? 0) <= 2000;
         });
       });
 
+      // With AniDB we can have mixed libraries with movies in a "show" library
+      // If we found the movie and we have an anidbId we set the other resolution to true
+      if (anidbId && metadata.Type === 'Series') {
+        hasOtherResolution = true;
+      }
```
gauthier-th: What if the movie in the show library is in 4k?

gauthier-th (changes requested):
Could you please provide a few examples of shows where these changes improve the scan?

src/i18n/locale/en.json
ale183: removed

server/lib/scanners/jellyfin/index.ts
ale183: fixed in the latest commit

ale183:
> Could you please provide a few examples of shows where these changes improve the scan?

All anime shows can be used as an example. If someone uses only AniDB as a metadata provider, they won't receive any results from the scan. These changes are necessary to properly map AniDB shows to the corresponding TVDB/TMDB/IMDB show

server/lib/scanners/jellyfin/index.ts
```diff
@@ -71,8 +73,18 @@ class JellyfinScanner {
         return;
       }
 
+      const anidbId = Number(metadata.ProviderIds.AniDB);
```
fallenbagel: I think it would be better to nullish coalesce this to avoid NaN/0 being assigned to it

fallenbagel (changes requested):

server/lib/scanners/jellyfin/index.ts
ale183: added a nullish coalesce like the other ids (?? null)

server/lib/scanners/jellyfin/index.ts
```diff
@@ -89,6 +101,39 @@ class JellyfinScanner {
...
+        const episode = (
+          await this.jfClient.getEpisodes(jellyfinitem.Id, season.Id)
+        ).at(0);
+        if (!episode) {
+          this.log('No episode found for anidb movie', 'debug', {
+            jellyfinitem,
+          });
+          return;
+        }
+        metadata = await this.jfClient.getItemData(episode.Id);
```
gauthier-th: Just a minor change, but cleaner like this IMO.
```suggestion
        const episodes = this.jfClient.getEpisodes(jellyfinitem.Id, season.Id);
        if (!episodes[0]) {
          this.log('No episode found for anidb movie', 'debug', {
            jellyfinitem,
          });
          return;
        }
        metadata = await this.jfClient.getItemData(episodes[0].Id);
```

gauthier-th (approved):
LGTM

fallenbagel:
Please fix the build issues
@ale183

gauthier-th (approved):

gauthier-th:
@all-contributors please add @ale183 for code

PR #1537, opened 2025-03-27
fix(avatar): fix avatar cache busting by using avatarVersion 

fallenbagel:
#### Description
Previously, avatar caching did not update the avatar when the remote image changed. This PR adds logic to check if the avatar was modified remotely by comparing a remote last-modified timestamp with a locally stored version (avatarVersion). For emby, this is done by comparing the `ETag`.  If a change is detected, the cache is cleared, a new image is fetched, and `avatarVersion` is updated along with `avatarEtag`. Otherwise, the cached image is retained.

#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX


server/lib/imageproxy.ts
```diff
@@ -160,6 +160,10 @@ class ImageProxy {
     this.headers = options.headers || null;
   }
 
+  public getHeaders(): HeadersInit | null {
+    return this.headers;
+  }
+
```
gauthier-th: What's the point of this function?

server/utils/imageHelpers.ts
gauthier-th, line 1: Wouldn't it be better to place these 2 functions inside `avatarproxy.ts` and `auth.ts` respectively since they are not used anywhere else?

server/lib/imageproxy.ts
fallenbagel: to get the headers in a more standard way (afaik)?

server/utils/imageHelpers.ts
fallenbagel: true. Let me change that

server/lib/imageproxy.ts
gauthier-th: Yeah but I don't see it used anywhere

fallenbagel: Oh yeah lol. I forgot I already refactored it... This was a remnant from my many versions of this :sweat_smile:

gauthier-th (approved):

PR #1564, opened 2025-04-06
feat(networking): implement forward auth

### LINKED ISSUE #1221: Support for Forward Auth

ishanjain28:
## Description

This PR adds the Forward Auth feature. It fixes the feedback received in previous attempt [here](https://github.com/fallenbagel/jellyseerr/pull/1341) 

1. A few different header names are allowed but they have to added to the allow list in the code. Supporting completely arbitrary headers was difficult.
2. Forward auth headers are only processed if request came from a trusted proxy and configuring trusted proxy is required to use the forward auth feature.
3. Forward auth can be done in 3 ways. 
    a. Verify the user field against `jellyfinUsername | plexUsername` columns. 
    b. Verify the email field against `email` column. 
    c. Verify both the user and email fields.

- Fixes #1221

## How Has This Been Tested?

This has been tested using the included test suite and manually in different environments. Manual test examples, 

1. With Seerr behind 1 trusted proxy and 2 trusted proxies.
2. With Seerr configured with no trusted proxies
3. Different header names in use and sometimes they were set, sometimes they were not set and I got the expected result every time.

Screenshot 2025-04-06 at 17 42 14

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Forward Authentication (SSO) via trusted proxies with configurable user/email headers and Advanced Network Settings UI (trusted IPv4/IPv6 lists, validation).

* **Documentation**
  * Added guide for configuring Forward Authentication with Traefik and example auth provider setup.

* **Chores**
  * Added runtime IP validation dependency.

* **Tests**
  * Updated end-to-end test covering trusted-proxy/forward-auth and restart modal behavior.

* **Quality**
  * Centralized server-side auth header handling and allowlist for forwarded headers.


ishanjain28:
@fallenbagel @gauthier-th This PR is now ready for review!

Dockerfile
fallenbagel, line 1: This file should not be changed by this PR

fallenbagel (changes requested):
I didn't fully review this yet but I noticed one issue. The Dockerfile. Also in terms of UI, I would recommend putting the forwardAuth stuff to be indented like how http(s) proxy is done:
image

Also I would argue that this should be under the `advanced networking` section as this is not for the normal user. Wdyt about that @gauthier-th

gauthier-th:
> I didn't fully review this yet but I noticed one issue. The Dockerfile. Also in terms of UI, I would recommend putting the forwardAuth stuff to be indented like how http(s) proxy is done:
> image
> 
> Also I would argue that this should be under the `advanced networking` section as this is not for the normal user. Wdyt about that @gauthier-th 

100% agree

ishanjain28:
@fallenbagel @gauthier-th Applied all the suggestions.
1. reverted changes to dockerfile
2. Moved the trusted proxies section, forward auth sections to advanced menu. forward auth depends on trusted proxy. 
3. Indented as requested 
Screenshot 2025-04-07 at 22 36 49

src/components/Settings/SettingsNetwork/index.tsx
```diff
@@ -56,19 +67,85 @@ const SettingsNetwork = () => {
...
+                return ctx.createError({
+                  message: `Invalid IPv4 address: ${address}`,
+                });
+              }
+            } else if (address.indexOf(':') != -1) {
+              if (!Address6.isValid(address)) {
+                return ctx.createError({
+                  message: `Invalid IPv6 address: ${address}`,
+                });
+              }
```
M0NsTeRRR:
```suggestion
            if (!Address4.isValid(address)) {
              return ctx.createError({
                message: `Invalid IPv4 address: ${address}`,
              });
            } else if (!Address6.isValid(address)) {
              return ctx.createError({
                message: `Invalid IPv6 address: ${address}`,
              });
            }
```

src/utils/forwardAuthList.ts
```diff
@@ -0,0 +1,8 @@
+export const ForwardAuthAllowlist = [
+  'Remote-User',
+  'Remote-Email',
+  'Remote-Name',
+  'Cf-Access-Authenticated-User-Email',
+  'X-authentik-username',
```
M0NsTeRRR: Use camel case like the other headers. It's still not mandatory, but let's stay consistent.

server/middleware/auth.ts
```diff
@@ -10,9 +10,29 @@ export const checkUser: Middleware = async (req, _res, next) => {
...
+
+  const ipHeader = req.header('X-Forwarded-For');
+  if (ipHeader) {
+    const lastComma = ipHeader.lastIndexOf(',');
+    if (lastComma != -1) {
+      proxyIP = ipHeader.slice(lastComma + 1).trim();
+    }
+  }
 
+  if (proxyIP && proxyIP.indexOf('.') != -1) {
```
M0NsTeRRR: Since we have `ip-address`, we can use `isValid` instead of `indexOf` to perform proper validation.

src/components/Settings/SettingsNetwork/index.tsx
```diff
@@ -104,6 +186,19 @@ const SettingsNetwork = () => {
           validationSchema={NetworkSettingsSchema}
           onSubmit={async (values) => {
             try {
+              const trustedProxies: { v4: string[]; v6: string[] } = {
+                v4: [],
+                v6: [],
+              };
+              for (let value of values.trustedProxies.split(',')) {
+                value = value.trim();
+                if (value.indexOf('.') != -1) {
```
M0NsTeRRR, line 261: Since we have `ip-address`, we can use `isValid` instead of `indexOf` to perform proper validation.

src/components/Settings/SettingsNetwork/index.tsx
ishanjain28: @M0NsTeRRR Reverted this change. These two are not the same thing. The latter will return invalid ipv4 error for a valid ipv6 address.

src/components/Settings/SettingsNetwork/index.tsx
ishanjain28: @M0NsTeRRR This is not doing validation. It is just for separating v4 and v6 addresses for pretty printing in the text field. We do not have to validate them.

If they are invalid, Right after they are rendered to the input, User will see an address validation error

src/utils/forwardAuthList.ts
ishanjain28: These headers are defaults and come from Authentik, Authelia, Cloudflare.

src/components/Settings/SettingsNetwork/index.tsx
M0NsTeRRR: Ah yes, sorry — I didn't read it carefully. At that point, we can have a single condition: if the IP is neither IPv4 nor IPv6, then we return an error.

server/middleware/auth.ts
ishanjain28: We do not have to do validation here. The only way to reach this code is if the header was set by a trusted proxy and it is assumed to be correct. The only reason we do indexOf is to check if it's a v4 or a v6 address and then search the right store for it.

src/components/Settings/SettingsNetwork/index.tsx
M0NsTeRRR: I agree, but since we have a proper library to validate (detect if you prefer) the IP, why not use it? It makes it clearer what we are trying to achieve.

src/components/Settings/SettingsNetwork/index.tsx
ishanjain28: I wanted the message to more helpful and have the additional context. It should show if an ipv4 address is invalid or an ipv6 address is invalid along with the invalid address itself.

src/utils/forwardAuthList.ts
M0NsTeRRR: Yes, but headers are not case-sensitive. See [RFC 7230, Section 3.2](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2), and you're using them in a function with lowercase.

src/components/Settings/SettingsNetwork/index.tsx
ishanjain28: The ip-address library operates on strings/regex rather than constructing an int or [int;2] and have fast implementations for these operations. This will be hit on every request and I didn't want the needless validation to slow things down.

src/utils/forwardAuthList.ts
ishanjain28: That is very good point! Thank you. 

I'll convert them all to lower case since we are doing that anyway and make it a bit faster

src/components/Settings/SettingsNetwork/index.tsx
M0NsTeRRR: I don't fully agree, as you're trying to guess user input based on `.` and `:`. If they were different inputs, I would agree with you, but no problem, still minor things. :smile:

src/components/Settings/SettingsNetwork/index.tsx
M0NsTeRRR: Understable !

bash9810:
excited for this, would love to setup authentik as the auth for jellyseerr, think this is the way to get that done.

ishanjain28:
Rebased this PR on top of latest develop and fixed a small bug. Ready for review!

ishanjain28:
Rebased the PR on top of latest develop and it is ready for review!

caesay:
Would really love to see this merged. This is probably the only thing preventing me from adopting Jellyseerr. I have a lot of users and could never justify them needing separate auth when I already have a totally unified auth setup with oidc at my reverse proxy.

The author seems to have done everything needed to get this going, and has now rebased this a couple of times. What do we need to do to get a review / merge from a maintainer?

fallenbagel:
> What do we need to do to get a review / merge from a maintainer? 

Patience ;)

grzonu:
> Would really love to see this merged. This is probably the only thing preventing me from adopting Jellyseerr. I have a lot of users and could never justify them needing separate auth when I already have a totally unified auth setup with oidc at my reverse proxy.
> 
> The author seems to have done everything needed to get this going, and has now rebased this a couple of times. What do we need to do to get a review / merge from a maintainer?

And ... here we go again ...

WickedColdfront:
_This pull request has merge conflicts. Please resolve the conflicts so the PR can be successfully reviewed and merged._

Good lord.

fallenbagel:
> _This pull request has merge conflicts. Please resolve the conflicts so the PR can be successfully reviewed and merged._
> 
> Good lord. 

If you dont have anything productive to say in relation to this PR, please refrain from commenting here. Merge conflicts are gonna occur now that we have restarted merging in new features after a feature freeze for the past couple of weeks in order to fix bugs. And this means reviewing open PRs and merging them **including this PR**.

**This is not the only PR that gets merge conflicts and needs to be rebased when new features get merged in as they are reviewed ONE-BY-ONE**.

server/middleware/auth.ts
```diff
@@ -10,9 +10,32 @@ export const checkUser: Middleware = async (req, _res, next) => {
...
+  // address to figure out if it should be trusted
+  let proxyIP = req.socket.remoteAddress;
+
+  const ipHeader = req.header('X-Forwarded-For');
+  if (ipHeader) {
+    const lastComma = ipHeader.lastIndexOf(',');
+    if (lastComma != -1) {
+      proxyIP = ipHeader.slice(lastComma + 1).trim();
+    }
+  }
```
codyps: This code _appears_ to allow the content of a header (`X-Forwarded-For`), which at the point it's checked may-or-may not be from a trusted proxy, to override our idea of what the proxyIP is.

As a result, it seems feasible for a client to provide `X-Forwarded-For: junk,127.0.0.1` (or any ip in `network.trustedProxies`) and thus cause all the headers the client provides to be trusted as if they originated from a proxy.

That said, we're also placing these checks behind the `trustProxy` flag, which means we're already trusting the proxy to some extent (`trustProxy` configures express, see https://expressjs.com/en/guide/behind-proxies.html, jellyseerr is only using the 'true' value, which essentially means "trust that if a request gets to us it has passed through a proxy we trust").

It seems like we've got some confusion going on in the code about what we actually trust and when we trust it.

I'd recommend:

 1. consider keeping `trustProxy` separate, as it currently means "do we trust the X-Forwarded-For header regardless of the `req.socket.remoteAddress`", which may be a different level of trust than "trust the user is who they say they are".
 2. validate `req.socket.remoteAddress` as the proxyIP against the trusted proxy list, do not consider the content of `X-Forwarded-For` (someone _might_ have a use case for validating particular _paths_ of the `remoteAddress` + `X-Forwarded-For`. I'd punt on that).

Alternately:

 1. merge the configuration of trusted proxies so that express is configured with our trusted proxy list (or alternately `trustProxy` is set to true to trust all proxies).
 2. trust input in the same pattern as express (ie: if `trustProxy` is true, trust everything, if we have a list of proxies, validate `req.socket.remoteAddress`. We don't necessarily need to walk the proxy chain the way express does)

Based on the comments in the previous PR, I suspect the first (not alternative) option may be more desirable (because `trustProxy` currently has a lower level of trust associated with it, so using it for forwardAuth effectively changes the meaning of the config option)

gauthier-th: @M0NsTeRRR any opinion on this?

0xSysR3ll: IMO, we shouldn't trust `X-Forwarded-For` header unless the request actually comes from a trusted proxy. To avoid changing the actual meaning of `trustProxy`, I'd go with the first option: validate the socket's remote address, then trust the header if the proxy is trusted. Safer and clearer that way.

M0NsTeRRR: I fully agree with @codyps and @0xSysR3ll :+1: and I'd go with the first option too.

ishanjain28: @0xSysR3ll suggestion is not feasible because you cannot predict a user's deployment setup and validating it against the socket address will fail if the jellyfin and reverse proxy are more than 1 l3 layer apart. It'll break in my setup as an example because I have some stuff for HA. 


@codyps your point is very valid but my assumption is people are using sane reverse proxies in their infra which will validate the x-forwarded-for header and not allow arbitrary junk to the backend applications. If they do, that's a huge bug in the reverse proxy. 

Your suggested scenario will fail if you are using caddy as an example and I am quite sure nginx does the same thing.

It is not the backend applications job to verify this header and the heuristics you suggest will not work for everyone. The right way to avoid this problem is to configure the proxy correctly.

ishanjain28: For more information, see the security considerations section in the RFC that introduced this header, 
https://www.rfc-editor.org/rfc/rfc7239.html#section-8.1


@codyps neither of your 2 suggestions will work.

Either you fully trust the proxy and jellyseerr assumes you configured it correctly and proxy will do it's job OR you don't trust the proxy and you shouldn't be using forward auth. 

Assuming the header may contain junk and validating it against socket address is flaky and then you'll get tickets because it works for some people and doesn't work for others. You just do not have enough information to verify it in anymore detail in jellyseerr instance sitting behind 1 or more proxies.

M0NsTeRRR: > reverse proxy are more than 1 l3 layer apart

Maybe I'm missing something but I don't understand why it would not work.
Proxy <- (Network 1) -> Router <- (Network 2) -> Jellyseerr

In this case req.socket.remoteAddress will be equal to Proxy IP in the Network 1 it would not work only if you use NAT (edge case and I don't see why and where you will do that).

ishanjain28: This assumption does not hold true in more complex setups. I'll give two examples.

1. Running multiple instances of jellyseerr for HA and then using VIPs for routing traffic. The application will be in a different subnet and not see the ingress reverse proxy address in the socket address field. 

2. Applications deployed in k8s setups that use service mesh. Applications see address related to the mesh, not the reverse proxy. 

Sorry I am just repeating myself over and over.. 

Either you trust the proxy and expect it to do its job or you don't trust the proxy and don't use forward auth.

M0NsTeRRR: 1. It's not a supported setup
2. Jellyseerr could trust your Envoy proxy, and your Envoy proxy could trust your reverse proxy, thus maintaining the chain of trust as described in RFC 7239

> Sorry I am just repeating myself over and over..

You don't it's the first time you give a full example. Nobody wants to say yes or no without fully understanding of the situation.

ishanjain28: 1. Jellyseerr does not currently say that anywhere. I don't get how you arrived at that conclusion? The suggested alternative here will break those setups so the correct statment will be that, jellyseerr doesn't care if it breaks deployments like that. 
2. The linkerd/service mesh node talking to the service can change at any time. You would have to add the whole subnet in the trusted proxies list and then check if the socketaddress is in the same subnet. But, the attacker can just add a address from that subnet in the x-forwarded-for header, the check in jellyseerr will pass and you have the same problem again!  You will also have the same problem if attacker finds out your proxy address(that's talking to jellyseerr) and add that as a x-forwarded-for header + some fake values for the forward auth header. The check will once again pass and you have the same problem! 


To reiterate, there is no way to fix the reported problem in jellyseerr. You have no way to know what should be trusted and what shouldn't be trusted and you need to defer that job to the proxy. The proxy knows real client ip, it should be ignoring x-forwarded-for value in incoming request from a untrusted source and setting it.

Forward auth fundamentally relies on trusting the proxy. If you don't trust your proxy to do the right things, you obviously should not be using it. 


Do you see the problem yet?

M0NsTeRRR: 1. It have never been mentionned that it's supported and PR are not reviewed with this feature in mind. If people build infrastructure based on something that as never been supported even if it worked on some setup it can breaks yes but it could be a great enhancement for sure but it's a separate topic.
2.
> The linkerd/service mesh node talking to the service [...]

Thank you for detailing your example. It’s a complex setup you’re describing (and not used by a lot of people here) I haven’t used a service mesh in Kubernetes with Cilium or Istio for example, which is why I’m asking for details. Sorry if it bothers you; that’s not my intention.

> But, the attacker can just add a address from that subnet in the x-forwarded-for header, the check in jellyseerr will pass

To add to my previous answer (let me know if I’m wrong), but are you referring to the sidecar proxy inside the pod? If so, what I meant is that Jellyseerr should trust your Envoy proxy, and Envoy should be configured to trust only your service. This way, you establish a trusted chain.
If not can you point me to a schema or something like that, that describe the architecture you have in mind ?

Also, the code snippet we discus wouldn’t make sense to be set up [there](https://expressjs.com/en/guide/behind-proxies.html) as a custom function ?
 
> To reiterate, there is no way to fix the reported problem in jellyseerr. You have no way to know what should be trusted and what shouldn't be trusted and you need to defer that job to the proxy. The proxy knows real client ip, it should be ignoring x-forwarded-for value in incoming request from a untrusted source and setting it.

Yes I don't speak about that, I agree with you it's the meaning of trusting a proxy.

ishanjain28: Hey @M0NsTeRRR I am travelling for the last few weeks and the next few weeks so missed replying to this. 


1. Agreed, but right now it works perfectly. Introducing this change will break it. 
2. Yess, you could add the sidecar address as trusted but it might break on new deployments if the sidecar picks up a new address. 
3. I am not in agreement of making this a custom function exposed to the user. It overcomplicates things if people have to modify the code in a file after deployment to make it work with their setup and may be a source of vulnerabilities (if you let user provide a custom function through webui).



With all this said consider the following, What's stopping an attacker to send a x-forwarded-for header with the last address the same as your reverse proxy(if they guess it right or some other app leaks it) coupled with a malicious forward auth header ? 

Presumably your proxy is not doing it's job properly and allowing junk, so in the above scenario attacker gets total access. 


If your proxy is stripping and setting the correct x-forwarded-for header, then the suggested change is completely unnecessary

M0NsTeRRR: No problem :)

2. The sidecar will have an address of 127.0.0.1 or ::1 for Seerr app, so I don’t see the issue there unless I’m missing something.
3. Sorry, maybe I wasn’t clear: the link just explains how to configure a custom trust implementation for Express. I was asking if it wouldn’t make more sense to configure it there. (We don't want user to touch it)

> Presumably your proxy is not doing it's job properly and allowing junk, so in the above scenario attacker gets total access.

I agree it’s not a perfect solution, but it’s still better than nothing in some cases. If the attacker isn’t on the same network, a properly configured firewall could block the attempt. Also if the header is not present you are already verifying the socket.remoteAddress in the proxy list.

caesay: I have not been able to follow all of the conversation here, but just wanted to add - The meaning of "enable trusted proxy" should be, the proxy is expected to be the only thing which can reach the service, and we trust the proxy to authenticate the user and strip any sensitive headers set by the user before http requests reach the service (*seerr). This is an incredibly common scenario whether it is an enterprise deployment (e.g. aws vpc, security groups, alb) or a home deployment (docker, Kubernetes, or just a simple firewall rule). 

Analysing `X-Forwarded-For` headers, or even caring about a "Trusted Proxy" whitelist feels to me like total overkill. A proxy whitelist certainly shouldn't be a mandatory setting, because in most of these (very common) deployment scenarios, IP addresses change dynamically. Making a IP address whitelist it a mandatory / core part of this feature would make it useless to most people I'd assume.

Janhouse: Great PR @ishanjain28!
I concur that if you are using forward-auth, you likely already limit access to the service only through the proxy. It makes no sense to engineer something extremely complicated here since in a real-world scenario the trusted proxy is determined not just by and IP addresses, it frequently involves security headers and other mechanisms. And it is not the job of the service to do those checks. Forward auth is supposed to be a simple way to authenticate using external systems.
Just add a message to the user that if enabling it, user should allow access only through the proxy.

server/lib/settings/index.ts
```diff
@@ -547,6 +560,15 @@ class Settings {
         csrfProtection: false,
         forceIpv4First: false,
         trustProxy: false,
+        trustedProxies: {
+          v4: [],
+          v6: [],
+        },
```
brodock, line 631: if forward auth is going to take a while, can we have the trusted proxies list merged in separate? as it is currently, enabling "trustProxy" has security implications as requests can be forged, unless a firewall prevents accessing the instance directly into the 5055 port.

server/middleware/auth.ts
```diff
@@ -21,9 +44,68 @@ export const checkUser: Middleware = async (req, _res, next) => {
...
+          email: emailValue,
+        },
+      ];
+    }
 
+    if (query.length > 0) {
+      user = await userRepository.findOne({
+        where: query,
+      });
+    }
```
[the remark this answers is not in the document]
ishanjain28: This is fine!

src/i18n/locale/en.json
```diff
@@ -1013,41 +1013,6 @@
...
-  "components.Settings.SettingsNetwork.proxySsl": "Use SSL For Proxy",
-  "components.Settings.SettingsNetwork.proxyUser": "Proxy Username",
-  "components.Settings.SettingsNetwork.toastSettingsFailure": "Something went wrong while saving settings.",
-  "components.Settings.SettingsNetwork.toastSettingsSuccess": "Settings saved successfully!",
-  "components.Settings.SettingsNetwork.trustProxy": "Enable Proxy Support",
-  "components.Settings.SettingsNetwork.trustProxyTip": "Allow Seerr to correctly register client IP addresses behind a proxy",
-  "components.Settings.SettingsNetwork.validationApiRequestTimeout": "You must provide a valid timeout value",
-  "components.Settings.SettingsNetwork.validationDnsCacheMaxTtl": "You must provide a valid maximum TTL",
-  "components.Settings.SettingsNetwork.validationDnsCacheMinTtl": "You must provide a valid minimum TTL",
-  "components.Settings.SettingsNetwork.validationProxyPort": "You must provide a valid port",
```
ishanjain28, line 1071: I am not sure why is it removing this ? I ran pnpm run i18n:extract a few times and it seems to be removing this block every time.

ishanjain28:
Rebased PR, squashed every thing into a single commit to make it easier for me to rebase in future and applied suggestions from the 🐰

src/i18n/locale/en.json
ishanjain28: Fixed

server/middleware/auth.ts
kevincox: The current code seems wrong. It is extracting the last element of the `X-Forwarded-For` which would be the client IP in most setups, so why is it being called `proxyIP`? In multi-proxy setups this would be the second-last proxy IP but that also seems like not the value that we want to check?

IMHO there are two reasonable options here:

1. Use a different header and anyone who enables this must set it with the client IP and prevent access from any other sources.
2. Walk backwards from the socket IP then right-to-left via `X-Forwarded-For` the client IP is then the first non-trusted IP.

The current code is doing something else that I don't understand.

src/utils/forwardAuthList.ts
```diff
@@ -0,0 +1,13 @@
...
+// This is required to make forward auth work properly without showing the
+// login page on every page load.
+// The headers must be added in lower case.
+export const ForwardAuthAllowlist = [
+  'remote-user',
+  'remote-email',
+  'remote-name',
+  'cf-access-authenticated-user-email',
+  'x-authentik-username',
+  'x-authentik-email',
```
kevincox: Why is this hardcoded? Shouldn't this be based on the configured auth headers?

ishanjain28: @kevincox The list is presented in the forward auth UI in network settings(See screenshot in the first message) and It covers forward auth headers used by popular identity providers. I preferred this to having users type it manually because I felt this was easier. 

We'll have two options if some new IdP comes out or if people need to enter some other header name. 
1. They can edit the JSON file directly. 
2. We can add that header name in Seerr. 

Usually, the header names are kind of standardized because applications need some consistency and not every self-hostable app offers an option to edit the forward auth header.

server/middleware/auth.ts
ishanjain28: @kevincox The `X-Forwarded-For` RFC specifies this order.

> client, proxy1, proxy2

where proxy2 is closer to the service. See the [wiki page](https://en.wikipedia.org/wiki/X-Forwarded-For)

Historically, there has been some confusion and http servers implementing this incorrectly. New ones like Caddy do it correctly :) 

The current code expects `X-Forwarded-For` to have a value like, 
```
X-Forwarded-For: 10.0.10.5, 10.0.50.3
```
OR
```
X-Forwarded-For: 10.0.50.3
```

In first case, it'll read the address of closest proxy(10.0.50.3), in second case it'll read the whole value after trimming it. That can happen if the proxy itself is sending requests(maybe for active healthchecks?) although it is very very unlikely to hit that path.

kevincox: I am well aware how `X-Forwarded-For` works.

For a lot of people the setup will be a single proxy. So the `X-Forwarded-For` will just contain the client IP. This code will then call that the `proxyIP` then (presumably) fail the trusted subnet check. What is actually intended in this case is to check that `req.socket.remoteAddress` is a trusted IP.

For the example of `X-Forwarded-For: 10.0.10.5, 10.0.50.3` it seems like a very strange case:

1. You ignore the trustworthiest of the nearest proxy.
2. You check the trustworthiness of the second proxy (`10.0.50.3`)

I guess it could be reasonable if you implicitly trust the first proxy (for example you are listening on localhost and trust anything that can connect to localhost) but this is a weird setup, and it would still be better to actually check the trustworthiness of the nearest proxy.

Honestly this really raises the question what is the point of this check at all. It seems to be hardcoded to check the IP of one hop away which seems like a somewhat niche use case (maybe if you are using a global CDN then a local reverse proxy). But I doubt most people care at all about that second hop. If the nearest hop is implicitly trusted it may as well do all checks and clear undesirable headers.

src/utils/forwardAuthList.ts
kevincox: Why not just use this list as a autocomplete and let users use whatever value they need? I see no point restricting it.

> They can edit the JSON file directly.

Can they? This constant is used in the code and I don't see any way for configuration to override it.

https://github.com/seerr-team/seerr/pull/1564/changes/BASE..902c39b47b3fa78fc0ab51656db00e4c9155bf3b#diff-8e0c630fb9bd5297c001b08a3d6810a6cab9a9a05e8745d7c2151ccaecb56093R19

> Usually, the header names are kind of standardized because applications need some consistency and not every self-hostable app offers an option to edit the forward auth header.

In my experience almost all allow editing the header.

server/middleware/auth.ts
ishanjain28: > For a lot of people the setup will be a single proxy. So the X-Forwarded-For will just contain the client IP. 

Wrong. See the documentation, please. In this situation, it'll still have two address. The client ip followed by the proxy's address.

> I guess it could be reasonable if you implicitly trust the first proxy (for example you are listening on localhost and trust anything that can connect to localhost) but this is a weird setup, and it would still be better to actually check the trustworthiness of the nearest proxy.

This is incorrect. The trustworthiness of a proxy in the chain is determined by the proxy downstream to it. i.e. if 10.0.10.5 can be trusted will be determined by 10.0.50.3. The proxy at 10.0.50.3 will discard `X-Forwarded-For` header completely and set it to `{socketRemoteAddress} {proxy's own address}` and if the request came from a trusted source, it'll append it's own address and trust the rest of it is correct. 

The service itself(seerr) can not make the decision if it should trust all the proxies in the chain. The proxies in the chain have to calculate that for their source. 

> Honestly this really raises the question what is the point of this check at all. It seems to be hardcoded to check the IP of one hop away which seems like a somewhat niche use case (maybe if you are using a global CDN then a local reverse proxy). But I doubt most people care at all about that second hop. If the nearest hop is implicitly trusted it may as well do all checks and clear undesirable headers.

Please see existing implementations of how this is done in applications and proxies and read documentation on the header itself. You seem to have a misunderstanding here.

src/utils/forwardAuthList.ts
ishanjain28: I personally do not foresee this as something people may need(specially, since there is an option to edit the config file directly). However, If the maintainers _want_ this now, I'll look into adding it. Otherwise, it can be added in future if/when someone asks for it.

server/middleware/auth.ts
kevincox: From your own source:

> The general format of the field is:
>
> ```
> X-Forwarded-For: client, proxy1, proxy2
> ```
>
> where the value is a comma+space separated list of IP addresses, the left-most being the original client, and each successive proxy that passed the request adding the IP address where it received the request from. In this example, the request passed through proxy1, proxy2, and **then proxy3 (not shown in the header). proxy3 appears as remote address of the request**.
>
> https://en.wikipedia.org/wiki/X-Forwarded-For#Format

Emphasis mine.

ishanjain28: @kevincox Actually you are righgt about this! 

So it needs to look just the socket address and check if it is _trusted_.

kevincox: > So it needs to look the last address in x-forwarded-for

I'm pretty sure that we don't need to look at `X-Forwarded-For` at all. We only need to do that if we want to find the client IP (which I don't think Seerr currently does).

What I would recommend:
1. Check the socket address.
2. Warn users that whatever proxies they mark as trusted are responsible for setting/clearing all of the relevant headers.

docs/extending-seerr/forward-auth.md
```diff
@@ -0,0 +1,77 @@
...
+
+:::warning
+The user has to exist, it will not be created automatically.
+:::
+
+:::info
+If the user has no email set, the username will also work
+:::
+
+## Example with Goauthentik and Traefik
```
M0NsTeRRR: Regarding the authentik documentation, I think it would be more effective to update the existing authentik integration guide available [here](https://github.com/goauthentik/authentik/blob/main/website/integrations/media/jellyseerr/index.md) and add a link here instead, wdyt ?

server/middleware/auth.ts
ishanjain28: > I'm pretty sure that we don't need to look at X-Forwarded-For at all.

Yes, agreed. It took a second for my comment edit to go through. We already have a message to only enable this behind trusted proxies. I'll make the other change.

package.json
```diff
@@ -70,6 +70,7 @@
     "http-proxy-agent": "^7.0.2",
     "https-proxy-agent": "^7.0.6",
     "lodash": "4.17.23",
+    "ip-address": "^10.0.1",
```
M0NsTeRRR: You can bump to 10.1.0

server/middleware/auth.ts
ishanjain28: Thanks again @kevincox for pointing this out. I have fixed this now and also tested it in my setup with 1 proxy and 2 proxies. it was previously working fine before because traffic from public internet goes through two proxies and I had added addresses of both proxies in trusted proxies list in seerr and I never tested it at home where traffic would go through a single proxy.

src/utils/forwardAuthList.ts
M0NsTeRRR: Providing a default list is a good approach, but it’s also important to allow users to specify their own values. This way, we avoid having to review pull requests just to add missing headers, which can delay users who need additional options while waiting for their PR to be merged.

We also want to provide the best user experience, so we should avoid asking users to edit this file for common operations.

So I agree with @kevincox approach

> list as a autocomplete and let users use whatever value they need

docs/extending-seerr/forward-auth.md
ishanjain28: @M0NsTeRRR Makes sense! Is it okay if that goes in a follow up PR once update is merged in Authentik ?

M0NsTeRRR: Yes sure

ishanjain28: @M0NsTeRRR Or honestly, I'll prefer to skip it completely. I don't use traefik and I don't have a setup to test the config and make sure it is correct. That part of the document was carried over from previous 1-2 people who worked on this feature.

src/utils/forwardAuthList.ts
ishanjain28: @M0NsTeRRR Okay, that makes sense! My preference is to keep the PR in the current state and eventually if someone needs it, it can be changed to a auto complete field. How does that sound ? The current list covers forward auth headers used by cloudflare, authentik, authelia and iirc 1 or 2 more IdP

M0NsTeRRR: I would prefer to have it day one 😄

wdyt @seerr-team/seerr-core ?

docs/extending-seerr/forward-auth.md
M0NsTeRRR: > I'll prefer to skip it completely. I don't use traefik and I don't have a setup to test the config and make sure it is correct.

Are you referring to Authentik as well, or just Traefik doc ?

If you can’t test it, we can remove it unless another maintainer can step in. I think @gauthier-th or @0xSysR3ll use Docker setups, so they might have Traefik available.

ishanjain28: > Are you referring to Authentik as well, or just Traefik doc ?

Both. My setup is Authelia / Caddy / LXC Containers so I can't test it. That document was carried over from previous contributors.

src/utils/forwardAuthList.ts
ishanjain28: Sure but I don't really want to go make more changes in that part of the code again :/

docs/extending-seerr/forward-auth.md
M0NsTeRRR: Okay let's see if others maintainers can step in for this part and if not we will remove it.

Would you mind updating this for authelia and link it here as you have a working setup ?

https://www.authelia.com/integration/openid-connect/clients/jellyseerr/
https://github.com/authelia/authelia/blob/master/docs/content//integration/openid-connect/clients/jellyseerr/index.md

ishanjain28: Added docs in trusted header sso section here https://github.com/authelia/authelia/pull/11757

fallenbagel:
@ishanjain28 is this pr now fully ready for review?

ishanjain28:
@fallenbagel All the code, api documentation is ready for review. Need to decide if we want to keep [docs/extending-seerr/forward-auth.md](https://github.com/seerr-team/seerr/pull/1564/files/902c39b47b3fa78fc0ab51656db00e4c9155bf3b#diff-48dadd85083c6dbaccfa7ef2edb314474360403e62067ae3e9225ede9077d724) since I have no way to test it and it was carried over from previous contributors who worked on it.

fallenbagel:
> @fallenbagel All the code, api documentation is ready for review. Need to decide if we want to keep [docs/extending-seerr/forward-auth.md](https://github.com/seerr-team/seerr/pull/1564/files/902c39b47b3fa78fc0ab51656db00e4c9155bf3b#diff-48dadd85083c6dbaccfa7ef2edb314474360403e62067ae3e9225ede9077d724) since I have no way to test it and it was carried over from previous contributors who worked on it.

I would prefer we keep the documentation up-to-date to this implementation since we have a habit of adding new features and not documenting it at all (by we i meant even us, the maintainers, are guilty) 😅

Also I know this PR was created way before our new contributing guidelines but due to the new validations and check requirements to be passed added, would you mind updating the pr description to follow the pr template as described in this comment (see https://github.com/seerr-team/seerr/pull/1564#issuecomment-4228372244). Sorry for the trouble.

I will start reviewing this pr over this weekend in the meantime

ishanjain28:
> I would prefer we keep the documentation up-to-date to this implementation since we have a habit of adding new features and not documenting it at all (by we i meant even us, the maintainers, are guilty) 😅

The general steps for adding forward auth to any application are exactly the same i.e. make sure the proxy sends required headers to the application. The exact steps vary depending on what proxy is in use. Usually, the proxy documentation has a section on this. Example [Caddy](https://caddyserver.com/docs/caddyfile/directives/forward_auth)

Then, IdP also document which headers are available, example [Authelia](https://www.authelia.com/integration/trusted-header-sso/introduction/) and [Authentik](https://docs.goauthentik.io/add-secure-apps/providers/proxy/forward_auth/). 

I have added a section in Authelia -> Integrations -> Trusted Header SSO on steps to enable forward auth in Seerr. (The PR for it is open right now). 

I think we should avoid documenting specifics since those might change over time and it's difficult to keep it updated. Furthermore, people will have different preferences in IdP and proxies and we can't document setting this up on all of them.

server/middleware/auth.ts
```diff
@@ -22,13 +37,63 @@ export const checkUser: Middleware = async (req, _res, next) => {
...
-
     user = await userRepository.findOne({
       where: { id: req.session.userId },
     });
-  }
+  } else if (
+    settings.network.trustProxy &&
+    settings.network.forwardAuth.enabled &&
+    trustedProxy
+  ) {
```
fallenbagel, line 52: There's no validation here to prevent `forwardAuth.enabled = true` with an empty `trustedProxies` list. Wouldn't this be confusing UX since in that case `trustedProxy` will always be false so forward auth will silently do nothing?

server/middleware/auth.ts
fallenbagel, line 1: nit: I would prefer strict equality is used for consistency with the rest of the codebase instead of `==` / `!=`

server/middleware/auth.ts
```diff
@@ -10,9 +10,24 @@ export const checkUser: Middleware = async (req, _res, next) => {
...
+
+  if (socketAddress && socketAddress.indexOf('.') != -1) {
+    trustedProxy =
+      socketAddress == '127.0.0.1' ||
+      settings.network.trustedProxies.v4.includes(socketAddress);
+  } else if (socketAddress) {
+    trustedProxy =
+      socketAddress == '::1' ||
+      settings.network.trustedProxies.v6.includes(socketAddress);
+  }
```
fallenbagel: `indexOf('.')` for IP version detection will misclassify IPv4-mapped IPv6 addresses like `::ffff:127.0.0.1` as they contain both `.` and `:`, so they'll hit the v4 branch but won't match anything in the v4 trusted list. I would suggest using `net.isIPv4()` / `net.isIPv6()` from Node's `net` module instead as that handles mapped addresses correctly. This would be consistent with our codebase too as we use that in:
- https://github.com/seerr-team/seerr/blob/061121c326d416b6d6c126544a65bc2bb848e51f/server/routes/auth.ts#L296-L300
- https://github.com/seerr-team/seerr/blob/061121c326d416b6d6c126544a65bc2bb848e51f/server/routes/user/usersettings.ts#L407-L411

src/utils/forwardAuthList.ts
fallenbagel: > I would prefer to have it day one 😄
> 
> wdyt [@seerr-team/seerr-core](https://github.com/orgs/seerr-team/teams/seerr-core) ?

I agree with the autocomplete approach. The hardcoded list also has a practical issue I believe (unless I'm mistaken?). The `getAuthHeaders` in `serverSidePropsHelpers` only forwards headers in this allowlist during SSR, so any custom header set via `settings.json` won't be forwarded on SSR requests right? Wouldn't that break forward auth on server-rendered pages. So imo, the autocomplete with the current list as suggestions but allowing custom values would fix both the UX and this issue.

docs/extending-seerr/forward-auth.md
```diff
@@ -0,0 +1,77 @@
+# Forward auth
+
+You can use [forward-auth](https://doc.traefik.io/traefik/middlewares/http/forwardauth/) mechanism to log into Jellyseer.
```
fallenbagel:
```suggestion
You can use [forward-auth](https://doc.traefik.io/traefik/middlewares/http/forwardauth/) mechanism to log into Seerr.
```

seerr-api.yml
```diff
@@ -9,6 +9,9 @@ info:
 
     - **Cookie Authentication**: A valid sign-in to the `/auth/plex` or `/auth/local` will generate a valid authentication cookie.
     - **API Key Authentication**: Sign-in is also possible by passing an `X-Api-Key` header along with a valid API Key generated by Seerr.
+
+    Additionally, if forward authentication is enabled, Jellyseerr will authenticate as the user set in the header defined by `forwardAuth.userHeader`.
```
fallenbagel:
```suggestion
    Additionally, if forward authentication is enabled, Seerr will authenticate as the user set in the header defined by `forwardAuth.userHeader`.
```

fallenbagel (changes requested):

fallenbagel:
> I think we should avoid documenting specifics since those might change over time and it's difficult to keep it updated. Furthermore, people will have different preferences in IdP and proxies and we can't document setting this up on all of them.

Yes. So i vouch on removing authentik docs

docs/extending-seerr/forward-auth.md
M0NsTeRRR: I’ll let you add a link for the Authelia configuration.

You can remove the Authentik documentation for now we’ll add it back once it’s updated for Seerr.

For the Traefik part, I’ll wait for input from the other maintainers before making any changes.

src/utils/forwardAuthList.ts
Janhouse: Also having a list of headers looks like a security vulnerability. Any client can pass any header they like. Unless proxy is set up in a way where it filters out all of these, they will be passed to seer from client.
I think you are putting in too much effort to make this feature "beginner friendly". Forward auth will never be a one click setup, so it makes no sense to try to make it like that on seerr side.

ishanjain28: > The getAuthHeaders in serverSidePropsHelpers only forwards headers in this allowlist during SSR, so any custom header set via settings.json won't be forwarded on SSR requests right? Wouldn't that break forward auth on server-rendered pages.

Yess, this is correct. I also vaguely remember this was the reason I went with a allow list instead of allowing anything. The header needs to be copied in `getAuthHeaders`. I am not sure if I can get the settings object there and read the configured header names ? 
I'll check this again

> Also having a list of headers looks like a security vulnerability. Any client can pass any header they like. Unless proxy is set up in a way where it filters out all of these, they will be passed to seer from client.

The config needs to be aligned in the proxy and the application. If the proxy sets `Remote-User` but application is looking at `X-Authentik-User`, then that's a incorrectly configured. That'll also be a problem with completely custom header names.

server/middleware/auth.ts
ishanjain28: @fallenbagel `trustedProxy` can still be true if the request came from `localhost`. `localhost` is implicitly always trusted.

ishanjain28:
@fallenbagel @M0NsTeRRR Soo this is the problem and the reason why I went with a predefined list of headers back when I implemented this.

The headers _have_ to be copied in `serverSidePropHelpers` and I have to know the header names for that. I can not make a `getSettings` call there to read the configured headers. And, I don't want to copy _all_ the headers because that feels inefficient. 

Is there a way to read configured headers there? If yes, I can make it completely user configurable, otherwise I don't see an option but to keep a list of predefined headers.

ishanjain28:
The last commit addresses feedback from @ fallenbagel 

1. stricter comparisons 
2. use net.ipv4 / net.ipv6 for checking address types
3. corrected typos 

I also removed the section on authentik / traefik completely because traefik config revolved around authentik. Since, I can't test it I have no idea what will actually work. 

And, adding completely dynamic header names is complicated because we need to know the header name to copy them in the SSR path. I do not want to copy all headers for all requests. I still believe there are not infinite combinations we need to support since people are using one of the few available IdPs and the current version may be workable. With that said, I am still open to modifying it if someone can help me find a way to read configured header names in `getServerSideProps`.

Janhouse:
> And, adding completely dynamic header names is complicated because we need to know the header name to copy them in the SSR path. I do not want to copy all headers for all requests. I still believe there are not infinite combinations we need to support since people are using one of the few available IdPs and the current version may be workable. With that said, I am still open to modifying it if someone can help me find a way to read configured header names in `getServerSideProps`.

What a nonsense. The pattern is to use functions in middleware/proxy for dynamic header checks (not static patterns). This is what middlewares are ususally used for. And all requests go through middleware (nextjs <=15)/proxy (nextjs 16+) ensuring sufficient security.

I usually combine it with a simple in memory cache object, so it would not try to read it from database on each request. This way you get dynamic flexibility while still maintaining good performance.

server/routes/settings/index.ts
```diff
@@ -91,6 +91,7 @@ settingsRoutes.get('/network', (req, res) => {
 settingsRoutes.post('/network', async (req, res) => {
   const settings = getSettings();
 
+  settings.network.trustedProxies = { v4: [], v6: [] };
```
gauthier-th, line 99: Why would you reset that?

gauthier-th (changes requested):

server/routes/settings/index.ts
ishanjain28: @gauthier-th I don't remember. I don't see any reason why we should do this so I'll revert this and test.

ishanjain28:
@Janhouse Alright then, tell me how do I read the configured headers for forward auth there and I'll look into making the change.

server/routes/settings/index.ts
ishanjain28: @gauthier-th This is required to unset proxies.

The web interface gets all proxies and then sends a response with updated proxies. We clear out existing proxies in the config and then set it to the proxies in received response. 

If we don't clear it, it'll merge the existing set of proxies with the received set and it'll never actually remove the proxy that was removed by the user.

Janhouse:
> @Janhouse Alright then, tell me how do I read the configured headers for forward auth there and I'll look into making the change. 

https://nextjs.org/docs/app/getting-started/proxy

You just import whatever function you need and call it if nodejs runtime is in use.
In nextjs 15 you had to add `export const runtime = 'nodejs'` in middleware but from nextjs 16 proxy uses nodejs runtime by default.

But middleware/proxy runs in a separate context from the rest, meaning you are not supposed to access same singletons from middleware as used in pages. But you can initialize db connection from middleware and you can pass values to pages by setting values in request headers.
You can probably bypass this design restriction in self hosted nextjs since it runs on the same nodejs process anyway, so just export the database connection and import in both places.
If it sill creates new object, create new connection in read-only mode if using sqlite.

I am currently on the phone so can't fully verify but this should be enough info for you to try. 🙂 If you think it is still a problem, I can cereate an example later on.

gauthier-th:
> > @Janhouse Alright then, tell me how do I read the configured headers for forward auth there and I'll look into making the change.
> 
> https://nextjs.org/docs/app/getting-started/proxy
> 
> You just import whatever function you need and call it if nodejs runtime is in use. In nextjs 15 you had to add `export const runtime = 'nodejs'` in middleware but from nextjs 16 proxy uses nodejs runtime by default.
> 
> But middleware/proxy runs in a separate context from the rest, meaning you are not supposed to access same singletons from middleware as used in pages. But you can initialize db connection from middleware and you can pass values to pages by setting values in request headers. You can probably bypass this design restriction in self hosted nextjs since it runs on the same nodejs process anyway, so just export the database connection and import in both places. If it sill creates new object, create new connection in read-only mode if using sqlite.
> 
> I am currently on the phone so can't fully verify but this should be enough info for you to try. 🙂 If you think it is still a problem, I can cereate an example later on.

We don't use Next.js for our backend, we use Express. I provided @ishanjain28 a solution on our Discord to pass the settings object from the Express server to the Next.js getInitialProps using the Request object.

ishanjain28:
Rebased this PR and thanks to huge help from @gauthier-th, I have switched the system from a hardcoded set of header names to allowing anything with autocomplete from the common set. 

screenshot attached and ready for review! 
Screenshot 2026-04-14 at 23 09 14

server/routes/settings/index.ts
gauthier-th: Got it. Could you please add a brief comment to clarify this? It’s not immediately clear at first glance.

gauthier-th:
@Janhouse @kevincox review is welcome whenever you're free :pray:

fallenbagel (changes requested):
@ishanjain28 Sorry for the delayed review on this. Back when this PR was originally opened, we were a small team juggling a lot of PRs and were in the middle of a significant merge (including early discussions with the Overseerr team), so things slipped through the cracks.

> @fallenbagel @gauthier-th Applied all the suggestions.
> 
>     1. reverted changes to dockerfile
> 
>     2. Moved the trusted proxies section, forward auth sections to advanced menu. forward auth depends on trusted proxy.
> 
>     3. Indented as requested
> 
> 
> Screenshot 2025-04-07 at 22 36 49

ref: https://github.com/seerr-team/seerr/pull/1564#issuecomment-2784034354

We're now asking you to partially revert the above. Here's the background:

- The advanced networking menu was removed in https://github.com/seerr-team/seerr/commit/a488f850f3d6f0d9896362db873bd86ec1b7040a#diff-fe3711be57274a5af941ff098722503db9e681aae8db4c0377b110cec36cae1cL384-L423 after we dropped the DNS server setting (#1416) and `forceIpv4First` at that point, there was nothing left to put in an "advanced" section.
- `forceIpv4First` was later re-added, but we decided to keep the networking tab flat and simple rather than reintroducing the advanced section (#1719).
- That decision was made without circling back to this PR as we did not have time to review this PR and since then it was forgotten. Sorry for our oversight, and for the confusion.

Could you restructure the proxy settings back to the following layout?
```
[X] Enable Proxy Support
    └── Trusted Proxies
    └── Forward Authentication
          └── User Header
          └── Email Header
```

Appreciate your patience with the back-and-forth, and sorry for the extra work!

gauthier-th:
If those changes are a too much for you or if don't have enough availability, we can take care of implementing it. We understand we asked for a lot of changes in a long time.

ishanjain28:
@gauthier-th It's no issue, addressed the feedback in the new commit.  :) 

1. Reformatted fw auth autocomplete headers to match how they are specified in the docs of IdPs. 
2. Added comment to explain why we reset trustedProxies in the server before merging local / remote config. 
3. Removed Advanced Network Settings and restructured network settings page. attached screenshot. let me know if the indentation is okay or if I misunderstood structure 

Screenshot 2026-04-15 at 21 34 14

Screenshot 2026-04-15 at 21 34 33

natemccurdy:
Nice work on this — looking forward to using forward auth with Cloudflare Access.

One thing I noticed: I think the forward auth path in `auth.ts` might be the only auth method that doesn't establish a session? The Plex, Jellyfin, and local login routes all write `req.session.userId = user.id` after finding the user, but the forward auth lookup seems to skip that step.

I wonder if that means every API call through forward auth re-runs the user lookup by email/username (non-primary-key fields, with an OR across `jellyfinUsername`/`plexUsername`). With SWR firing multiple API calls per page navigation, that could add up. Would writing a session on first successful lookup let subsequent requests authenticate via the session's primary key lookup instead?

If so, I think the change would be small — after the user is found around line 95 of `auth.ts`:

```ts
if (user && req.session) {
  req.session.userId = user.id;
}
```

Not a blocker at all, just wondering if it'd be a nice consistency and performance improvement.

ishanjain28:
Thanks @natemccurdy, that was a good observation! It was indeed authenticating every request and setting the user id in session avoids a lot of unnecessary work! 

edit: Actually @natemccurdy, this change will introduce a bug. Authelia sets a cookie `_sso`, seerr sets a cookie `connect.sid`. Clear `_sso` cookie and refresh the page. It takes back to the authelia login page and now if you login as a different user, it'll still login as the user you were logged in as before deleting the `_sso` cookie. 


edit: reverted the change since it introduces a bug.

exenza:
is this still be worked on? My network setting do not show those options..

Would you consider as well to allow to disable auth altogether? [Tinyauth](https://tinyauth.app/) can then manage login

ishanjain28:
@exenza It is ready from my side. You need to enable, Enable proxy support and then Enable proxy forward authentication to use this feature.

exenza:
> @exenza It is ready from my side. You need to enable, Enable proxy support and then Enable proxy forward authentication to use this feature.

I rather wait to be in the main branch... are there major issues to merge?

Sapd:
3 issues:

---

`/api/v1/*` returns 401 `cookie 'connect.sid' required` once forward-auth is on, even when the proxy headers are set correctly.

In OpenAPIValidator: cookieAuth is the security scheme on every endpoint and the validator runs before the auth middleware, so requests without a session cookie get rejected before forward-auth even has a chance.

 Fixed like this `server/index.ts`:

  ```diff
   OpenApiValidator.middleware({
     apiSpec: API_SPEC_PATH,
     validateRequests: true,
  +  validateSecurity: false,
   })
```

  isAuthenticated() already does the actual auth check, so the validator's cookie requirement is redundant and it breaks any non-cookie auth path (this PR, OIDC in #2715, future API-key flows). @Jycreyn flagged the
  same thing on #2715.

---

Can we get also automatic user creation?

```
 if (!user && hasUserHeader && userValue) {
    user = userRepository.create({
      jellyfinUsername: userValue,
      email: emailValue || `${userValue}@forward-auth.local`,
      userType: UserType.JELLYFIN,
      permissions: settings.main.defaultPermissions,
    });
    await userRepository.save(user);
  }
```

--- 

 The lookup uses exact SQL equality on jellyfinUsername. That bites in practice because:

  - Jellyfin's AuthenticateByName lowercases the username before storing, so the row is **j**oe
  - Authentik (and most IDPs) preserve original casing in property mappings, so the header arrives as **J**oe (upper case)


  Could the WHERE be case-insensitive? With TypeORM:

```
  qb.where(
    'LOWER(user.jellyfinUsername) = LOWER(:u) OR LOWER(user.plexUsername) = LOWER(:u)',
    { u: userValue }
  );
```

--- 

Edit: I did fixes here, incl auto creation of users: https://github.com/Sapd/seerr/commits/forward-auth-fixes/

**Comment by Xatrekak** [comment]
@M0NsTeRRR I have carried forward the work of @ishanjain28 and @Sapd [HERE](https://github.com/Xatrekak/seerr/commits/add-auto-provisioning-gui) fixing some issues regarding the cloudflare forward auth workflow as well as a bug if the proxy is running a dual stack listener.

If the original author does not return I am willing to put in the work to get this over this finish line whether its fixing any actual issues or just resolving the merge conflicts. 

I have end-to-end tested my current branch and it seems fully functional.

**Comment by ishanjain28** [comment]
I can take a look at this again on upcoming weekend.

**Comment by M0NsTeRRR** [comment]
@Xatrekak feel free to open some pull requests on @ishanjain28 fork. That way, he can review your fix and merge it when he has time this weekend.

**Comment by ishanjain28** [comment]
This has been rebased with changes from @Xatrekak. I also verified it in my instance, it works correctly. Ready for review!

File: server/middleware/auth.ts [hunk 17]
```diff
@@ -22,13 +42,120 @@
...
+          // keep displaying via plex/jellyfinUsername as before.
+          username: userValue ? undefined : emailLocalPart,
+          permissions: settings.main.defaultPermissions,
+          userType: newUserType,
+          // Required NOT NULL column; resolved client-side via Gravatar/avatarproxy.
+          avatar: '',
+        });
+        await userRepository.save(user);
+        logger.info(
+          `Auto-provisioned user via forward-auth: ${provisionUsername}`,
```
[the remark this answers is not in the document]
Xatrekak: provisionUsername is a derived variable from the header value of an already authorized user from an upstream identity provider through a trusted proxy. The code path is only reachable through the forward authentication mechanism and not the typical username/password workflow. This field is not a user-provided value in the sense that an arbitrary value could be passed.

The only intended possible values are the user portion of an email address or the username provided by a remote-user header and in both cases must have already been approved by the admin in the upstream identity provider.

server/middleware/auth.ts
```diff
@@ -22,13 +42,120 @@
...
+          { label: 'Auth', userId: user.id, userType: newUserType }
+        );
+      } catch (e) {
+        logger.error(
+          `Failed to auto-provision forward-auth user ${provisionUsername}`,
```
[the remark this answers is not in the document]
Xatrekak: See previous comment https://github.com/seerr-team/seerr/pull/1564#discussion_r3420044343

paulo-roger:
this is awesome!

pelaxa:
just curious what the hold up is on this one and if there is any way to help push it forward?

SubprimeHardware:
I've been running this branch (the current head, plus fixes) in production behind two proxy paths simultaneously — Cloudflare Access with the email header on one hostname, and a second SSO reverse proxy (Pangolin) that's deliberately *not* in the trusted list on another. That turned out to be a good stress test, and I'd like to feed back what it found.

**Fixes — opened as a PR on the head fork** (per the earlier suggestion to route fixes through @ishanjain28): https://github.com/ishanjain28/jellyseerr/pull/3. The one reviewers here may care most about: SSR page loads launder forward-auth headers through an unconditionally-trusted loopback (`getAuthHeaders` forwards them, `checkUser` trusts `127.0.0.1`), so identity headers arriving over *untrusted* paths are honored on every server-rendered page. In my setup that produced an unescapable `/` ↔ `/login` redirect loop after logout; it's also spoofable in the same direction. Fixed by forwarding the original peer address and evaluating trust against that. The rest: auto-provisioned users were typed after the configured media server (breaking the Plex import's LOCAL→PLEX upgrade path), broken avatars on provisioned users, and nondeterministic matching when the same username exists on different users. Also included, clearly separated and optional: hostname support in Trusted Proxies (docker service names, forward-resolved per request), since compose deployments otherwise have no stable IP to whitelist.

**Separately**, one bug I hit turned out not to be forward-auth-specific at all (`UserContext` yanks in-flight navigations back to `/login` when the `auth/me` revalidation is aborted) — that's https://github.com/seerr-team/seerr/pull/3231 against develop directly.

**One design gap in auto-provisioning worth a maintainer opinion before anyone codes it**: the asserted email is both the identity join key *and* a user-editable profile field. With auto-provision enabled, a user renaming their Seerr email gets a duplicate account on the very next headered request — the IDP still asserts the old address, nothing matches, provisioning fills the gap. (There's no session to cushion it: per-request matching is deliberate here, and correctly so — the earlier session experiment was reverted for real reasons.) Options, roughly in ascending effort:

- Cheap guard: refuse to provision when an existing user already holds the would-be username (in the rename case the derived username is byte-identical to the original provision's, so this catches it precisely), log a warning instead. Fail-closed, no migration.
- Proper fix: bind the asserted identity at provision time in dedicated non-user-editable columns, match those ahead of profile email, backfill on first successful match for existing users. Needs a migration, so it should be a deliberate decision on this PR or a follow-up — not something I wanted to unilaterally bolt on.
- Either way, IDP-side renames remain a documented limitation: header-based auth carries no stable subject claim, so a value join is the best available.

Related, since it came up earlier in this thread: the reverted session-perf idea has a middle path that addresses the Authelia user-switch objection — write the session *and record which header identity minted it*, then invalidate the session whenever the current header disagrees. Happy to prototype that too if there's interest.

Disclosure: diagnosis and patches were done with AI assistance (Claude); I've reviewed everything and it's all running on my instance.

shaypet:
I recently built something similar specifically for Cloudflare authentication, but this approach is much cleaner and more abstract.

One crucial security enhancement missing here is JWT verification. Cloudflare Access allows you to validate the JSON Web Token to ensure requests are genuinely authorized. For reference, you can check how I handled it:

[Auth flow logic](https://github.com/shaypet/seerr/blob/feature-add-cloudflare-autologin-access/server/routes/auth.ts#L59-L128)

[Cloudflare verification utils](https://github.com/shaypet/seerr/blob/feature-add-cloudflare-autologin-access/server/lib/cloudflareAccess.ts)

While my full implementation might be redundant if this PR goes through, I think adding JWT verification is essential. Once this PR is merged, I’d gladly contribute and expand on it to support this security layer and workflow as a Cloudflare specialized case.

More on my approach:
My initial approach used pure auto-login, but I ran into edge cases where users needed the ability to log out and switch to a different account. To solve this, I added a dedicated "Connect with Cloudflare" button:

This button only appears if the Cloudflare headers exist and the integration is enabled.

Clicking it logs the user in based on the Cloudflare headers only after successful JWT verification.

image

While your solution doesn't handle edge cases like logging out and relogging in with a different account (instead of auto-falling back to the CF headers), I'm still seriously considering using it because it's so well-structured.


Disclosure: this comment and all my solution in my fork were done with AI assistance (Claude); I've reviewed everything and it's all running on my instance.

PR #1574, opened 2025-04-09
feat: make chart probes configurable

dguihal:
#### Description

Allow to configure liveness/readiness probes + Add optional startup probe

Additional comment fixes on values.yaml so autostyle is properly working

#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX

### DIFF
--- a/charts/jellyseerr-chart/values.yaml
+++ b/charts/jellyseerr-chart/values.yaml
@@ -23,10 +23,8 @@ probes:
     httpGet:
       path: /
       port: http
-    initialDelaySeconds: 60
   # -- Configure readiness probe (set to null to disable)
   readinessProbe:
-    enabled: true
     httpGet:
       path: /
       port: http
--- a/charts/jellyseerr-chart/values.yaml
+++ b/charts/jellyseerr-chart/values.yaml
@@ -18,16 +18,20 @@ strategy:
 
 # Liveness / Readiness / Startup Probes
 probes:
-  # -- Configure liveness probe (set to null to disable)
-  livenessProbe:
-    httpGet:
-      path: /
-      port: http
-  # -- Configure readiness probe (set to null to disable)
-  readinessProbe:
-    httpGet:
-      path: /
-      port: http
+  # -- Configure liveness probe
+  livenessProbe: {}
+  # initialDelaySeconds: 60
+  # periodSeconds: 30
+  # timeoutSeconds: 5
+  # successThreshold: 1
+  # failureThreshold: 5
+  # -- Configure readiness probe
+  readinessProbe: {}
+  # initialDelaySeconds: 60
+  # periodSeconds: 30
+  # timeoutSeconds: 5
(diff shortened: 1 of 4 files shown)


gauthier-th:
cc @M0NsTeRRR

charts/jellyseerr-chart/values.yaml
```diff
@@ -16,6 +16,25 @@ fullnameOverride: ""
...
+probes:
+  # -- Configure liveness probe (set to null to disable)
+  livenessProbe:
+    httpGet:
+      path: /
+      port: http
+    initialDelaySeconds: 60
+  # -- Configure readiness probe (set to null to disable)
+  readinessProbe:
+    enabled: true
```
M0NsTeRRR: There is no enabled field inside [probe](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.32/#probe-v1-core)

charts/jellyseerr-chart/values.yaml
```diff
@@ -16,6 +16,25 @@ fullnameOverride: ""
...
   type: Recreate
 
+# Liveness / Readiness / Startup Probes
+probes:
+  # -- Configure liveness probe (set to null to disable)
+  livenessProbe:
+    httpGet:
+      path: /
+      port: http
+    initialDelaySeconds: 60
```
M0NsTeRRR: What motivates you to change the default value ?

charts/jellyseerr-chart/values.yaml
```diff
@@ -16,6 +16,25 @@ fullnameOverride: ""
 strategy:
   type: Recreate
 
+# Liveness / Readiness / Startup Probes
+probes:
+  # -- Configure liveness probe (set to null to disable)
```
M0NsTeRRR: Doing so without proper context seems like a really bad practice, you can edit the configuration to fit your needs.

M0NsTeRRR (commented):
Hello,
As it's an enhancement, please bump the minor version not the patch version.
Regards,

charts/jellyseerr-chart/values.yaml
dguihal: Hi, I don't understand your point, doing what ? Disabling probe ? If so it's indeed a bad practice. But I think that it's important to keep users the possibility to explicitly do unclean setups. By default probes remains

M0NsTeRRR: Yes, allowing the probe to be disabled. I don’t see a need for it at the moment, so if you don’t need it, I’d prefer to keep it as is. If someone requests it in the future, we can discuss it then, and it wouldn’t be a breaking change.

dguihal: updated

M0NsTeRRR:
LGTM, thanks for your contribution :)

gauthier-th (approved):

PR #1520, opened 2025-03-24
refactor: switch from Fetch API to Axios

gauthier-th:
#### Description

This PR is a WIP to check if some network-related bugs are fixed when using Axios instead of Fetch API.

#### To-Dos

- [x] Successful build `pnpm build`
- [x] Translation keys `pnpm i18n:extract`
- [ ] Database migration (if required)

PR #1581, opened 2025-04-11
fix(imagecache): fix avatar cache folder creation

fallenbagel:
#### Description
This fixes a regression caused by #1520 which broke the file extension detection thereby breaking avatars. 

Previously, the image caching code derived the file extension using the URL (by splitting on a dot). This worked correctly when using the native fetch API, as the URLs contained proper extensions. However, when switching to axios, some image URLs (especially ones with query parameters) no longer produced a valid extension. For example, URLs for avatar images started returning parts like "com:443/UserImage?UserId=…" as the extension, causing cache file paths to be invalid (leading to ENOENT errors when reading or writing files).

#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX

### DIFF
--- a/server/lib/imageproxy.ts
+++ b/server/lib/imageproxy.ts
@@ -3,6 +3,7 @@ import axios from 'axios';
 import rateLimit, { type rateLimitOptions } from 'axios-rate-limit';
 import { createHash } from 'crypto';
 import { promises } from 'fs';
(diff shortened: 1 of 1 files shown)


fallenbagel:
Just rebased to develop

gauthier-th (approved):
LGTM

Issue #480, opened 2023-09-16
Ability to send notifications for new episodes of ongoing seasons


russjr08:
### Description

I've noticed when requesting shows with Jellyseer, that if I request a TV show season that has already aired all of its episodes, I'll get a notification when everything is downloaded as expected. However, when requesting a season that is still actively airing new episodes, you won't receive notifications until the full season has aired, and the last episode has been downloaded.

### Desired Behavior

It would be nice if it were possible to enable an option to get a notification that an episode was downloaded/imported, even if the full season hasn't been downloaded yet (because its still actively airing). Ideally though, Jellyseer would be able to *only* do this when a newly aired episode has just been imported (so that you don't request a season of say 26 episodes that have all aired, resulting in 26 notifications as they all import).

This could even be extended to have a "series partially available" notification option. As an example, lets say three weeks ago I requested Foundation (the last episode of the season only just aired this last week), once all episodes that had aired at that point were imported, a notification could be sent. After that point, it could send a notification each week when the next episode is downloaded.

### Code of Conduct

- [X] I agree to follow Overseerr's Code of Conduct

ojmaster:
bump, i'd like this aswell

martindacost:
Same here ! Would be cool to have notification when new episode comes out and it's available on the media server, like movies

SpiderMan-ParkerTalley:
I'm definitely interested in something like this. A way to receive notifications about new content added from stuff on your watchlist would also be greatly appreciated.

Vexz00:
I just started using Jellyseerr and am very baffled that this isn't even an option. I thought this would be the default. Without that feature it feels like this software only does 50% of what I want. I hope this feature will be added soon, but I have my doubts, considering how long this feature request has been open.

fallenbagel:
> but I have my doubts, considering how long this feature request has been open.

If this was not a feature that is planned, the feature request would be closed as not planned. And if you check one message above, there already is a **draft** PR opened to implement this exact feature. Currently there are discussions on how to implement this properly so it does not end up bulk spam.

Vexz00:
Fair point. At least Sonarr can send notifications for single file imports. Should be a good workaround for now.

Sdhb1:
There seem to be many duplicates of this very request. Can someone confirm if it is on the roadmap? All requests seems to link back to one another. I'm not clear if this has or will be availible in jellyseer or overseer. Ty

fallenbagel:
[...]

PR #1589, opened 2025-04-14
feat(notifications): adds a NEW_EPISODE notification for Jellyfin

parkuman:
This feature adds a new NEW_EPISODE notification type and a trigger specifically built for Jellyfin. It fires during the "recently added" scan of Jellyfin when Jellyfin returns an episode in the newly added array. 

 - it will only fire notifications for seasons in the partiall available status that have new episodes 
 - it only detects new episodes that Jellyfin returns as new episodes

re #480

#### Description
 - updates the `MediaSubscriber.ts` class to trigger this new notification type before an update the the Media entity
 - adds the new notification setting in the 

discord example:
image

example log for the first fire (in dev mode):
```plaintext
2025-04-14T20:10:04.541Z [info][Jellyfin Sync]: Beginning to process recently added for library: TV Shows
2025-04-14T20:10:04.753Z [info][Jellyfin Sync]: New standard episode detected for The Last of Us S2E1
2025-04-14T20:10:04.756Z [debug][Notifications]: Checking for new episodes {"mediaId":47,"is4k":false}
2025-04-14T20:10:04.756Z [debug][Notifications]: Partially available seasons {"mediaId":47,"is4k":false,"partiallyAvailableSeasons":[2]}
2025-04-14T20:10:04.756Z [debug][Notifications]: Detected lastSeasonChange update {"mediaId":47,"is4k":false,"oldTimestamp":"2025-04-14T20:00:04.719Z","newTimestamp":"2025-04-14T20:10:04.753Z"}
2025-04-14T20:10:04.764Z [debug][Notifications]: Found requests for media {"mediaId":47,"is4k":false,"requestCount":1,"requestIds":[3]}
2025-04-14T20:10:04.798Z [debug][Notifications]: Checking request for matching seasons {"mediaId":47,"is4k":false,"requestId":3,"requestedSeasons":[1,2],"partiallyAvailableSeasons":[2],"matchFound":true}
2025-04-14T20:10:04.798Z [info][Notifications]: Sending new episode notification {"mediaId":47,"is4k":false,"requestId":3,"userId":1,"seasons":[2]}
2025-04-14T20:10:04.798Z [info][Notifications]: Sending notification(s) for EPISODE_AVAILABLE {"subject":"The Last of Us (2023)"}
2025-04-14T20:10:04.798Z [debug][Notifications]: Sending Discord notification {"type":"EPISODE_AVAILABLE","subject":"The Last of Us (2023)"}

```

example log if the notification has already been fired:
```plaintext
2025-04-14T20:30:04.497Z [info][Jellyfin Sync]: Beginning to process recently added for library: TV Shows
2025-04-14T20:30:04.608Z [debug][Jellyfin Sync]: Skipping notification - show The Last of Us was recently notified (0 hours ago)
```

#### To-Dos

- [x] fire new episode notification for jellyfin
- [ ] handle full scans?
- [ ] plex

#### Issues Fixed or Closed

- implements #480

### DIFF
--- a/server/lib/notifications/agents/discord.ts
+++ b/server/lib/notifications/agents/discord.ts
@@ -138,6 +138,10 @@ class DiscordAgent
           color = EmbedColors.GREEN;
           status = 'Available';
           break;
+        case Notification.EPISODE_AVAILABLE:
+          color = EmbedColors.BLUE;
+          status = 'New Episode Available';
+          break;
         case Notification.MEDIA_DECLINED:
           color = EmbedColors.RED;
           status = 'Declined';
--- a/server/lib/notifications/agents/email.ts
+++ b/server/lib/notifications/agents/email.ts
@@ -102,6 +102,11 @@ class EmailAgent
             is4k ? 'in 4K ' : ''
           }is now available:`;
           break;
+        case Notification.EPISODE_AVAILABLE:
+          body = `A new episode of the following ${mediaType} ${
+            is4k ? 'in 4K ' : ''
+          }is now available:`;
+          break;
         case Notification.MEDIA_DECLINED:
           body = `Your request for the following ${mediaType} ${
             is4k ? 'in 4K ' : ''
--- a/server/lib/notifications/agents/gotify.ts
+++ b/server/lib/notifications/agents/gotify.ts
@@ -72,6 +72,9 @@ class GotifyAgent
         case Notification.MEDIA_AVAILABLE:
           status = 'Available';
           break;
+        case Notification.EPISODE_AVAILABLE:
+          status = 'New Episode Available';
+          break;
         case Notification.MEDIA_DECLINED:
           status = 'Declined';
           break;
--- a/server/lib/notifications/agents/pushbullet.ts
+++ b/server/lib/notifications/agents/pushbullet.ts
@@ -69,6 +69,9 @@ class PushbulletAgent
         case Notification.MEDIA_AVAILABLE:
           status = 'Available';
           break;
+        case Notification.EPISODE_AVAILABLE:
+          status = 'New Episode Available';
+          break;
         case Notification.MEDIA_DECLINED:
           status = 'Declined';
           break;
--- a/server/lib/notifications/agents/pushover.ts
+++ b/server/lib/notifications/agents/pushover.ts
@@ -109,6 +109,9 @@ class PushoverAgent
         case Notification.MEDIA_AVAILABLE:
           status = 'Available';
           break;
+        case Notification.EPISODE_AVAILABLE:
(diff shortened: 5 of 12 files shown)


parkuman:
> Hey!
> 
> I had idea of solving the issue myself, but never got around to it so I had a look how you made it. Since the feature request mentioned ongoing seasons, my implementation idea was that whenever new episode is detected in Jellyfin, the TMDB would be queried for the release date of that episode and if it is after season/media request date, send notification. But some shows release whole(or half) season at once so it would have been quite spammy.
> 
> I noticed some potential issues in your code tho. Note that I may be wrong about those, since I never actually tested the code, just thought it through.
> 
>     1. If multiple seasons are requested at the same time and their episodes start getting downloaded, only one notification for random episode would be sent out every 12h.
> 
>     2. If first episode of the season arrives, the existing code updates the seasonLastUpdated date(since the season is newly created) and the notification wouldn't be sent out for episode, since 12h has not passed since last update.
> 
>     3. If last episode of the season arrives, the existing code updates the season status to AVAILABLE and no episode notification is sent out, since notifications are sent out only for PARTIALLY_AVAILABLE seasons
> 
>     4. If multiple episodes are released for same season at the same time, only some episodes would get notification due to 12h notification limit.

After messing around with my approach and considering all the points you've brought up I think we need a better solution! I was laser focused on the one scenario, but you're right there's many places where this won't work.

I like your approach by checking the release date, though. I think it would cover the main use case which is notifications for episodes of shows that are actively airing, rather than notifications just for when new episodes get pulled into Jellyfin (which would get very spammy). In the scenario where a whole season drops at once, we could try to batch the notification and list all the new episodes that got added in one notification so it isn't all that spammy.

I think it could also be quite cool to be able to opt in / opt out of notifications on a per-series basis. Example:
 - I enable notifications for new episodes of actively airing series episodes 
 - I request a show that is actively airing
 - by default I'll get notifications for this
 - I can opt out if I want
 
Would also be useful to track this on a per-user basis, since Jellyseerr has notifications server-wide and per-user. What do you think about this? Would definitely take much more work but I think would be a more complete feature.

Seems to me like we might need a new entity to for Episodes 👀 Thoughts?

gauthier-th:
Hey, sorry for the late answer.

I think we should either close this or rewrite it once #1671 is implemented.
It will be much more clean and easier to implement notification for each episode once the logic for episode availability is implemented.

parkuman:
> Hey, sorry for the late answer.
> 
> I think we should either close this or rewrite it once #1671 is implemented. It will be much more clean and easier to implement notification for each episode once the logic for episode availability is implemented.

Agreed! Happy to close, this solution requires indirect and sort of hacky ways to determine things.

PR #1592, opened 2025-04-16
chore: merge upstream

gauthier-th:
#### Description

Pull changes from upstream/develop

#### To-Dos

- [x] Successful build `pnpm build`
- [x] Translation keys `pnpm i18n:extract`
- [ ] Database migration (if required)

PR #1606, opened 2025-04-22
fix: handle pending media removal after request deletion (PGSQL)

fallenbagel:
#### Description
This fixes **more TypeORM lifecycle hook issues** for PostgreSQL. The current behavior is that after a media request is removed (either by cancellation or deletion), the media remains in a pending state unless a user with management permissions clears the media data in the PostgreSQL database. With this change, the media status will be changed to `UNKNOWN` upon the removal of the media request.

Though the `@AfterRemove() hook` in MediaRequest works fine in SQLite it doesn't reliably update the media status in PostgreSQL which is likely due to transaction timing or relation loading differences between the two databases.

#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX

### DIFF
--- a/server/routes/request.ts
+++ b/server/routes/request.ts
@@ -569,6 +569,31 @@ requestRoutes.delete('/:requestId', async (req, res, next) => {
 
     await requestRepository.remove(request);
 
+    const mediaRepository = getRepository(Media);
+
+    const media = await mediaRepository.findOne({
+      where: { id: request.media.id },
+      relations: { requests: true },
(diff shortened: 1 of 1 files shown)


gauthier-th (approved):
LGTM

fallenbagel:
@gauthier-th This needs testing on approved/rejected requests deletion (not just cancellation) once #1592 is merged

fallenbagel:
Superseded by #1631

PR #1615, opened 2025-04-28
feat(userlist): add sortable columns to User List

### LINKED ISSUE #1562: Make Columns Sortable on the Users Tab
### Description

It would be great to have sortable columns on the Users tab

- Clicking on the "Users" column should sort users alphabetically (A–Z / Z–A).
- Clicking on the "Requests" column should sort users by the number of requests (most to least / least to most).
- The same functionality should apply to "Type", "Role", and "Joined", allowing sorting by each respective value (e.g., newest to oldest for "Joined").

### Desired Behavior

I would like the columns on the Users tab to be sortable. Clicking on a column header (e.g., Users, Requests, Type, Role, Joined) should sort the list based on the values in that column.

For example:

- Clicking Users sorts alphabetically (A–Z / Z–A)
- Clicking Requests sorts by number of requests (highest to lowest / lowest to highest)
- Clicking Joined sorts by registration date (newest to oldest / oldest to newest)
- Similar sorting behavior for Type and Role

### Additional Context

_No response_

### Code of Conduct

0xSysR3ll:
#### Description

Added sortable columns to the User List table.

- Clicking on any column header (Users, Requests, Type, Role, Joined) sorts the table by that column.
- Clicking the same column again toggles between ascending and descending order.
- Column headers display an arrow indicating the current sort direction.

**Example with username sorting (ASC)**
image

**Example with username sorting (DESC)**
image

#### To-Dos

#### Issues Fixed or Closed

- Fixes #1562

## Summary by CodeRabbit

* **New Features**
  * User list: bidirectional sorting with direction indicators, header toggle, persisted choice; added account type and user role as sortable fields and updated header labels/accessibility.

* **Tests**
  * Added end-to-end tests for column header sorting, direction toggling, request params, and row order; adjusted intercept patterns.

* **Chores**
  * API docs updated to accept a sortDirection parameter and additional sort keys (usertype, role).


0xSysR3ll:
Oops, think I broke the cypress test ? I looked at the videos/screenshots and tests seem OK (visually speaking) though.

fallenbagel:
> Oops, think I broke the cypress test ? I looked at the videos/screenshots and tests seem OK (visually speaking) though.

It's okay, it was just a hitch. Don't worry about it ;)

fallenbagel:
> I removed the old dropdown sorting menu to simplify the UI since column sorting is more intuitive.

I would say it's more preferable that, that is kept too. It's especially useful on mobile view than using the column to do so imho.

Wdyt @gauthier-th?

0xSysR3ll:
> > I removed the old dropdown sorting menu to simplify the UI since column sorting is more intuitive.
> 
> 
> 
> I would say it's more preferable that, that is kept too. It's especially useful on mobile view than using the column to do so imho.
> 
> 
> 
> Wdyt @gauthier-th?

Yes you're right about the mobile view. I didn't consider how this would be less user-friendly on small screens.
Should we restore the dropdown as it was before, or we should enhance it to include the new sort options (usertype, role) that I added to the column headers?​​​​​​​​​​​​​​​​

gauthier-th:
> > I removed the old dropdown sorting menu to simplify the UI since column sorting is more intuitive.
> 
> I would say it's more preferable that, that is kept too. It's especially useful on mobile view than using the column to do so imho.
> 
> Wdyt @gauthier-th?

If it's enough user-friendly and mobile-friendly why would we keep it?

fallenbagel:
> > > I removed the old dropdown sorting menu to simplify the UI since column sorting is more intuitive.
> > 
> > I would say it's more preferable that, that is kept too. It's especially useful on mobile view than using the column to do so imho.
> > 
> > Wdyt @gauthier-th?
> 
> If it's enough user-friendly and mobile-friendly why would we keep it?

I think you misread @0xSysR3ll .
They are saying we should keep it because its user-friendly and mobile-friendly. The dropdown along with the sort column. Because the sort column thing is not mobile-friendly per se

gauthier-th:
> > > > I removed the old dropdown sorting menu to simplify the UI since column sorting is more intuitive.
> > > 
> > > 
> > > I would say it's more preferable that, that is kept too. It's especially useful on mobile view than using the column to do so imho.
> > > Wdyt @gauthier-th?
> > 
> > 
> > If it's enough user-friendly and mobile-friendly why would we keep it?
> 
> I think you misread @0xSysR3ll . They are saying we should keep it because its user-friendly and mobile-friendly. The dropdown along with the sort column. Because the sort column thing is not mobile-friendly per se

I think i read it properly. I guess I don't understand why would clicking on a column would be not mobile-friendly?

gauthier-th:
> > > > > > I removed the old dropdown sorting menu to simplify the UI since column sorting is more intuitive.
> > > > > 
> > > > > 
> > > > > I would say it's more preferable that, that is kept too. It's especially useful on mobile view than using the column to do so imho.
> > > > > Wdyt @gauthier-th?
> > > > 
> > > > 
> > > > If it's enough user-friendly and mobile-friendly why would we keep it?
> > > 
> > > 
> > > I think you misread @0xSysR3ll . They are saying we should keep it because its user-friendly and mobile-friendly. The dropdown along with the sort column. Because the sort column thing is not mobile-friendly per se
> > 
> > 
> > I think i read it properly. I guess I don't understand why would clicking on a column would be not mobile-friendly?
> 
> I would prefer having two options. And its not mobile-friendly because column does not fit the mobile already. Meaning to change the sorting you have to scroll sideways whereas its much easier to just choose from a dropdown the minute the page loads instead of waiting for page to load then scroll to the right to find the column you want

Yeah i don't totally agree but i get your point. Let's make 2 different things for desktop and mobile then?

fallenbagel:
> > > > > > > I removed the old dropdown sorting menu to simplify the UI since column sorting is more intuitive.
> > > > > > 
> > > > > > 
> > > > > > I would say it's more preferable that, that is kept too. It's especially useful on mobile view than using the column to do so imho.
> > > > > > Wdyt @gauthier-th?
> > > > > 
> > > > > 
> > > > > If it's enough user-friendly and mobile-friendly why would we keep it?
> > > > 
> > > > 
> > > > I think you misread @0xSysR3ll . They are saying we should keep it because its user-friendly and mobile-friendly. The dropdown along with the sort column. Because the sort column thing is not mobile-friendly per se
> > > 
> > > 
> > > I think i read it properly. I guess I don't understand why would clicking on a column would be not mobile-friendly?
> > 
> > I would prefer having two options. And its not mobile-friendly because column does not fit the mobile already. Meaning to change the sorting you have to scroll sideways whereas its much easier to just choose from a dropdown the minute the page loads instead of waiting for page to load then scroll to the right to find the column you want
> 
> Yeah i don't totally agree but i get your point. Let's make 2 different things for desktop and mobile then?

Or why not keep both. It doesn't make it less friendly? You get both options then

0xSysR3ll:
I actually agree with both of you. I should have checked the mobile version earlier.
For now, let’s keep both options available, and we can revisit this later, especially when the mobile app is "introduced".

gauthier-th:
@0xSysR3ll please fix the cypress tests, as they are failing

0xSysR3ll:
> @0xSysR3ll please fix the cypress tests, as they are failing

Could this Cypress error be just a fluke? It has happened before. I reviewed the video of the failing test but couldn’t find anything in my latest commit that would have triggered it.

fallenbagel:
> > @0xSysR3ll please fix the cypress tests, as they are failing
> 
> Could the cypress error be just a fluke? It has happened before. I reviewed the video of the failing test but couldn’t find anything in my latest commit that would have triggered it.
> 

Yes dont worry

gauthier-th:
@all-contributors please add @0xSysR3ll for code

gauthier-th:
@0xSysR3ll sorry about that, you were right to ask :)

Gauvino:
Hello, can a preview-tag can be created for this please ?

gauthier-th:
> Hello, can a preview-tag can be created for this please ?

`preview-sort-userlist` available in ~30min

Gauvino:
Works really well, only thing I'm saying is why is it refreshing when clicking on a column for now I'm having like 25 users and on the bottom I've selected to show 50 items

0xSysR3ll:
> Works really well, only thing I'm saying is why is it refreshing when clicking on a column for now I'm having like 25 users and on the bottom I've selected to show 50 items

Not sure to understand your concern but yeah, the table refreshes on sort because we’re fetching sorted data from the backend every time you click a column. Even if there are fewer users than the page size, it always does a new API call to keep things consistent and scalable. For small lists it’s a bit overkill, but that’s just how it’s set up right now. If we wanted to avoid the reload for small datasets, we’d have to add some client-side sorting logic.

Gauvino:
As you want, just a little review on my user side !

0xSysR3ll:
> As you want, just a little review on my user side !

Well, I’ll have to wait for the bosses’ opinion 🫡
But thanks for your review anyway!

Gauvino:
Would love to see this implemented

0xSysR3ll:
> Would love to see this implemented

Following your review, I’ve moved the sorting logic to the client side to avoid unnecessary API requests and page reloads. Now, users are only fetched when the page number or maximum user count changes.

Gauvino:
> > Would love to see this implemented
> 
> Actually, following your review, I've just changed the sorting to happen client-side, so it should be instant now without any page refreshes. The data is only fetched when changing the number of user displayed or pages.

thank you that was a good idea, can an admin make a preview of this pr to test all the changes ?

gauthier-th:
> > > Would love to see this implemented
> > 
> > 
> > Actually, following your review, I've just changed the sorting to happen client-side, so it should be instant now without any page refreshes. The data is only fetched when changing the number of user displayed or pages.
> 
> thank you that was a good idea, can an admin make a preview of this pr to test all the changes ?

`preview-sort-userlist` will be updated in 30min

Gauvino:
Another review after testing the changes, really good work it's now more fluent when clicking on sorting or so thanks for taking my review into consideration.

By default, when going on the users' pages, it should sort by demanding and not username, but maybe I'm wrong, as it was before was good

Others you integrate are really good and working perfectly. Tested everything I could, even spamming, and absolutely no problem

Really good job !

0xSysR3ll:
> Another review after testing the changes, really good work it's now more fluent when clicking on sorting or so thanks for taking my review into consideration.
> 
> 
> 
> By default, when going on the users' pages, it should sort by demanding and not username, but maybe I'm wrong, as it was before was good
> 
> 
> 
> Others you integrate are really good and working perfectly. Tested everything I could, even spamming, and absolutely no problem
> 
> 
> 
> Really good job !

Thank you again for your review!
You're right about one thing: default sort has changed. I mistakenly changed it from `displayname` (which was default before) to `created`.
I will fix this tonight.

0xSysR3ll:
Think I read a bit too fast the changes.
The default sort remains `displayname` (alphabetical), but server-side data is fetched by `created`  because it's the most efficient sort at the database level.

So I guess I will keep it that way.

Gauvino:
Can this be reviewed, to me merged later, really waiting for this feature and another time thanks for the job @0xSysR3ll !

gauthier-th:
Really sorry if this comes a bit too late :sweat_smile: 

> Following your review, I’ve moved the sorting logic to the client side to avoid unnecessary API requests and page reloads. Now, users are only fetched when the page number or maximum user count changes.

I get the efficiency argument, but IMHO rewriting 2 times the same logic to sort the items is not a very good idea. It's not that bad to have a small delay when refreshing the item to change the sort column.

Gauvino:
> Really sorry if this comes a bit too late 😅
> 
> > Following your review, I’ve moved the sorting logic to the client side to avoid unnecessary API requests and page reloads. Now, users are only fetched when the page number or maximum user count changes.
> 
> I get the efficiency argument, but IMHO rewriting 2 times the same logic to sort the items is not a very good idea. It's not that bad to have a small delay when refreshing the item to change the sort column.

But if jellyseerr can have a smoother UI navigation why would it be a problem, do you see anything wrong with the logic implemented in this PR ?

gauthier-th:
> But if jellyseerr can have a smoother UI navigation why would it be a problem, do you see anything wrong with the logic implemented in this PR ?

I don't see anything wrong with a smoother UI, but I do see an issue with the same logic for the exact same purpose being written twice in two different places

0xSysR3ll:
> I don't see anything wrong with a smoother UI, but I do see an issue with the same logic for the exact same purpose being written twice in two different places

What do you think we should do about this then ? Should I bring back the server side filtering or try and improve the current implementation and make it more DRY ?

gauthier-th:
> > I don't see anything wrong with a smoother UI, but I do see an issue with the same logic for the exact same purpose being written twice in two different places
> 
> What do you think we should do about this then ? Should I bring back the server side filtering or try and improve the current implementation and make it more DRY ?

Yes please. I think it's for the best. Sorry for the repeated changes.

0xSysR3ll:
> > What do you think we should do about this then ? Should I bring back the server side filtering or try and improve the current implementation and make it more DRY ?
> 
> Yes please. I think it's for the best. Sorry for the repeated changes.

Just reverted to the old implementation :)

gauthier-th (commented):
Two small comments, otherwise LGTM:
- Shouldn't we sort by user creation date by default instead?
- The UI of the `<th>` is not really clean: since we already have the `cursor-pointer`, I don't think we need to change the background color when hovering. But even if we keep the background color change, the `transition-colors` is missing and I find the "gap" between `bg-gray-500` and `bg-gray-700` too big.

0xSysR3ll:
> Shouldn't we sort by user creation date by default instead?

Yep, makes sense.

> The UI of the <th> is not really clean: since we already have the cursor-pointer, I don't think we need to change the background color when hovering. But even if we keep the background color change, the transition-colors is missing and I find the "gap" between bg-gray-500 and bg-gray-700 too big.

True true.  I'll just remove the transition background then, since we already have a cursor + chevrons.

src/components/UserList/index.tsx
```diff
@@ -110,13 +110,13 @@ const UserList = () => {
...
   const { user: currentUser, hasPermission: currentHasPermission } = useUser();
-  const [currentSort, setCurrentSort] = useState<Sort>('displayname');
+  const [currentSort, setCurrentSort] = useState<Sort>('created');
   const [currentPageSize, setCurrentPageSize] = useState<number>(10);
 
   const page = router.query.page ? Number(router.query.page) : 1;
   const pageIndex = page - 1;
   const updateQueryParams = useUpdateQueryParams({ page: page.toString() });
-  const [sortDirection, setSortDirection] = useState<SortDirection>('asc');
+  const [sortDirection, setSortDirection] = useState<SortDirection>('desc');
```
gauthier-th: Better to sort in ascending order? So that old accounts are showed first? Like it is right now.

0xSysR3ll: Yeah, so apparently my brain thought DESC = latest. Bold assumption. Very wrong. Don't code at night 🗿

gauthier-th (approved):

Gauvino:
I'm very excited for this to be merged, it's very important in my case!
I already reviewed it; if you need anything from me, like testing the preview, or so tell me!

fallenbagel (approved):

PR #1619, opened 2025-04-30
fix(mediarequest): refactor to .save() instead of .update() for TypeORM

### LINKED ISSUE #1618: Request not visible in /requests?filter=all
### Description

When a user make a request, it is automatically approved but I can't find it in requests. I need to search the movie to see that it has been requested.

See screenshot for example.

No error in log, only info for notification.

### Version

develop (c3b8574515de4f5a2c13f36b3a00708d0858966e)

### Steps to Reproduce

1. A user make a request
2. An admin goes to requests page

### Screenshots

Image

Image

Image

Image

Image

### Logs

```shell

```

### Platform

desktop

### Database

SQLite (default)

### Device

Dell Laptop

### Operating System

Windows 11

### Browser

Chrome

### Additional Context

_No response_

### Code of Conduct

gauthier-th:
#### Description

The .update() method of TypeORM doesn't come with all the functionalities of the .save() method. Its goal is to partially update a Repository without doing any extra operation. This introduces issues with event from the database subscriber not triggered correctly, because of this partial update: https://github.com/typeorm/typeorm/issues/2809#issuecomment-451914877
A positive side effect of this change is that web push are now fixed.

This PR rollback to .save() instead of .update() to avoid the aforementioned issues, ~~at the cost of some lifehook issues. These issues are happening with PostgreSQL because the .save() method shouldn't be used inside an Entity Listener, an Event Subscriber should be used instead: https://orkhan.gitbook.io/typeorm/docs/listeners-and-subscribers#what-is-an-entity-listener~~

EDIT (By @fallenbagel):

This PR consolidates all database‐accessing lifecycle hooks into the MediaRequestSubscriber as this is the recommended way of doing database calls.
See: https://orkhan.gitbook.io/typeorm/docs/listeners-and-subscribers#what-is-an-entity-listener.

It also tightens the cascade options on the `MediaRequest→Media` relation to only `insert` and `remove` to avoid `hook‐within‐hook` recursions that could cause unexpected behavior.

These changes should guard against recursive hook invocations in any database backend and simplify the entity lifecycle.

Upstream PR also opened: https://github.com/sct/overseerr/pull/4124

#### To-Dos

### DIFF
--- a/server/entity/Media.ts
+++ b/server/entity/Media.ts
@@ -108,7 +108,9 @@ class Media {
   @Column({ type: 'int', default: MediaStatus.UNKNOWN })
   public status4k: MediaStatus;
 
-  @OneToMany(() => MediaRequest, (request) => request.media, { cascade: true })
+  @OneToMany(() => MediaRequest, (request) => request.media, {
+    cascade: ['insert', 'remove'],
+  })
   public requests: MediaRequest[];
 
   @OneToMany(() => Watchlist, (watchlist) => watchlist.media)
--- a/server/entity/MediaRequest.ts
+++ b/server/entity/MediaRequest.ts
@@ -1,10 +1,3 @@
-import type { RadarrMovieOptions } from '@server/api/servarr/radarr';
-import RadarrAPI from '@server/api/servarr/radarr';
-import type {
-  AddSeriesOptions,
-  SonarrSeries,
-} from '@server/api/servarr/sonarr';
-import SonarrAPI from '@server/api/servarr/sonarr';
 import TheMovieDb from '@server/api/themoviedb';
 import { ANIME_KEYWORD_ID } from '@server/api/themoviedb/constants';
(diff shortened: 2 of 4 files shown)


fallenbagel:
The last commit reverts this revert but its just to show what works for postgres to @gauthier-th (can be reverted if there is a way to fix this). It seems `.save()` breaks it completely for postgres where it goes on an infinite loop of hitting the `afterUpdate` unfortunately. 

From my testing, the .update changes I brought do not affect any subscribers that need an Entity as they're added for request declines which afaik is not needed by any subscribers. 

However, there is currently a weird ui issue that I have not delved deeply into. Even after an approval/rejection (though the status changes from `pending` to `requested` in the status badge, the request is still approvable/reject-able so you have to press twice for it to disappear which ofcourse tries to add the same media to services again).

fallenbagel:
Superseded by #1631

PR #1651, opened 2025-05-09
fix(jellyfin): clean up Jellyfin sessions on Jellyseerr logout

### LINKED ISSUE #1572: Jellyfin Session Token don't get revoked upon user login out off jellyseerr
### Description

Hi there,

i restricted the maximum simultaneous sessions per user(5 atm in my case) in jellyfin as a good rule of practice id say.
The problem:
If a user closes the jellyseerr tab/website or even logs completely out of jellyseerr, the jellyfin server doesn't notices this and as soon as the user logs back into jellyseerr a new session/token gets created.
The leads to the user not being able to log into jellyfin quite fast and i have to delete the sessions manually from within the jf dashboard.

### Version

2.5.2

### Steps to Reproduce

Create a new JF User and restrict the max simultaneous connections to 2 for example and login and out of jellyseerr 2-3 times.

Error with login fail can be seen in jellyfin container logs.

### Screenshots

Image

Image

### Logs

```shell

```

### Platform

desktop

### Database

SQLite (default)

### Device

Desktop x64

### Operating System

Linux fedora GNOME

### Browser

Firefox 

### Additional Context

_No response_

### Code of Conduct

0xSysR3ll:
#### Description

Currently, when users log out of Jellyseerr (or simply close their browser), their Jellyfin "Device" entries remain active on the Jellyfin server under the name "Jellyseerr."
This means old sessions pile up.
And this can lead to issues when users have a limited number of allowed simultaneous sessions in Jellyfin (e.g., 5 sessions per user), as each new Jellyseerr login creates a new session without cleaning up the old ones.

#### To-Dos

#### Issues Fixed or Closed

- Fixes #1572

### DIFF
--- a/server/api/externalapi.ts
+++ b/server/api/externalapi.ts
@@ -97,24 +97,6 @@ class ExternalAPI {
     return response.data;
   }
 
-  protected async delete<T>(
-    endpoint: string,
-    config?: AxiosRequestConfig
-  ): Promise<T> {
-    const cacheKey = this.serializeCacheKey(endpoint, {
-      ...config?.params,
-      headers: config?.headers,
-    });
-
-    const cachedItem = this.cache?.get<T>(cacheKey);
-    if (cachedItem) {
-      return cachedItem;
-    }
-
-    const response = await this.axios.delete<T>(endpoint, config);
-    return response.data;
-  }
-
   protected async getRolling<T>(
     endpoint: string,
     config?: AxiosRequestConfig,
--- a/server/api/externalapi.ts
+++ b/server/api/externalapi.ts
@@ -97,6 +97,24 @@ class ExternalAPI {
     return response.data;
   }
 
+  protected async delete<T>(
+    endpoint: string,
+    config?: AxiosRequestConfig
+  ): Promise<T> {
+    const cacheKey = this.serializeCacheKey(endpoint, {
+      ...config?.params,
+      headers: config?.headers,
+    });
+
+    const cachedItem = this.cache?.get<T>(cacheKey);
+    if (cachedItem) {
+      return cachedItem;
+    }
+
+    const response = await this.axios.delete<T>(endpoint, config);
+    return response.data;
+  }
+
   protected async getRolling<T>(
     endpoint: string,
     config?: AxiosRequestConfig,
--- a/server/api/jellyfin.ts
+++ b/server/api/jellyfin.ts
@@ -22,6 +22,23 @@ export interface JellyfinUserResponse {
   PrimaryImageTag?: string;
 }
 
+export interface JellyfinDevice {
+  Id: string;
+  Name: string;
+  LastUserName: string;
+  AppName: string;
+  AppVersion: string;
+  LastUserId: string;
+  DateLastActivity: string;
+  Capabilities: Record<string, unknown>;
+}
+
+export interface JellyfinDevicesResponse {
+  Items: JellyfinDevice[];
+  TotalRecordCount: number;
+  StartIndex: number;
+}
+
 export interface JellyfinLoginResponse {
   User: JellyfinUserResponse;
   AccessToken: string;
@@ -435,6 +452,51 @@ class JellyfinAPI extends ExternalAPI {
       );
 
       throw new ApiError(e.response?.status, ApiErrorCode.InvalidAuthToken);
+    }
+  }
+
+  public async deleteUserDevice(
+    userId: string,
+    deviceId: string
+  ): Promise<void> {
+    try {
+      const response = await this.get<JellyfinDevicesResponse>('/Devices', {
+        params: { UserId: userId },
+      });
+
+      logger.debug('Found Jellyfin devices', {
+        label: 'Jellyfin API',
+        deviceCount: response.Items.length,
+        userId,
+      });
+
+      const device = response.Items.find(
+        (item: JellyfinDevice) => item.Id === deviceId
+      );
+      if (!device) {
+        logger.debug('No matching Jellyfin device found', {
+          label: 'Jellyfin API',
+          deviceId,
+          userId,
+        });
+        return;
+      }
+
+      await this.delete('/Devices', { params: { Id: device.Id } });
+      logger.info('Successfully deleted Jellyfin device', {
+        label: 'Jellyfin API',
+        deviceId: device.Id,
+        deviceName: device.Name,
+        userId,
+      });
+    } catch (error) {
+      logger.error('Failed to delete Jellyfin device', {
+        label: 'Jellyfin API',
+        error: error instanceof Error ? error.message : 'Unknown error',
+        userId,
+        deviceId,
+      });
+      throw error;
     }
   }
 }
--- a/server/api/jellyfin.ts
+++ b/server/api/jellyfin.ts
@@ -454,51 +454,6 @@ class JellyfinAPI extends ExternalAPI {
       throw new ApiError(e.response?.status, ApiErrorCode.InvalidAuthToken);
     }
   }
-
-  public async deleteUserDevice(
-    userId: string,
-    deviceId: string
-  ): Promise<void> {
-    try {
-      const response = await this.get<JellyfinDevicesResponse>('/Devices', {
-        params: { UserId: userId },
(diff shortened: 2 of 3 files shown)


server/api/externalapi.ts
```diff
@@ -97,6 +97,24 @@ class ExternalAPI {
...
+
+    const cachedItem = this.cache?.get<T>(cacheKey);
+    if (cachedItem) {
+      return cachedItem;
+    }
+
+    const response = await this.axios.delete<T>(endpoint, config);
+    return response.data;
+  }
+
```
gauthier-th: This is not necessary. These `.get`, `.post`, etc. are useful for cache. Here you are retrieving a cached item that you didn't store with `cache.set`.
For `.delete`, there is no need for caching, so no need for this function. You can just do `this.axios.delete(...)` in the relevant place instead.

server/api/jellyfin.ts
```diff
@@ -437,6 +454,51 @@ class JellyfinAPI extends ExternalAPI {
...
+      if (!device) {
+        logger.debug('No matching Jellyfin device found', {
+          label: 'Jellyfin API',
+          deviceId,
+          userId,
+        });
+        return;
+      }
+
+      await this.delete('/Devices', { params: { Id: device.Id } });
```
gauthier-th:
```suggestion
      await this.axios.delete('/Devices', { params: { Id: device.Id } });
```

server/api/jellyfin.ts
```diff
@@ -437,6 +454,51 @@ class JellyfinAPI extends ExternalAPI {
...
+    try {
+      const response = await this.get<JellyfinDevicesResponse>('/Devices', {
+        params: { UserId: userId },
+      });
+
+      logger.debug('Found Jellyfin devices', {
+        label: 'Jellyfin API',
+        deviceCount: response.Items.length,
+        userId,
+      });
```
gauthier-th: Is this log really useful?

server/api/jellyfin.ts
```diff
@@ -437,6 +454,51 @@ class JellyfinAPI extends ExternalAPI {
       throw new ApiError(e.response?.status, ApiErrorCode.InvalidAuthToken);
     }
   }
+
+  public async deleteUserDevice(
+    userId: string,
+    deviceId: string
+  ): Promise<void> {
```
gauthier-th: Couldn't you just do `await this.axios.delete('/Devices', { params: { Id: deviceId } });` and check the result code of this delete instead of all this?

server/routes/auth.ts
```diff
@@ -714,17 +714,66 @@ authRoutes.post('/local', async (req, res, next) => {
...
+
+    if (user?.jellyfinUserId && user.jellyfinDeviceId) {
+      try {
+        const settings = getSettings();
+        if (!settings.jellyfin.apiKey) {
+          throw new Error('No Jellyfin API key configured');
+        }
+
+        const protocol = settings.jellyfin.useSsl ? 'https' : 'http';
+        const baseUrl = `${protocol}://${settings.jellyfin.ip}:${settings.jellyfin.port}${settings.jellyfin.urlBase}`;
```
gauthier-th: We have a function `getHostname()` in `server/utils/getHostname.ts` to do this.

server/routes/auth.ts
```diff
@@ -714,17 +714,66 @@ authRoutes.post('/local', async (req, res, next) => {
...
+      .addSelect(['user.jellyfinUserId', 'user.jellyfinDeviceId'])
+      .where('user.id = :id', { id: userId })
+      .getOne();
+
+    if (user?.jellyfinUserId && user.jellyfinDeviceId) {
+      try {
+        const settings = getSettings();
+        if (!settings.jellyfin.apiKey) {
+          throw new Error('No Jellyfin API key configured');
+        }
```
gauthier-th: I'm not sure it's in the scope of this controller to check if there is a Jellyfin server configured. We don't check it elsewhere.
@fallenbagel wdyt?

gauthier-th (changes requested):

server/routes/auth.ts
fallenbagel: Yes

server/api/jellyfin.ts
0xSysR3ll: I figured it could be a useful function for future implementations, but I can remove it yes.

server/api/jellyfin.ts
0xSysR3ll: Nope you're right, it was just for the debug :)

gauthier-th (approved):

fallenbagel (commented):
Should probably scope this to just `jellyfin/emby` server type

0xSysR3ll:
> Should probably scope this to just `jellyfin/emby` server type

Wdym ? I'm not sure to understand.

fallenbagel:
> > Should probably scope this to just `jellyfin/emby` server type
> 
> Wdym ? I'm not sure to understand.

This is currently querying anytime someone logs out right? Whether the mediaServerType is jellyfin or plex or local doesnt matter?

I want to scope the login to just if `mediaServerType` is equal to `MediaServerType.JELLYFIN` or `MediaServerType.EMBY`

0xSysR3ll:
> > > Should probably scope this to just `jellyfin/emby` server type
> > 
> > Wdym ? I'm not sure to understand.
> 
> This is currently querying anytime someone logs out right? Whether the mediaServerType is jellyfin or plex or local doesnt matter?
> 
> I want to scope the database query to delete the devices to just if `mediaServerType` is equal to `MediaServerType.JELLYFIN` or `MediaServerType.EMBY`

Well, the query to Jellyfin's API is conditioned by this : `if (user?.jellyfinUserId && user.jellyfinDeviceId) {` but yes I could add a check before that.

fallenbagel:
> > > > Should probably scope this to just `jellyfin/emby` server type
> > > 
> > > Wdym ? I'm not sure to understand.
> > 
> > This is currently querying anytime someone logs out right? Whether the mediaServerType is jellyfin or plex or local doesnt matter?
> > 
> > I want to scope the database query to delete the devices to just if `mediaServerType` is equal to `MediaServerType.JELLYFIN` or `MediaServerType.EMBY`
> 
> Well, the query to Jellyfin's API is conditioned by this : `if (user?.jellyfinUserId && user.jellyfinDeviceId)` but yes I could add a check before that.

I know. But then that makes it an unnecessary query to the db for mediaServerType.PLEX's every login.

You could also future-proof this with scoping to just Jellyfin/Emby userType instead of mediaServerType. Which gives the benefit of not doing a query for local users as well

fallenbagel:
> > > > Should probably scope this to just `jellyfin/emby` server type
> > > 
> > > Wdym ? I'm not sure to understand.
> > 
> > This is currently querying anytime someone logs out right? Whether the mediaServerType is jellyfin or plex or local doesnt matter?
> > 
> > I want to scope the database query to delete the devices to just if `mediaServerType` is equal to `MediaServerType.JELLYFIN` or `MediaServerType.EMBY`
> 
> Well, the query to Jellyfin's API is conditioned by this : `if (user?.jellyfinUserId && user.jellyfinDeviceId)` but yes I could add a check before that.

I know. But then that makes it an unnecessary query to the db for mediaServerType.PLEX's every login.

0xSysR3ll:
> > > > > Should probably scope this to just `jellyfin/emby` server type
> > > > 
> > > > Wdym ? I'm not sure to understand.
> > > 
> > > This is currently querying anytime someone logs out right? Whether the mediaServerType is jellyfin or plex or local doesnt matter?
> > > 
> > > I want to scope the database query to delete the devices to just if `mediaServerType` is equal to `MediaServerType.JELLYFIN` or `MediaServerType.EMBY`
> > 
> > Well, the query to Jellyfin's API is conditioned by this : `if (user?.jellyfinUserId && user.jellyfinDeviceId)` but yes I could add a check before that.
> 
> I know. But then that makes it an unnecessary query to the db for plex setups for every logout

Yep you're right. I will change that.

gauthier-th (approved):

fallenbagel (approved):

PR #1671, opened 2025-05-19
feat(episodes): add episode availability tracking and sync

### LINKED ISSUE #270: See which episode of a series is avaliable and which not
### Description

Currently, there is o way of seeing which episode of a "partially available" season is available and which is not.

### Desired Behavior

It would be nice to see which episode is still missing.

### Additional Context

Btw. you still got Overseerr in the Code of Conduct field
image

### Code of Conduct

0xSysR3ll:
## Description

This allows seerr to track the availability status of individual episodes, enabling better status reporting for partially available seasons.

- Fixes #270 

## How Has This Been Tested?

- Complete season except one episode available.
- Correctly show the "available" badge for each episode except for the missing one.

### Settings
image

### Result
image

### If TVDB not set
image

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Toggle to enable episode availability tracking in Settings with helper text and localization
  * Episode availability badges shown on TV season/episode views when enabled
  * Public settings now expose episode availability and metadata provider options

* **Chores**
  * Database and models updated to store and surface per-episode availability
  * Background sync updated to fetch, cache and persist per-episode availability
  * API responses include availability mapping for seasons

* **Migrations**
  * Added migration to create episode table for availability storage


fallenbagel:
This will not work properly when theres a mismatch between tvdb and tmdb right? How are you handling that

0xSysR3ll:
> This will not work properly when theres a mismatch between tvdb and tmdb right? Thats why our original plan was to wait till tvdb pr is merged until this feature. 
> 
> Unless you have some other ideas? (Haven't gone through the code yet)

You’re absolutely right about the TVDB/TMDB mismatch concern - I considered that during implementation. Instead of relying on direct TMDB ID mapping, I leveraged the existing Sonarr connection via externalServiceId and match episodes based on season and episode numbers. This approach works well for the majority of shows where episode numbering aligns across databases.

That said, I recognize there will be edge cases where TVDB and TMDB have inconsistent episode structures. I saw this as a pragmatic first step that covers most use cases.

I thought about a few options moving forward:
	•	Merge this as-is to support the majority of shows now.
	•	Hold off until the TVDB PR lands to enable more precise mapping.
	•	Make this behavior optional until we have more robust TVDB integration.

Wdyt ?

fallenbagel:
> Hold off until the TVDB PR lands to enable more precise mapping.

I think this would be the better idea. Or it will cause an influx of support requests saying why is this episode not available when it is etc etc

Gauvino:
Can a preview be made for this PR when the merge conflicts are resolved?

0xSysR3ll:
> Can a preview be made for this PR when the merge conflicts are resolved?

It still needs reviews and the merge of tvdb implementation's PR.

Gauvino:
The merge of the tvdb is done

0xSysR3ll:
> The merge of the tvdb is done

Now it is yes

gauthier-th:
Following up on the previous comments about episode mismatch with TMDB:
Do you think it would be a good idea to be able to enable this feature ONLY when TVDB is configured as metadata provider?
As said above, people will not understand why there are a lot of mismatch when they're using TMDB (the default metadata provider) and will constantly ask why there is a "bug".

0xSysR3ll:
> Do you think it would be a good idea to be able to enable this feature ONLY when TVDB is configured as metadata provider?

Yep, I think so. But we should be pretty clear about this - with a tooltip or something.

0xSysR3ll:
@gauthier-th Just added a setting to make it optional. And sync only works when TVDB is set as metadata provider for either tv or anime.

cypress/config/settings.cypress.json
```diff
@@ -201,4 +202,4 @@
       "forceMaxTtl": -1
     }
   }
-}
+}
```
gauthier-th: Missing newline

src/components/TvDetails/Season/index.tsx
```diff
@@ -52,6 +54,13 @@ const Season = ({ seasonNumber, tvId }: SeasonProps) => {
                     {episode.airDate && (
                       <AirDateBadge airDate={episode.airDate} />
                     )}
+                    <Badge badgeType={episode.available ? 'success' : 'danger'}>
+                      {intl.formatMessage(
+                        episode.available
+                          ? globalMessages.available
+                          : globalMessages.unavailable
+                      )}
+                    </Badge>
```
gauthier-th: This badge will show even if the metadata provider is not TVDB right?
Shouldn't we display it only for TVDB? It doesn't really make sense to display it otherwise.

server/lib/availabilitySync.ts
```diff
@@ -832,6 +840,84 @@ class AvailabilitySync {
...
+                    ? MediaStatus.UNKNOWN
+                    : ep.hasFile
+                    ? MediaStatus.AVAILABLE
+                    : MediaStatus.UNKNOWN;
+                  newEpisode.status4k = is4k
+                    ? ep.hasFile
+                      ? MediaStatus.AVAILABLE
+                      : MediaStatus.UNKNOWN
+                    : MediaStatus.UNKNOWN;
+                  newEpisode.season = Promise.resolve(season);
```
gauthier-th: Why this `Promise.resolve`?

server/routes/tv.ts
```diff
@@ -82,7 +82,30 @@ tvRoutes.get('/:id/season/:seasonNumber', async (req, res, next) => {
...
+        } else if (dbSeason.status === MediaStatus.PARTIALLY_AVAILABLE) {
+          if (dbSeason.episodes) {
+            for (const episode of dbSeason.episodes) {
+              availableMap[episode.episodeNumber] =
+                episode.status === MediaStatus.AVAILABLE;
+            }
+          }
+        }
+      }
+    }
```
gauthier-th: This is not necessary if the metadata provider is TMDB?

gauthier-th (changes requested):

gauthier-th:
Isn't it weird to display an "error" badge when the episode is not available but not released yet? Or not requested yet?

cypress/config/settings.cypress.json
0xSysR3ll: Shouldn't prettier handle this ?

src/components/TvDetails/Season/index.tsx
0xSysR3ll: Yep. You're right, I totally forgot about the frontend!

server/lib/availabilitySync.ts
0xSysR3ll: Well, I assumed we should use it as the relation is typed as `Promise<Season>`.

server/routes/tv.ts
0xSysR3ll: You mean we should add a condition to check if the provider is tvdb or not right ?

0xSysR3ll:
> Isn't it weird to display an "error" badge when the episode is not available but not released yet? Or not requested yet?

Would an airdate check be enough? IMO the "Unavailable" badge should still appear even if the episode hasn't been requested. This could be changed later as part of the "per-episode request feature" ?

server/routes/tv.ts
gauthier-th: Yes

gauthier-th:
> > Isn't it weird to display an "error" badge when the episode is not available but not released yet? Or not requested yet?
> 
> Would an airdate check be enough? IMO the '‘Unavailable" badge should still appear even if the episode hasn't been requested yet.

I mean, we don't write the status of a season when it's not requested. We only write "Requested" or "Available"/"Partially Available". Not a status in the error color, implying an error when it's "just" missing

0xSysR3ll:
> I mean, we don't write the status of a season when it's not requested. We only write "Requested" or "Available"/"Partially Available". Not a status in the error color, implying an error when it's "just" missing

Ah, that makes sense - my mistake. I thought it might be a small UX improvement, but you're right, it could be misleading.

src/components/TvDetails/Season/index.tsx
```diff
@@ -30,6 +39,15 @@ const Season = ({ seasonNumber, tvId }: SeasonProps) => {
     return <div>{intl.formatMessage(messages.somethingwentwrong)}</div>;
   }
 
+  const isAnime = tvData?.keywords.some(
+    (keyword) => keyword.id === ANIME_KEYWORD_ID
+  );
+  const isTvdbProvider = metadataSettings
```
0xSysR3ll: I'm really struggling to find the best approach here

cypress/config/settings.cypress.json
gauthier-th: Probably

src/components/TvDetails/Season/index.tsx
gauthier-th: You mean for the availability badge of the episode?

0xSysR3ll: I mean to check if tvdb is set as the provider. Not sure my implementation is the best.

gauthier-th: WDYT of that:
- Include the `metadataSettings` in the public settings, so you can then use the `useSettings()` hooks without having to fetch the metadata settings every time you open the Disclosure containing the episodes.
- Add a `isAnime` parameter to the `<Season />` component and use it in `src/components/TvDetails/index.tsx` to avoid one more API call to fetch the TV show detail.

0xSysR3ll: > WDYT of that:
> 
> * Include the `metadataSettings` in the public settings, so you can then use the `useSettings()` hooks without having to fetch the metadata settings every time you open the Disclosure containing the episodes.
> * Add a `isAnime` parameter to the `<Season />` component and use it in `src/components/TvDetails/index.tsx` to avoid one more API call to fetch the TV show detail.

TBH, I initially thought about doing this but assumed it might be out of scope for this PR. If you’re fine with it, I can go ahead and include it.

gauthier-th: Adding `metadataSettings` to the public settings shouldn't be a big change, and it will make it cleaner to this PR.

0xSysR3ll: Done then!

ctacquet:
What a great feature ! Hype to be able to request single épisode later ! 

_note: edited by a mod to keep in line with our community rules._

jmdv-es:
This will be a great feature, thank you all for your job! :)

bin/duplicate-detector/detect.mjs
```diff
@@ -188,7 +188,9 @@ function formatComment(candidates) {
     '',
     'A maintainer will review this. If this is **not** a duplicate, no action is needed.',
     '',
-    `<!-- duplicate-bot: candidates=${candidates.map((c) => c.number).join(',')} -->`
+    `<!-- duplicate-bot: candidates=${candidates
+      .map((c) => c.number)
+      .join(',')} -->`
```
fallenbagel: Why is this being changed in this PR 🗿

server/routes/tv.ts
```diff
@@ -80,10 +81,41 @@ tvRoutes.get('/:id/season/:seasonNumber', async (req, res, next) => {
     const season = await metadataProvider.getTvSeason({
       tvId: Number(req.params.id),
       seasonNumber: Number(req.params.seasonNumber),
-      language: (req.query.language as string) ?? req.locale,
```
fallenbagel, line 84: Why was this removed?

fallenbagel (changes requested):

bin/duplicate-detector/detect.mjs
0xSysR3ll: I'll take a look but I swear I didn't push this 🗿

gauthier-th (changes requested):
From what I'm seeing TypeORM migrations are 6+ months old. Please re-generate them.

0xSysR3ll:
> From what I'm seeing TypeORM migrations are 6+ months old. Please re-generate them.

YYR, done !

server/entity/Episode.ts
```diff
@@ -0,0 +1,43 @@
...
+  public status: MediaStatus;
+
+  @Column({ type: 'int', default: MediaStatus.UNKNOWN })
+  public status4k: MediaStatus;
+
+  @ManyToOne(() => Season, (season: Season) => season.episodes, {
+    onDelete: 'CASCADE',
+    nullable: true,
+  })
+  public season?: Promise<Season>;
```
gauthier-th, line 32: You should add an index here

server/entity/Season.ts
```diff
@@ -29,6 +31,11 @@ class Season {
   @Index()
   public media: Promise<Media>;
 
+  @OneToMany(() => Episode, (episode) => episode.season, {
+    cascade: true,
+  })
+  public episodes?: Episode[];
```
gauthier-th, line 38: Same here, you should add an index.

src/components/Settings/SettingsMain/index.tsx
```diff
@@ -535,6 +541,34 @@ const SettingsMain = () => {
...
+                      name="enableEpisodeAvailability"
+                      onChange={() => {
+                        setFieldValue(
+                          'enableEpisodeAvailability',
+                          !values.enableEpisodeAvailability
+                        );
+                      }}
+                    />
+                  </div>
+                </div>
```
gauthier-th, line 640: When this is enabled with TVDB not enabled, you could maybe add a warning saying that this setting is not used because of that? I know it's in the tooltip but having an additional warning when enabled (but not functional because no TVDB) can't hurt.

gauthier-th (changes requested):

server/entity/Season.ts
0xSysR3ll: If we already add an index on Episode's entity I'm not sure what we'd gain from an index on Season's side?

gauthier-th: @fallenbagel wdyt?

fallenbagel: I would say an index is needed here since the episode index covers the episode.seasonId FK, which should help when joining episodes to a season. The Season index covers season.mediaId, which helps when joining seasons to a media. Two different things on two different columnts. One doesn't help the other and indexes are fairly low cost in writing so should help a lot with performance when reading.

Though make sure it is added on the ManyToOne episode entity side where the `seasonId` FK column actually lives, not on the OneToMany here. OneToMany doesn't own a column so index here wouldn't really do anything

0xSysR3ll: Mmm maybe I'm mistaken here but hasn't the Season's index you're talking about already been implemented here : https://github.com/seerr-team/seerr/pull/2461 ?

gauthier-th: > If we already add an index on Episode's entity I'm not sure what we'd gain from an index on Season's side?

Yeah my bad, not needed. We already have the index on `season.mediaId` from #2461, and you added the index on `episode.season`, so it should be ok. Just one thing: could you move the `@Index` of `Episode` just before the `season` field? I find this more clear than having it at the beginning of the file.

0xSysR3ll: Nitpick but I understand :p

server/entity/Episode.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  @DbAwareColumn({ type: 'datetime', default: () => 'CURRENT_TIMESTAMP' })
+  public createdAt: Date;
+
+  @DbAwareColumn({
+    type: 'datetime',
+    default: () => 'CURRENT_TIMESTAMP',
+    onUpdate: 'CURRENT_TIMESTAMP',
```
gauthier-th: `onUpdate` should not be used. See #2823

gauthier-th (changes requested):

gauthier-th (dismissed):
LGTM. A rebase with a new generation of the migration will be needed before we merge this though.

gauthier-th:
@0xSysR3ll could you please rebase and fix the merge conflict? I'd like to get this in in the next release.

0xSysR3ll:
> @0xSysR3ll could you please rebase and fix the merge conflict? I'd like to get this in in the next release.

It should be good now.

matt-pinder:
Looking forward to this functionality 🙌

Thankyou

gauthier-th (changes requested):
Due to #2712 being merged with db migration, can you please re-generate yours? And then it should be good 😅

0xSysR3ll:
> Due to #2712 being merged with db migration, can you please re-generate yours? And then it should be good 😅

I was kinda expecting this, will do!

gauthier-th (dismissed):
LGTM. Can you rebase on develop so I can create a preview tag?

server/routes/tv.ts
```diff
@@ -84,7 +85,39 @@ tvRoutes.get('/:id/season/:seasonNumber', async (req, res, next) => {
       language: (req.query.language as string) ?? req.locale,
     });
 
-    return res.status(200).json(mapSeasonWithEpisodes(season));
+    const availableMap: Record<number, boolean> = {};
```
fallenbagel: `availableMap` is declared as `{}` unconditionally and only populated inside `if (shouldTrackEpisodes)` below. Since `mapEpisodeResult` checks `availableMap ? ... : undefined`, an empty object is still truthy, so wouldn't this mean when episode tracking is disabled (or TVDB isn't configured), every episode gets `available: false` instead of `available: undefined`?

I think we should declare it as a `let availableMap: Record<number, boolean> | undefined;` and only assigning `{}` inside the `shouldTrackEpisodes` block. Wdyt? @0xSysR3ll @gauthier-th 

Also available isn't documented in seerr-api.yml's Episode schema. Please add as an optional boolean.

server/routes/tv.ts
```diff
@@ -84,7 +85,39 @@ tvRoutes.get('/:id/season/:seasonNumber', async (req, res, next) => {
...
+        if (dbSeason) {
+          if (dbSeason.status === MediaStatus.AVAILABLE) {
+            for (const episode of season.episodes) {
+              availableMap[episode.episode_number] = true;
+            }
+          } else if (dbSeason.status === MediaStatus.PARTIALLY_AVAILABLE) {
+            if (dbSeason.episodes) {
+              for (const episode of dbSeason.episodes) {
+                availableMap[episode.episodeNumber] =
+                  episode.status === MediaStatus.AVAILABLE;
```
fallenbagel: `episodeNumber` here comes from Sonarr's `getEpisodes` (season-relative), while `episode_number` in the AVAILABLE branch above comes from TMDB.

In practice these should align for TVDB-backed anime since both are season-relative, but there's no explicit guard if a mismatch ever occurs meaning episodes could silently show wrong availability rather than erroring. 

This could cause issues in the the future when debugging if anime availability badges look off.

src/components/Settings/SettingsMain/index.tsx
```diff
@@ -70,6 +72,11 @@ const messages = defineMessages('components.Settings.SettingsMain', {
   validationApplicationUrlTrailingSlash: 'URL must not end in a trailing slash',
   partialRequestsEnabled: 'Allow Partial Series Requests',
   enableSpecialEpisodes: 'Allow Special Episodes Requests',
+  enableEpisodeAvailability: 'Enable Episode Availability Tracking',
+  enableEpisodeAvailabilityTip:
+    'Track individual episode availability status (requires TVDB as metadata provider for TV shows or anime)',
```
fallenbagel:
```suggestion
    'Track individual episode availability status (requires TVDB as metadata provider for either TV shows or anime)',
```

fallenbagel (changes requested):

gauthier-th:
Preview tag: `preview-episode-availability`

sdchew:
For TV series, I would like Seerr to track exactly which episodes are not fulfilled yet. Each episode in a requested season should have an Available or Pending tag.

gauthier-th:
@0xSysR3ll Can you please rebase on develop (again :sweat_smile:) so that we can try to merge it in the next release?

sdchew:
@0xSysR3ll I implemented my proposal that each episode in a requested season should have an Available or Pending tag. The implementation is in commit sdchew/seerr@76f84dd0. If it fits this PR, please feel free to cherry-pick it.

AI disclosure: I used Codex to prepare the rebase and implementation.

fallenbagel:
> @0xSysR3ll I implemented my proposal that each episode in a requested season should have an Available or Pending tag. The implementation is in commit sdchew/seerr@76f84dd0. If it fits this PR, please feel free to cherry-pick it.
> 
> AI disclosure: I used Codex to prepare the rebase and implementation.

The reason why this pr is only doing availability is because we are literally going to support episode requests. And we are doing it in small sequence of prs. 

In addition, our contributing policy doesn't allow fully ai-driven implementation. This PR will not include episode requests as it's currently well scoped for availability first.

server/entity/Media.ts
```diff
@@ -69,9 +69,27 @@ class Media {
...
+        requests: true,
+        issues: true,
+      };
+
+      // Only load seasons for TV shows
+      if (mediaType === MediaType.TV) {
+        relations.seasons = {
+          episodes: true,
+        };
+      }
```
fallenbagel: This isn't gated on `enableEpisodeAvailability`, so every TV fetch through `getMedia()` now eager-loads every episode of every season. 

The only caller that needs episode data is the season route, which already fetches media separately and gates it behind `shouldTrackEpisodes`. The main `/:id` route calls `getMedia()` unconditionally too, and `mapTvDetails` never reads episode data. So this adds a heavier join to every show page load even when the feature is off. Can this be conditioned on the setting, or moved out of this shared method?

server/lib/availabilitySync.ts
```diff
@@ -858,6 +886,83 @@ class AvailabilitySync {
...
+                (ep) => ep.seasonNumber === season.seasonNumber
+              );
+              if (seasonEpisodes.length === 0) {
+                continue;
+              }
+
+              const existingEpisodes = await episodeRepository.find({
+                where: { season: { id: season.id } },
+                relations: ['season'],
+              });
```
fallenbagel: This does a DB read (`episodeRepository.find`) per season, per media item, on every sync cycle. Given this runs against potentially thousands of shows, might be worth batching this.

Like fetch all episodes for the media item once (or once per sync pass) instead of re-querying per season.

server/lib/availabilitySync.ts
fallenbagel, line 1: This job only ever downgrades already **AVAILABLE**/**PARTIALLY_AVAILABLE** items when content's **gone** (`loadAvailableMediaPaginated`'s query, and every other status write in this file goes toward `DELETED`, never `AVAILABLE`). **The new episode code goes the other way**. It's the only place `Episode.status` gets set at all, and it sets `AVAILABLE` here. Scanners (`baseScanner.ts` + `jellyfin`/`plex`/`sonarr`/`radarr`) are what establish season/media availability everywhere else and aren't touched by this PR. 

Episode availability should be established there (the actual scanners), in sync with when the season itself is scanned, instead of depending on this job's sweep cadence. Also, using `UNKNOWN` instead of `DELETED` for "no file" is inconsistent with how this file treats everything else that's gone missing (soft removal).

In addition:

https://github.com/0xSysR3ll/jellyseerr/blob/5a3e96f2d5e616dc7656f003d0cb0f04dbee0f02/server/lib/availabilitySync.ts#L60-L63

This ORs tv/anime globally and with only `anime: TVDB` set, non-anime shows get Sonarr episode calls and Episode row writes here too. The mismatch check in https://github.com/seerr-team/seerr/pull/1671/commits/a91506da477ae216cb8923c041d5b53242914dd1 will suppress the badge for most of these anyway, so does the write side need the OR (ignoring the fact that episodes shouldn't be made available in file)?

fallenbagel (changes requested):

server/lib/availabilitySync.ts
0xSysR3ll: Though I had already done this. Good catch.

server/entity/Media.ts
0xSysR3ll: Makes sense, I'll move this from here.

server/lib/availabilitySync.ts
0xSysR3ll: Hum, would it be merge-blocking to not move episode availability into scanners ? Cause I can do it but this means a bigger PR. Unless I do it in a stacked PR. Wdyt ?

fallenbagel: > Hum, would it be merge-blocking to not move episode availability into scanners ? Cause I can do it but this means a bigger PR. Unless I do it in a stacked PR. Wdyt ?

Stacked PR is okay. We can merge from the top stack instead of this once thats done so both lands together. But remove this logic entirely from availability sync. It should never promote any media. Separation of concerns

0xSysR3ll: Sounds good to me then, will push a stacked PR.

0xSysR3ll:
Closed in favor of stacked PR : https://github.com/seerr-team/seerr/pull/3361

PR #1673, opened 2025-05-20
feat(requests): allow duplicate requests (#375)

### LINKED ISSUE #375: [Feature Request] Notify multiple users for same item
### Description

If an item is requested by user1, user2 can't request it and will get no notification when it's available. 
Adding a button like "notify me aswell" would be nice for user2. 

### Desired Behavior

user1 and user2 get a notification when the item is available.

### Additional Context

_No response_

### Code of Conduct

anhaehne:
#### Description

This change adds a setting and the neccesary changes to allow multiple users to request a tv show or a movie.
The RequestButton component has been adjusted to always show the request button unless the current user has a approved request when this setting is enabled.

#### To-Dos

#### Issues Fixed or Closed
- Fixes #375


anhaehne:
I've thought about how to implement a subscription based solution but i think in the end multiple request would better reflect the intent of multiple users being interested in seeing a particular media.

Let's look at an example with a subscription bases solution:

- User A: Requests a movie
- User B: Subscribes to the movie

Now one of multiple things can happen:

| Action | Outcome                                   |
|--------------------------------------------|----------------------------------------------------|
| The requests gets fulfilled             | User A and B get notified => OK                         |
| User B unsubscribes                     | User A is unaffected  => OK                               |
| User A’s request is deleted      | The movie will never be available and User B will never be notified  => Not OK  |

The last case can be avoided when User B created their own Request instead of subscribing.

I think the best solution would be to dedublicate the requests send to sonarr and radarr and allow multiple requests. 
To incorporate this into the other systems of Jellyseerr, duplicate requests should not detract from one's request quota and should be automatically approved.

Otherwise, implementing both approaches would also be possible.

fallenbagel:
> The last case can be avoided when User B created their own Request instead of subscribing

My plans were if User A deletes the request, the subscription changes into a request. To avoid multiple requests as that would require a substantial amount of refactor of the core logic to not break *arr. 

A subscription based model makes more sense imo

anhaehne:
I’m not sure how others are using Jellyseerr, but in my use case, deleting requests is only an administrative action. If I understand correctly, normal users can’t delete their own requests.

With a subscription-based system, the user who creates a request could automatically be subscribed to it. Users can manage their own subscriptions, but not the underlying request itself. This means the request is no longer user-specific, only the subscription is.

There are still some edge cases to consider, such as multiple users having pending requests for the same movie and only one being approved.

For TV shows, we should also decide whether “Request More” feature and managing subscriptions should be separate dialogs or unified into a single interface.

fallenbagel:
> I’m not sure how others are using Jellyseerr, but in my use case, deleting requests is only an administrative action. If I understand correctly, normal users can’t delete their own requests.
> 

No they can 100% cancel their own requests.

anhaehne:
Ok, maybe it is just the way I have configured my instance.

How about this: The user requesting a media can't cancel their request, they can only cancel their subscription. But if all subscriptions have been cancelled, the associated request also gets deleted.

fallenbagel:
> Ok, maybe it is just the way I have configured my instance.
> 

I'm not sure what you mean by this because you cannot disable cancelling requests. The only thing you can do is, give `auto-approval` permission to your users. So once a request is **APPROVED** they cannot cancel it. But when it's in pending state, they can 100% cancel it.

And no we don't let requests get deleted. Thats the whole reason why we even added soft delete behavior where when a media is removed, instead of getting the request deleted, it will just change status to DELETED. 

As for how to handle subscriptions, no we should not remove cancel request when its in pending state. When a request is in pending state, another user may subscribe to it, but if the OG user cancels it, then we can make the subscription of the other user another request that's waiting for approval.

But if the request is already **APPROVED** then the original user cant cancel it (that is the current behavior as well) and anyone can subscribe to that request. The subscription is only for notifications. When the item gets **available**. It will not be treated as a request. Its just for notifications.

anhaehne:
Ok, so far I was only thinking about approved requests.

I think I understand now how you would like this to be implemented.

I guess we can close this PR and I can create a proof of concept for the subscriptions approach.

Piddox:
> > The last case can be avoided when User B created their own Request instead of subscribing
> 
> My plans were if User A deletes the request, the subscription changes into a request. To avoid multiple requests as that would require a substantial amount of refactor of the core logic to not break *arr.
> 
> A subscription based model makes more sense imo

I have question about this approach, in the scenario where you have a separate radarr for HD and 4K, like me. 

Let say users A and B users have requested a movie, user A requested it in HD format and user B in 4K format. Now user C comes along and hits the subscribe button. User A cancels his request, so only the 4K is still requested. What happens to the subscription of user C? Does it become a request for the HD version? Or does it remain a subscription for the 4K?

What I think there should be is two subscription buttons in this case: one for the HD version (if it is already requested) and for the 4K version (if it is already requested). Then the user can decide to subscribe to both, and your solution to switching the subscription to a request in case the request gets deleted still works.

PR #1710, opened 2025-06-12
fix(jellyfin): use the same deviceId for admins

gauthier-th:
#### Description

This PR will make Jellyseerr use the same deviceId for the admin user everytime he logins to Jellyfin/Emby. The previous behavior with different deviceId was creating new entries on the media at every request.

#### To-Dos

### DIFF
--- a/server/routes/auth.ts
+++ b/server/routes/auth.ts
@@ -277,7 +277,7 @@ authRoutes.post('/jellyfin', async (req, res, next) => {
       select: { id: true, jellyfinDeviceId: true },
     });
 
-    let deviceId = '';
+    let deviceId = 'BOT_jellyseerr';
     if (user && user.id === 1) {
       // Admin is always BOT_jellyseerr
       deviceId = 'BOT_jellyseerr';
@@ -287,9 +287,6 @@ authRoutes.post('/jellyfin', async (req, res, next) => {
       deviceId = Buffer.from(`BOT_jellyseerr_${body.username}`).toString(
         'base64'
       );
-    } else {
-      // Fallback for when username is not provided
-      deviceId = 'BOT_jellyseerr';
     }
 
     // First we need to attempt to log the user in to jellyfin


server/routes/auth.ts
```diff
@@ -278,12 +278,18 @@ authRoutes.post('/jellyfin', async (req, res, next) => {
...
+      // Admin is always BOT_jellyseerr
+      deviceId = 'BOT_jellyseerr';
+    } else if (user && user.jellyfinDeviceId) {
+      deviceId = user.jellyfinDeviceId;
+    } else if (body.username) {
+      deviceId = Buffer.from(`BOT_jellyseerr_${body.username}`).toString(
         'base64'
       );
+    } else {
+      // Fallback for when username is not provided
```
fallenbagel: What do you mean by username is not provided. When would that case be?

gauthier-th: Yes, fixed

server/routes/auth.ts
```diff
@@ -277,11 +277,14 @@ authRoutes.post('/jellyfin', async (req, res, next) => {
...
+    let deviceId = 'BOT_jellyseerr';
+    if (user && user.id === 1) {
+      // Admin is always BOT_jellyseerr
+      deviceId = 'BOT_jellyseerr';
+    } else if (user && user.jellyfinDeviceId) {
+      deviceId = user.jellyfinDeviceId;
+    } else if (body.username) {
+      deviceId = Buffer.from(`BOT_jellyseerr_${body.username}`).toString(
         'base64'
       );
```
fallenbagel, line 289: Wouldn't this be cleaner and easier to read with:
```ts
let deviceId;

if (user?.id === 1) {
  // Admin is always BOT_jellyseerr
  deviceId = 'BOT_jellyseerr';
} else {
  deviceId = user?.jellyfinDeviceId 
    ?? (body?.username ? Buffer.from(`BOT_jellyseerr_${body.username}`).toString('base64') : 'BOT_jellyseerr');
}
```

gauthier-th: To be honest I prefer my version, I found it more simple :sweat_smile:

PR #1726, opened 2025-06-20
docs(docker): add healtheck and default to ghcr.io image

M0NsTeRRR:
#### Description

- add healtheck 
- default to ghcr.io image

### DIFF
--- a/server/api/externalapi.ts
+++ b/server/api/externalapi.ts
@@ -37,6 +37,8 @@ class ExternalAPI {
         ...options.headers,
       },
     });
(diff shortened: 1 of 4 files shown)


fallenbagel:
> IIRC @fallenbagel prefers to use Dockerhub by default, not ghcr.io

Not sure why but it looks cleaner 😆

M0NsTeRRR:
@fallenbagel I personnally prefer to use ghcr.io regarding [dockerhub rate limit](https://docs.docker.com/docker-hub/usage/). That's up to you guys ;)

PR #1736, opened 2025-06-22
fix: cover more content in the blacklisted tags processing job

TacoCake:
#### Description
1. I've modified the newly added “blacklisted tags” job, to include adult & video content, this correctly blacklists more adult films that were **always** missed, even if they had the tag.

Testing for the tag `softcore`
number of results blacklisted before: 3631
number of results blacklisted after: 3855

2. I've also removed video content from tmdb results by default, because it resulted in less incorrectly tagged adult content shown.

Before:
图片

After:
图片

Notice the one less adult video at the start.

#### To-Dos

#### Issues Fixed or Closed

- N/A

### DIFF
--- a/server/api/themoviedb/index.ts
+++ b/server/api/themoviedb/index.ts
@@ -31,6 +31,7 @@ interface SearchOptions {
   query: string;
   page?: number;
   includeAdult?: boolean;
+  includeVideo?: boolean;
   language?: string;
 }
 
@@ -72,6 +73,7 @@ export interface TmdbCertificationResponse {
 interface DiscoverMovieOptions {
   page?: number;
   includeAdult?: boolean;
+  includeVideo?: boolean;
   language?: string;
   primaryReleaseDateGte?: string;
   primaryReleaseDateLte?: string;
@@ -150,12 +152,17 @@ class TheMovieDb extends ExternalAPI {
     query,
     page = 1,
     includeAdult = false,
+    includeVideo = false,
     language = this.locale,
   }: SearchOptions): Promise<TmdbSearchMultiResponse> => {
     try {
       const data = await this.get<TmdbSearchMultiResponse>('/search/multi', {
         params: { query, page, include_adult: includeAdult, language },
       });
+
+      if (!includeVideo) {
+        data.results = data.results.filter((result: any) => !result.video)
+      }
 
       return data;
     } catch (e) {
@@ -172,6 +179,7 @@ class TheMovieDb extends ExternalAPI {
     query,
     page = 1,
     includeAdult = false,
+    includeVideo = false,
     language = this.locale,
     year,
   }: SingleSearchOptions): Promise<TmdbSearchMovieResponse> => {
@@ -185,6 +193,10 @@ class TheMovieDb extends ExternalAPI {
           primary_release_year: year,
         },
       });
+
+      if (!includeVideo) {
+        data.results = data.results.filter((result: any) => !result.video)
+      }
 
       return data;
     } catch (e) {
@@ -480,6 +492,7 @@ class TheMovieDb extends ExternalAPI {
     sortBy = 'popularity.desc',
     page = 1,
     includeAdult = false,
+    includeVideo = false,
     language = this.locale,
     primaryReleaseDateGte,
     primaryReleaseDateLte,
@@ -516,14 +529,15 @@ class TheMovieDb extends ExternalAPI {
           sort_by: sortBy,
           page,
           include_adult: includeAdult,
+          include_video: includeVideo,
           language,
           region: this.discoverRegion || '',
           with_original_language:
             originalLanguage && originalLanguage !== 'all'
               ? originalLanguage
               : originalLanguage === 'all'
-              ? undefined
-              : this.originalLanguage,
+                ? undefined
+                : this.originalLanguage,
           // Set our release date values, but check if one is set and not the other,
           // so we can force a past date or a future date. TMDB Requires both values if one is set!
           'primary_release_date.gte':
@@ -614,8 +628,8 @@ class TheMovieDb extends ExternalAPI {
             originalLanguage && originalLanguage !== 'all'
               ? originalLanguage
               : originalLanguage === 'all'
-              ? undefined
-              : this.originalLanguage,
+                ? undefined
+                : this.originalLanguage,
           include_null_first_air_dates: includeEmptyReleaseDate,
           with_genres: genre,
           with_networks: network,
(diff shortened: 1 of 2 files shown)


server/api/themoviedb/index.ts
```diff
@@ -150,13 +152,18 @@ class TheMovieDb extends ExternalAPI {
...
   }: SearchOptions): Promise<TmdbSearchMultiResponse> => {
     try {
       const data = await this.get<TmdbSearchMultiResponse>('/search/multi', {
         params: { query, page, include_adult: includeAdult, language },
       });
 
+      if (!includeVideo) {
+        data.results = data.results.filter((result: any) => !result.video)
+      }
+
```
gauthier-th: Why this addition? You could just add the `includeVideo` parameter into the TMDB Search query above?

server/api/themoviedb/index.ts
```diff
@@ -186,6 +194,10 @@ class TheMovieDb extends ExternalAPI {
         },
       });
 
+      if (!includeVideo) {
+        data.results = data.results.filter((result: any) => !result.video)
+      }
+
```
gauthier-th: Same

gauthier-th (changes requested):
Please fix the API request and the white-space changes (run `pnpm format` before committing)

server/api/themoviedb/index.ts
TacoCake: I would love to, but those endpoints ignore that parameter for some reason... I know it's dumb, but I followed the tmdb documentation.

So the only way I found to filter it, is to do it *after* getting the data.

https://developer.themoviedb.org/reference/search-multi

server/api/themoviedb/index.ts
TacoCake: Same reply as above, but with this link https://developer.themoviedb.org/reference/search-movie

fallenbagel:
> I agree with including the items marked as video in the blacklist tag processor, but not to have them removed by default of the search. You may want to search for something that is not a movie. @fallenbagel wdyt?
> 

Sounds logical to me

gauthier-th (changes requested):
Please make sure that the items are not removed by default of the search.

TacoCake:
Sorry for closing then re-opening the PR, had some issues fixing merge conflicts.

I reapplied my changes, but defaulted to included video results in the search.

Please re-review @gauthier-th

server/api/themoviedb/index.ts
```diff
@@ -153,13 +155,18 @@ class TheMovieDb extends ExternalAPI implements TvShowProvider {
...
   }: SearchOptions): Promise<TmdbSearchMultiResponse> => {
     try {
       const data = await this.get<TmdbSearchMultiResponse>('/search/multi', {
         params: { query, page, include_adult: includeAdult, language },
       });
 
+      if (!includeVideo) {
+        data.results = data.results.filter((result: any) => !result.video);
+      }
+
```
fallenbagel: Dead code? `includeVideo` defaults to `true` everywhere and the blacklist processor passes `true` explicitly. This filter never runs.

server/api/themoviedb/index.ts
```diff
@@ -189,6 +197,10 @@ class TheMovieDb extends ExternalAPI implements TvShowProvider {
         },
       });
 
+      if (!includeVideo) {
+        data.results = data.results.filter((result: any) => !result.video);
+      }
+
```
fallenbagel: Same as above

server/api/themoviedb/index.ts
TacoCake: Voluntarily left it there, if we ever want to set it to false this is necessary.

However, if you'd prefer, I can remove it.

server/api/themoviedb/index.ts
TacoCake: Same as above

server/api/themoviedb/index.ts
fallenbagel: Don't leave dead code anywhere. When a feature requires it, then that PR should add it. Keep it simple

TacoCake:
I removed the now-dead code. Please re-review @fallenbagel

server/job/blacklistedTagsProcessor.ts
```diff
@@ -121,6 +121,8 @@ class BlacklistedTagProcessor implements RunnableScanner<StatusBase> {
               page,
               sortBy,
               keywords: tag,
+              includeAdult: true,
+              includeVideo: true,
```
fallenbagel: > 'Video' content from tmdb would not be considered in the blacklisted tags job, but would be shown in the search results. This resulted in a small subset of undesired movies that couldn't be auto-blacklisted based on their tags

Adding `includeVideo` makes sense as you mentioned that `video` content wasn't being returned by blacklist job endpoints but was appearing in search results, so videos need to be explicitly included for proper blacklisting.

However, why is `includeAdult: true` needed here? Search endpoints already filter out adult content by default (`includeAdult = false`), so users never see it anyway. Why blacklist content that's already filtered on an api level and never reaches users?

TacoCake: What happens when someone comes along and changes that default `includeAdult` value to true? Then the blacklist processor wouldn't consider all the same results as the search.

Explicitly including everything, I feel, is the most permanent "fool-proof" solution. 

Regardless of what the user currently sees, it guarantees that they will _never_ see results with tags they black listed. 

At this point, I've been trying to get this merged for the past 4 months, beginning to regret ever opening a PR. Didn't think a 10 line change would take this much back and fourth.

fallenbagel: > What happens when someone comes along and changes that default `includeAdult` value to true? Then the blacklist processor wouldn't consider all the same results as the search.
> 
> Explicitly including everything, I feel, is the most permanent "fool-proof" solution. 
> 

I understand your concern, but we should only make changes when there’s a concrete need for them.

The default value for `includeAdult` isn’t something we plan to change, and if a future PR attempted to modify it, that would naturally go through the review process and be rejected.

That’s exactly why thorough reviews and discussions like this exist. To prevent unintended changes and maintain consistent behavior across the project.

> At this point, I've been trying to get this merged for the past 4 months, beginning to regret ever opening a PR. Didn't think a 10 line change would take this much back and fourth.

I understand the frustration with the review timeline.  Open-source contributions can definitely take longer than expected, especially when everyone’s contributing in their free time. This is not the only PR that has undergone thorough review processes or been open for a while. We are all volunteers. Working on this project for free in our own time. 

That said, part of maintaining a healthy project is ensuring we keep a consistent standard of quality through proper review and discussion, even for small changes. We want to make sure every modification is necessary and aligns with the overall design, rather than preemptively adding behavior that might not be needed yet.

Everyone here is volunteering their time, and we really appreciate your patience and contribution. The goal isn’t to block progress but to ensure the code remains maintainable and predictable in the long run. If you’d prefer not to continue with this PR, feel free to leave it. Someone else from the team can pick it up and carry it forward.

TacoCake: I appreciate the reply.

I ended up removing both `includeAdult: true` and `includeVideo: true`.

`IncludeAdult` because of what you said, and `includeVideo` because I already set the default value to true.

Please re-review @fallenbagel.

sct (approved):

PR #1748, opened 2025-06-27
docs(docker): add healthcheck to documentation

M0NsTeRRR:
#### Description
following https://github.com/fallenbagel/jellyseerr/pull/1726 (sad mistake :clown_face:)

I've removed ghcr.io as default until you are okay with it :)

### DIFF
--- a/docs/getting-started/docker.mdx
+++ b/docs/getting-started/docker.mdx
@@ -33,12 +33,19 @@ docker run -d \
   --name jellyseerr \
   -e LOG_LEVEL=debug \
   -e TZ=Asia/Tashkent \
-  -e PORT=5055 `#optional` \
+  -e PORT=5055 \
   -p 5055:5055 \
   -v /path/to/appdata/config:/app/config \
+  --health-cmd "wget --no-verbose --tries=1 --spider http://localhost:5055/api/v1/status || exit 1" \
+  --health-start-period 20s \
+  --health-timeout 3s \
+  --health-interval 15s \
+  --health-retries 3 \
   --restart unless-stopped \
   fallenbagel/jellyseerr
 ```
+
+The argument `-e PORT=5055` is optional.
 
 To run the container as a specific user/group, you may optionally add `--user=[ user | user:group | uid | uid:gid | user:gid | uid:group ]` to the above command.
 
@@ -46,7 +53,7 @@ To run the container as a specific user/group, you may optionally add `--user=[ 
 
 Stop and remove the existing container:
 ```bash
-docker stop jellyseerr && docker rm Jellyseerr
+docker stop jellyseerr && docker rm jellyseerr
 ```
 Pull the latest image:
 ```bash
@@ -83,6 +90,12 @@ services:
       - 5055:5055
     volumes:
       - /path/to/appdata/config:/app/config
+    healthcheck:
+      test: wget --no-verbose --tries=1 --spider http://localhost:5055/api/v1/status || exit 1
+      start_period: 20s
+      timeout: 3s
+      interval: 15s
+      retries: 3
     restart: unless-stopped
 ```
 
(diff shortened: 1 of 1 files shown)


docs/getting-started/docker.mdx
```diff
@@ -137,7 +148,20 @@ Then, create and start the Jellyseerr container:
 <Tabs groupId="docker-methods" queryString>
   <TabItem value="docker-cli" label="Docker CLI">
 ```bash
-docker run -d --name jellyseerr -e LOG_LEVEL=debug -e TZ=Asia/Tashkent -p 5055:5055 -v "jellyseerr-data:/app/config" --restart unless-stopped fallenbagel/jellyseerr:latest
+docker run -d \
+  --name jellyseerr \
+  -e LOG_LEVEL=debug \
+  -e TZ=Asia/Tashkent \
```
Gauvino, line 160: Why using this TZ by default and not `UTC` ?

gauthier-th: Because that's the timezone of our friend @fallenbagel 😉

Gauvino: But, it's a worldwide project lol ?

fallenbagel: > But, it's a worldwide project lol ?

Its an example..

docs/getting-started/docker.mdx
```diff
@@ -33,9 +33,14 @@ docker run -d \
   --name jellyseerr \
   -e LOG_LEVEL=debug \
   -e TZ=Asia/Tashkent \
-  -e PORT=5055 `#optional` \
+  -e PORT=5055 \
```
gauthier-th, line 36: Why not mark it as optional anymore?

docs/getting-started/docker.mdx
```diff
@@ -137,7 +148,20 @@ Then, create and start the Jellyseerr container:
 <Tabs groupId="docker-methods" queryString>
   <TabItem value="docker-cli" label="Docker CLI">
 ```bash
-docker run -d --name jellyseerr -e LOG_LEVEL=debug -e TZ=Asia/Tashkent -p 5055:5055 -v "jellyseerr-data:/app/config" --restart unless-stopped fallenbagel/jellyseerr:latest
+docker run -d \
+  --name jellyseerr \
+  -e LOG_LEVEL=debug \
+  -e TZ=Asia/Tashkent \
+  -e PORT=5055 \
```
gauthier-th, line 161: Same as above

gauthier-th (commented):
Do you think it's better to check with `/` or `/api/v1/status`?

docs/getting-started/docker.mdx
```diff
@@ -33,9 +33,14 @@ docker run -d \
...
   -e TZ=Asia/Tashkent \
-  -e PORT=5055 `#optional` \
+  -e PORT=5055 \
   -p 5055:5055 \
   -v /path/to/appdata/config:/app/config \
+  --health-cmd "wget --no-verbose --tries=1 --spider http://localhost:5055 || exit 1" \
+  --health-start-period 20s \
+  --health-timeout 3s \
+  --health-interval 15s \
+  --health-retries 3 \
```
fallenbagel: Shouldn't healthcheck be optional as well. It might confuse complete beginners when its too complicated

docs/getting-started/docker.mdx
M0NsTeRRR: If you copy and paste the command, you get the error `#optional: command not found`. Therefore, I removed it to make it more user-friendly for beginners when they copy and paste.

docs/getting-started/docker.mdx
M0NsTeRRR: I don't think so. Health checks are there to restart your container if it becomes unhealthy. In my opinion, for production usage, it's far better to have them enabled by default.

M0NsTeRRR:
> Do you think it's better to check with `/` or `/api/v1/status`?

As you prefer, we just need an unauthenticated route without redirection to speed up the health check.

docs/getting-started/docker.mdx
0xSysR3ll: I’ve almost never seen health check parameters used with docker run, since it’s intended as a quick and easy way to start a container.

IMO, for production, users should stick to using compose files instead, as they’re easier to manage and read.

M0NsTeRRR: In my view, the docker run command is intended for systemd users who prefer not to use Docker Compose or who choose to use systemd to start their containers. Otherwise, I would suggest moving it to another section, such as 'How to try out Jellyseerr'.

docs/getting-started/docker.mdx
gauthier-th: Makes sense. But could you mark it below if people want to change the internal port of jellyseerr?

docs/getting-started/docker.mdx
gauthier-th: I don't have the impression that including healthchecks in the docker run example is the norm (I almost never see them).
But maybe it can be added just below the docker run command saying something like "for production, you should incluse these healthchecks: "?

PR #1761, opened 2025-07-04
fix(gotify): notifications blocked when priority set to 0

### LINKED ISSUE #1754: Gotify no notification for requests (priority value should be 1 instead of 0)
### Description

Hello, 
If I click the button "test notification" it works, I do receive the message "check check,1,2,3 ...", however I don't receive the notification for the events checked.

It worked in the past it's been broken for a while now (can't say when, I'd say a few months ?)

Image

> [!NOTE]
> This issue happened since 2.6.0, gotify priority is set to "0" by default, setting it to 1 fixed the issue for me
### Version

2.7.0

### Steps to Reproduce

1. Go on notification setting and use gotify
2. Request a Movie/Tv show
3. No notification

0xSysR3ll:
#### Description
This PR fixes an issue where Gotify notifications would not be sent when the priority was configured to 0.

#### To-Dos

#### Issues Fixed or Closed

- Fixes #1754

### DIFF
--- a/server/subscriber/MediaRequestSubscriber.ts
+++ b/server/subscriber/MediaRequestSubscriber.ts
@@ -770,35 +770,35 @@ export class MediaRequestSubscriber
     await manager.save(fullMedia);
   }
 
-  public async afterUpdate(event: UpdateEvent<MediaRequest>): Promise<void> {
+  public afterUpdate(event: UpdateEvent<MediaRequest>): void {
     if (!event.entity) {
       return;
     }
 
-    await this.sendToRadarr(event.entity as MediaRequest);
-    await this.sendToSonarr(event.entity as MediaRequest);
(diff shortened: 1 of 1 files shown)


server/lib/notifications/agents/gotify.ts
```diff
@@ -49,7 +49,7 @@ class GotifyAgent
   ): GotifyPayload {
     const { applicationUrl, applicationTitle } = getSettings().main;
     const settings = this.getSettings();
-    const priority = settings.options.priority ?? 1;
+    const priority = settings.options.priority;
```
gauthier-th: Why remove the `?? 1`? IMO it's a pretty valid safeguard.

0xSysR3ll: You're right, that was an oversight on my part. I'll restore the fallback 😄

PR #1785, opened 2025-07-15
chore(helm): bump jellyseerr to 2.7.1

M0NsTeRRR:
#### Description

bump jellyseerr to 2.7.1

#### To-Dos

### DIFF
--- a/charts/jellyseerr-chart/Chart.yaml
+++ b/charts/jellyseerr-chart/Chart.yaml
@@ -3,8 +3,8 @@ kubeVersion: ">=1.23.0-0"
 name: jellyseerr-chart
 description: Jellyseerr helm chart for Kubernetes
 type: application
-version: 2.6.0
-appVersion: "2.7.0"
+version: 2.6.1
+appVersion: "2.7.1"
 maintainers:
   - name: Jellyseerr
     url: https://github.com/Fallenbagel/jellyseerr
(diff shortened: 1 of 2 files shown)


Gauvino:
Maybe in the chart action an auto-bump version Could be added ?

M0NsTeRRR:
> Maybe in the chart action an auto-bump version Could be added ?

Yes, the entire release process needs to be revisited to avoid using Node.js plugins.

M0NsTeRRR:
Using renovatebot + custom hook instead of dependabot would allow us to do that like I did some weeks ago in my helm chart repo.

https://github.com/M0NsTeRRR/helm-charts/blob/main/.github/workflows/renovate-custom-hooks.yml

Gauvino:
Yes it Would be less pain for your team to maintain

gauthier-th:
> the entire release process needs to be revisited to avoid using Node.js plugins.

I'm not sure why you're saying that?
I think Gauvino was just talking about the update of the version of the Helm chart?

M0NsTeRRR:
> > the entire release process needs to be revisited to avoid using Node.js plugins.
> 
> I'm not sure why you're saying that? I think Gauvino was just talking about the update of the version of the Helm chart?

Everything is connected. If I want to automate Helm chart releases, I need to integrate it into the current workflow. It will take less time for me to rework the current workflow now rather than update it later.

gauthier-th:
> > > the entire release process needs to be revisited to avoid using Node.js plugins.
> > 
> > I'm not sure why you're saying that? I think Gauvino was just talking about the update of the version of the Helm chart?
> 
> Everything is connected. If I want to automate Helm chart releases, I need to integrate it into the current workflow. It will take less time for me to rework the current workflow now rather than update it later.

Oh, ok. Well, as always, PR are more than welcome :)

PR #1802, opened 2025-07-23
feat(settings): add blocklist region and language options

### LINKED ISSUE #1801: blacklist tags feature and discovery language.
### Description

hello!
is it intended that the blacklist feature takes the "Discover Language" into account?
for example, the admin sets a language and does a "proccess blacklist tags" scan it only blacklists content for this language/region.
When a user switches the filter to global or any other language/region unwanted content will still be displayed.
either make a separate option to set the region/language for the blacklist scan or an option to lock users to region/language.

### Version

2.7.2

### Steps to Reproduce

-

### Screenshots

_No response_

### Logs

```shell

```

### Platform

desktop

### Database

SQLite (default)

### Device

-

### Operating System

-

### Browser

-

### Additional Context

_No response_

### Code of Conduct

0xSysR3ll:
#### Description
This PR introduces new settings for blocklisting content based on region and language, allowing users to filter out specific content independently from the general discover settings.

image

#### To-Dos

#### Issues Fixed or Closed

- Fixes #1801

## Summary by CodeRabbit

* **New Features**
  * Added Blocklist Region and Blocklist Language fields to Settings (including translations); values are persisted and the language selector is now reusable via a customizable field key.
* **Bug Fixes**
  * Adjusted dropdown stacking order in the region selector to prevent overlap.
* **Documentation**
  * Added Blocklist Region/Language documentation (duplicated entry present).


asentris:
Should "All Languages" be hidden and applied by default? I can't imagine anyone wants a blocked movie allowed for any reason.

0xSysR3ll:
> Should "All Languages" be hidden and applied by default? I can't imagine anyone wants a blocked movie allowed for any reason.

The "All Languages" filter is already applied by default when the blacklist language setting is left empty.

As for its visibility in the selector, I intentionally kept it to mirror the behavior of the other selectors (like Discover Language), which I believe improves UX through consistency.

That said, I do understand the concern — it could potentially be misinterpreted as a requirement to explicitly select it in order to apply to all languages.

I’m open to adjusting this if the maintainers prefer it hidden, but I think it’s worth discussing as part of a consistent UX pattern.

asentris:
Yea makes sense, I'll just ask a couple questions to try to get a better grasp of the project cause I'm interested in it 🙂 So when a request is sent for media, is it sent with a language so that it gets the correct information? If "All Languages" is selected, will it send the same tags for every language and potentially block movies that shouldn't be blocked? Does the blacklisted tag feature block movies with the request so they aren't even returned? If this is the case, would it be more accurate to filter out the blocked movies after they're received using their id's so that no matter the requesting language, the correct movies will be blocked?

0xSysR3ll:
Great questions – happy to clarify how it works 🙂

The language and region settings (including "All Languages") are only used when running the blacklist processing job. They control what content is scanned and checked for blacklisted tags –not how requests are filtered in real time.

Also, the blacklist logic itself is ID-based, not language-based. Once an item is blacklisted, it is excluded from discover/search regardless of language, region, or who’s requesting it.

So to your last question – yes, blacklisted content is filtered out after it is received, using its ID. Your suggestion is actually how it is implemented today.

This PR just adds more control for the admin to choose whether they want to blacklist, for example, only French-language content, or everything globally.

5urpriser:
please make it work :)

ScothularNetwork:
Hi, sorry if am being a bit dense, will this be merged in due course? - it would be a super helpful feature. Thank you.

Sc0th:
Hi, sorry if am being a bit dense, will this be merged in due course? - it would be a super helpful feature. Thank you.

gauthier-th (approved):

M0NsTeRRR (approved):

fallenbagel (approved):

PR #1811, opened 2025-07-27
Update buildfromsource.mdx

RootingInLoad:
The service file should use pnpm, not node. Using node will always trigger this error: [error]: SyntaxError: Unexpected token '{'

#### Description
Build from Source page from the Docs is misleading. The service file shouldn't utilize node but pnpm instead to start Jellyseerr.

#### To-Dos

### DIFF
--- a/docs/getting-started/buildfromsource.mdx
+++ b/docs/getting-started/buildfromsource.mdx
@@ -26,8 +26,7 @@ sudo mkdir -p /opt/jellyseerr && cd /opt/jellyseerr
 ```
 2. Clone the Jellyseerr repository and checkout the develop branch:
 ```bash
-git clone https://github.com/Fallenbagel/jellyseerr.git
-cd jellyseerr
+git clone https://github.com/Fallenbagel/jellyseerr.git .
 git checkout main
 ```
 3. Install the dependencies:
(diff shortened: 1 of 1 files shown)


fallenbagel:
No service file should be using node. Pnpm start is basically doing the same thing.

(This setup is used by many including me.)

fallenbagel:
If youre having issues with the service setup please hop onto discord for support.

0xSysR3ll:
> No service file should be using node. Pnpm start is basically doing the same thing.

But that's what the documentation actually says about service configuration. That is why they were suggesting modifying it. Maybe not the good approach ?

RootingInLoad:
> No service file should be using node. Pnpm start is basically doing the same thing.

I'm not sure whether I understand what you're saying, as the service file in the docs is clearly using node.

And using node doesn't work at all, only using pnpm start actually starts the server.

fallenbagel:
> > No service file should be using node. Pnpm start is basically doing the same thing.
> 
> I'm not sure whether I understand what you're saying, as the service file in the docs is clearly using node.
> 
> And using node doesn't work at all, only using pnpm start actually starts the server.

Like I said. Pnpm start basically uses node too. Its literally the same thing. https://github.com/fallenbagel/jellyseerr/blob/develop/package.json#L14

The reason why yours is not starting up is because of a misconfiguration so please jump into discord for support.

Literally everyone else is using the exact method in documentation. Since v1.0.0.

PR #1815, opened 2025-07-28
fix(blacklist): handle invalid keywords gracefully

### LINKED ISSUE #1814: Bug with blacklisted tags if tag becomes invalid
### Description

I recently started using tag-based blacklisting to try and clean up some of the queues that were full of stuff that some family members were complaining about.  The feature has worked great until all of the sudden the "Blacklist Content with Tags" field was blank and the job would no longer work, the following error was being recorded when landing on that settings page and when trying to add a new tag to the field:

[API]: Something went wrong retrieving keyword data {"errorMessage":"[TMDB] Failed to fetch keyword: Request failed with status code 404"}

It took a bit of trial and error to figure out the cause but it appears to be related to a previously valid keyword becoming invalid (deleted?).  The contents of settings.json for blacklistedTags was "321739,324336,336375,159551,155477,10053" but after running each tag through a curl API test I found that 336375 no longer existed (returned "The resource you requested could not be found.").  Upon removing that tag from the list now everything is back to normal again.

I don't know what circumstances lead to a previously existing tag being removed but would imagine this is something that could happen again and again over time.  Perhaps a change could be made to account for this scenario so that tag blacklisting can continue to work even if one becomes invalid (e.g. remove the invalid tag from the list then retry).

### Version

jellyseerr:preview-OIDC (39b6f47c104f9f0356bf51c6cb7e3996f154a8c2)

### Steps to Reproduce

1. Add a tag to "Blacklist Content with Tags" list
2. Tag becomes invalid (for test purposes I suppose you could force an invalid tag into settings.json)
4. Issue occurs

### Screenshots

_No response_

### Logs

```shell

```

### Platform

desktop

### Database

SQLite (default)

### Device

iPhone, Mac Mini, Windows 11

### Operating System

iOS 26, MacOS Sequoia 15.5, Windows 11

### Browser

Safari, Firefox

### Additional Context

_No response_

### Code of Conduct

0xSysR3ll:
#### Description

This PR fixes an issue where tag-based blacklisting would break if a previously valid TMDB keyword became invalid, causing the settings page to crash and jobs to fail with 404 errors.

API routes now return proper 404 statuses for invalid keywords. Keyword validity is checked before processing, and invalid keywords are automatically removed from settings.

#### To-Dos

#### Issues Fixed or Closed

- Fixes #1814

### DIFF
--- a/server/api/themoviedb/index.ts
+++ b/server/api/themoviedb/index.ts
@@ -1054,7 +1054,7 @@ class TheMovieDb extends ExternalAPI {
     keywordId,
   }: {
     keywordId: number;
-  }): Promise<TmdbKeyword> {
+  }): Promise<TmdbKeyword | null> {
     try {
       const data = await this.get<TmdbKeyword>(
         `/keyword/${keywordId}`,
@@ -1065,7 +1065,7 @@ class TheMovieDb extends ExternalAPI {
       return data;
     } catch (e) {
       if (e.response?.status === 404) {
-        throw e;
+        return null;
       }
       throw new Error(`[TMDB] Failed to fetch keyword: ${e.message}`);
     }
--- a/server/api/themoviedb/index.ts
+++ b/server/api/themoviedb/index.ts
@@ -1064,6 +1064,9 @@ class TheMovieDb extends ExternalAPI {
 
       return data;
     } catch (e) {
+      if (e.response?.status === 404) {
+        throw e;
+      }
       throw new Error(`[TMDB] Failed to fetch keyword: ${e.message}`);
     }
   }
--- a/src/components/BlacklistedTagsSelector/index.tsx
+++ b/src/components/BlacklistedTagsSelector/index.tsx
@@ -5,7 +5,10 @@ import { encodeURIExtraParams } from '@app/hooks/useDiscover';
 import defineMessages from '@app/utils/defineMessages';
 import { Transition } from '@headlessui/react';
 import { ArrowDownIcon } from '@heroicons/react/24/solid';
-import type { TmdbKeywordSearchResponse } from '@server/api/themoviedb/interfaces';
+import type {
+  TmdbKeyword,
+  TmdbKeywordSearchResponse,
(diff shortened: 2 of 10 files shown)


server/api/themoviedb/index.ts
```diff
@@ -1064,6 +1064,9 @@ class TheMovieDb extends ExternalAPI {
 
       return data;
     } catch (e) {
+      if (e.response?.status === 404) {
+        throw e;
+      }
```
gauthier-th: If the error is expected, you could instead return null? IMO it's a cleaner way to say that this is expected to have no data than throwing another type of error

src/components/BlacklistedTagsSelector/index.tsx
```diff
@@ -124,17 +124,23 @@ const ControlledKeywordSelector = ({
...
 
+      const validKeywords = keywords.filter((keyword) => keyword !== null);
+
       onChange(
-        keywords.map((keyword) => ({
-          label: keyword.name,
-          value: keyword.id,
+        validKeywords.map((keyword) => ({
+          label: keyword!.name,
+          value: keyword!.id,
```
gauthier-th: You should do `const validKeywords: TmdbKeyword[] = ` and then remove the non-null assertion

src/components/BlacklistedTagsSelector/index.tsx
```diff
@@ -124,17 +124,23 @@ const ControlledKeywordSelector = ({
...
-          );
-          return data;
+          try {
+            const { data } = await axios.get<Keyword>(
+              `/api/v1/keyword/${keywordId}`
+            );
+            return data;
+          } catch (error) {
+            return null;
+          }
```
gauthier-th: With null returned if the tag doesn't exist anymore, this wouldn't be necessary anymore

0xSysR3ll:
@gauthier-th  I've made the changes - does that work for you?

gauthier-th (approved):
lgtm!

PR #1901, opened 2025-09-11
chore(helm): switch from Deployment to StatefulSet

sbogomolov:
#### Description

Jellyseerr is not designed to support multiple cuncurrently running instances. To ensure that only one instance can run at any given time we:

- Replace `Deployment` with `StatefulSet`: switch from running "at least N pods" to "at most N pods".
- Remove `replicaCount` value: default is `1`, we do not want to support more.

#### Testing

Chart with the proposed changes is available here:
```
oci://ghcr.io/sbogomolov/jellyseerr/jellyseerr-chart
```

I have tested the new chart in my cluster, successfully upgrading from `2.6.2` to this version.

#### Issues Fixed or Closed

- Resolves https://github.com/fallenbagel/jellyseerr/discussions/1899.

### DIFF
--- a/charts/jellyseerr-chart/README.md
+++ b/charts/jellyseerr-chart/README.md
@@ -1,6 +1,6 @@
 # jellyseerr-chart
 
-![Version: 2.6.2](https://img.shields.io/badge/Version-2.6.2-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 2.7.3](https://img.shields.io/badge/AppVersion-2.7.3-informational?style=flat-square)
+![Version: 2.7.0](https://img.shields.io/badge/Version-2.7.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 2.7.3](https://img.shields.io/badge/AppVersion-2.7.3-informational?style=flat-square)
 
 Jellyseerr helm chart for Kubernetes
 
@@ -19,6 +19,17 @@ Jellyseerr helm chart for Kubernetes
 ## Requirements
 
 Kubernetes: `>=1.23.0-0`
+
+## Update Notes
+
+### Updating to 2.7.0
+
+Jellyseerr is a stateful application and it is not designed to have multiple replicas. In version 2.7.0 we address this by:
+
+- replacing `Deployment` with `StatefulSet`
+- removing `replicaCount` value
+
+If `replicaCount` value was used - remove it. Helm update should work fine after that.
 
 ## Values
 
@@ -55,7 +66,6 @@ Kubernetes: `>=1.23.0-0`
 | probes.livenessProbe | object | `{}` | Configure liveness probe |
 | probes.readinessProbe | object | `{}` | Configure readiness probe |
 | probes.startupProbe | string | `nil` | Configure startup probe |
-| replicaCount | int | `1` |  |
 | resources | object | `{}` |  |
 | securityContext | object | `{}` |  |
 | service.port | int | `80` |  |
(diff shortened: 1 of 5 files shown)


fallenbagel:
CC: @M0NsTeRRR

sudo-kraken:
@fallenbagel @M0NsTeRRR 

Only a couple of things:

- Helm can’t really swap a Deployment for a StatefulSet cleanly, so upgrades might hit a kind mismatch. Might be worth a quick upgrade note in the README (delete the old Deployment but keep the PVC, then re-apply).
- Could also mention in the docs that Jellyseerr isn’t meant to run more than one replica, and that SQLite on a PVC only works with a single pod.
- `serviceName` on a StatefulSet usually points to a headless service, fine with one replica, but worth clarifying.

Otherwise all good.

sbogomolov:
@sudo-kraken thanks for the feedback!

> Helm can’t really swap a Deployment for a StatefulSet cleanly, so upgrades might hit a kind mismatch. Might be worth a quick upgrade note in the README (delete the old Deployment but keep the PVC, then re-apply).

Upgrade did work for me. I don't think it tried swapping these. I think what happened is:
- Deployment manifest is gone, so helm removed the deployment
- StatefulSet manifest has appeared, so it created it 

These are two different resources that just happen to have the same name. If we were to keep the deployment manifest - we could have both. Or am I missing something?

> Could also mention in the docs that Jellyseerr isn’t meant to run more than one replica, and that SQLite on a PVC only works with a single pod.

Good point. I have mentioned this in the commit message. We do not have a way now to run multiple replicas, so I'm not sure in which context it would make sense to mention this in the README. If you think it does make sense - let's add it of course.

> `serviceName` on a StatefulSet usually points to a headless service, fine with one replica, but worth clarifying.

Not sure I understand this comment. Could you please clarify? In our particular case we do not really care about that headless service, it's just something we have to define for StatefulSet.

sudo-kraken:
@sbogomolov Apologies, I posted last night then went to sleep.

Yeah you’re right,  Helm will just delete the Deployment and create the StatefulSet since they’re different kinds, so it does work as an upgrade. The main risk is people being surprised by the kind mismatch if they try to do it with stricter Helm flags (e.g. --atomic). That’s why I thought a short upgrade note might help.

On the replicas/SQLite bit, I think it’s worth calling out in the README just so it’s obvious at a glance. Even if it’s mentioned in the commit message, users looking only at the chart docs might miss it.

Regarding the headless service, StatefulSets usually use one so each pod gets a stable DNS entry ([podname]].[service].[ns].svc). With one replica it doesn’t matter much, but I just meant it might be clearer to note why serviceName is set. Nothing to change really, just a doc/clarity thing.

One thing I was wondering though, what happens with PVCs if a volumeClaimTemplate is used? since that would create a new claim, the old Deployment’s PVC wouldn’t be adopted. Did you run into that, or are you relying on existingClaim instead?

gauthier-th:
> On the replicas/SQLite bit

FYI Jellyseerr also supports Postgres

sbogomolov:
@sudo-kraken I've added a new section to the README. What do you think?

> One thing I was wondering though, what happens with PVCs if a volumeClaimTemplate is used? since that would create a new claim, the old Deployment’s PVC wouldn’t be adopted. Did you run into that, or are you relying on existingClaim instead?

For me upgrade went smoothly. I guess PVC stayed bound to PV and was reused by StatefulSet.

I have it configured like this (I use PostgreSQL, so my config volume can be on NFS, hence `ReadWriteMany`):
```yaml
    config:
      persistence:
        size: 10Gi
        accessModes:
        - ReadWriteMany
        volumeName: jellyseerr-config-pv
```

sudo-kraken:
Looks good to me now, good work👍

charts/jellyseerr-chart/README.md
```diff
@@ -64,7 +63,17 @@ Kubernetes: `>=1.23.0-0`
...
+## Update Notes
+
+### Updating to 2.7.0
+
+Jellyseerr is a stateful application and it is not designed to have multiple replicas. In version 2.7.0 we address this by:
+
+- replacing `Deployment` with `StatefulSet`
+- removing `replicaCount` value
+
+If `replicaCount` value was used - remove it. Helm update should work fine after that.
```
M0NsTeRRR: This note must be added to the `README.md.gotmpl` file, otherwise it will be lost during the next `helm-docs` run. Let's add it before `{{ template "chart.valuesSection" . }}`.

sbogomolov: Thanks for noticing this, fixed.

M0NsTeRRR: Don't forget to run helm-docs the README is missing the update notes section from the template

sbogomolov: Done.

M0NsTeRRR:
LGTM @fallenbagel @gauthier-th if the CI pass you can merge, don't forget to update the PR title to `chore(helm)`.

Thanks for your contribution @sbogomolov !

gauthier-th:
Thank you both for you reviews!

gauthier-th (approved):
LGTM, thanks!

fallenbagel (approved):

PR #1904, opened 2025-09-14
feat: native ARM builds + updated workflow architecture

sudo-kraken:
- Replace QEMU emulation with native ARM runners for performance reasons,
- Standardise all workflows with cache-warming, deterministic timestamps,
- Added more container metadata while maintaining clean registries

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,14 @@ on:
   push:
     branches:
       - develop
+  workflow_dispatch:
+permissions:
+  contents: read
+  packages: write
+
+concurrency:
+  group: ci-${{ github.ref }}
+  cancel-in-progress: true
 
 jobs:
   test:
@@ -44,111 +52,109 @@ jobs:
         run: pnpm build
 
   build:
-    name: Build & Publish Docker Images
+    name: Build (per-arch, native runners)
     if: github.ref == 'refs/heads/develop' && !contains(github.event.head_commit.message, '[skip ci]')
     strategy:
       matrix:
         include:
           - runner: ubuntu-24.04
             platform: linux/amd64
+            arch: amd64
           - runner: ubuntu-24.04-arm
             platform: linux/arm64
+            arch: arm64
     runs-on: ${{ matrix.runner }}
-    outputs:
-      digest-amd64: ${{ steps.set_outputs.outputs.digest-amd64 }}
-      digest-arm64: ${{ steps.set_outputs.outputs.digest-arm64 }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+
+      - name: Deterministic timestamp
+        id: ts
+        run: echo "now=$(date -u +%Y-%m-%dT%H:%M:%SZ)" >> "$GITHUB_OUTPUT"
+
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
-      - name: Log in to Docker Hub
-        uses: docker/login-action@v3
-        with:
-          username: ${{ secrets.DOCKER_USERNAME }}
-          password: ${{ secrets.DOCKER_TOKEN }}
-      - name: Log in to GitHub Container Registry
-        uses: docker/login-action@v3
-        with:
-          registry: ghcr.io
-          username: ${{ github.repository_owner }}
-          password: ${{ secrets.GITHUB_TOKEN }}
-      - name: Set lower case owner name
-        run: |
-          echo "OWNER_LC=${OWNER,,}" >>${GITHUB_ENV}
-        env:
-          OWNER: ${{ github.repository_owner }}
-      - name: Docker metadata
-        id: meta
-        uses: docker/metadata-action@v4
-        with:
-          images: |
-            fallenbagel/jellyseerr
-            ghcr.io/${{ env.OWNER_LC }}/jellyseerr
-          tags: |
-            type=ref,event=branch
-            type=sha,prefix=,suffix=,format=short
-      - name: Build and push by digest
-        id: build
-        uses: docker/build-push-action@v5
+
+      - name: Warm cache (no push) — ${{ matrix.platform }}
+        uses: docker/build-push-action@v6
         with:
           context: .
           file: ./Dockerfile
           platforms: ${{ matrix.platform }}
-          push: true
+          push: false
           build-args: |
             COMMIT_TAG=${{ github.sha }}
             BUILD_VERSION=develop
-            BUILD_DATE=${{ github.event.repository.updated_at }}
-          outputs: |
-            type=image,push-by-digest=true,name=fallenbagel/jellyseerr,push=true
-            type=image,push-by-digest=true,name=ghcr.io/${{ env.OWNER_LC }}/jellyseerr,push=true
+            BUILD_DATE=${{ steps.ts.outputs.now }}
           cache-from: type=gha,scope=${{ matrix.platform }}
           cache-to: type=gha,mode=max,scope=${{ matrix.platform }}
           provenance: false
-      - name: Set outputs
-        id: set_outputs
-        run: |
-          platform="${{ matrix.platform == 'linux/amd64' && 'amd64' || 'arm64' }}"
-          echo "digest-${platform}=${{ steps.build.outputs.digest }}" >> $GITHUB_OUTPUT
 
-  merge_and_push:
-    name: Create and Push Multi-arch Manifest
+  publish:
+    name: Publish multi-arch image
     needs: build
     runs-on: ubuntu-24.04
     steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Deterministic timestamp
+        id: ts
+        run: echo "now=$(date -u +%Y-%m-%dT%H:%M:%SZ)" >> "$GITHUB_OUTPUT"
+
+      - name: Set up QEMU (enable ARM emulation)
+        uses: docker/setup-qemu-action@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
       - name: Log in to Docker Hub
         uses: docker/login-action@v3
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_TOKEN }}
+
       - name: Log in to GitHub Container Registry
         uses: docker/login-action@v3
         with:
           registry: ghcr.io
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
-      - name: Set lower case owner name
-        run: |
-          echo "OWNER_LC=${OWNER,,}" >>${GITHUB_ENV}
+
+      - name: Lower-case owner
+        run: echo "OWNER_LC=${OWNER,,}" >> "$GITHUB_ENV"
         env:
           OWNER: ${{ github.repository_owner }}
-      - name: Create and push manifest
-        run: |
-          docker manifest create fallenbagel/jellyseerr:develop \
-            --amend fallenbagel/jellyseerr@${{ needs.build.outputs.digest-amd64 }} \
-            --amend fallenbagel/jellyseerr@${{ needs.build.outputs.digest-arm64 }}
-          docker manifest push fallenbagel/jellyseerr:develop
 
-          # GHCR manifest
-          docker manifest create ghcr.io/${{ env.OWNER_LC }}/jellyseerr:develop \
-            --amend ghcr.io/${{ env.OWNER_LC }}/jellyseerr@${{ needs.build.outputs.digest-amd64 }} \
-            --amend ghcr.io/${{ env.OWNER_LC }}/jellyseerr@${{ needs.build.outputs.digest-arm64 }}
-          docker manifest push ghcr.io/${{ env.OWNER_LC }}/jellyseerr:develop
+      - name: Build & Push (multi-arch, single tag)
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: ./Dockerfile
+          platforms: linux/amd64,linux/arm64
+          push: true
+          build-args: |
+            COMMIT_TAG=${{ github.sha }}
+            BUILD_VERSION=develop
+            BUILD_DATE=${{ steps.ts.outputs.now }}
+          labels: |
+            org.opencontainers.image.source=${{ github.server_url }}/${{ github.repository }}
+            org.opencontainers.image.revision=${{ github.sha }}
+            org.opencontainers.image.version=develop
+          tags: |
+            fallenbagel/jellyseerr:develop
+            fallenbagel/jellyseerr:${{ github.sha }}
+            ghcr.io/${{ env.OWNER_LC }}/jellyseerr:develop
+            ghcr.io/${{ env.OWNER_LC }}/jellyseerr:${{ github.sha }}
+          cache-from: |
+            type=gha,scope=linux/amd64
+            type=gha,scope=linux/arm64
+          cache-to: type=gha,mode=max
+          provenance: false
 
   discord:
     name: Send Discord Notification
-    needs: merge_and_push
+    needs: publish
     if: always() && github.event_name != 'pull_request' && !contains(github.event.head_commit.message, '[skip ci]')
     runs-on: ubuntu-24.04
     steps:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,6 +1,15 @@
-name: Jellyseer Release
+name: Jellyseerr Release
 
-on: workflow_dispatch
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  packages: write
+
+concurrency:
+  group: release-${{ github.ref }}
+  cancel-in-progress: true
 
 jobs:
   semantic-release:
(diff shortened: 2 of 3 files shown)


gauthier-th:
@M0NsTeRRR review welcome 😄

sudo-kraken:
With the new workflows we build on amd64 and arm64 first to warm the cache, then do a single multi-arch publish. The final build mostly hits cached layers, only the publish job holds registry secrets so pushes happen in a trusted context.

This way we prove both architectures actually compile before touching registries. This makes for one clean multi-arch tag and lets us re-run publish without redoing the warm-ups. Which leads to subsequent runs being noticeably faster because the per-arch builds write a Buildx cache to GitHub Actions that the multi-arch publish reads. It will stay fast so long as layers don’t change. 

The first test run of the preview was 9m 47s and the second was 5m 44s, this proves the theory in the above.

Overall, I think this strikes the best balance between best practice and speed.

.github/workflows/ci.yml
```diff
@@ -44,111 +52,109 @@ jobs:
...
-    outputs:
-      digest-amd64: ${{ steps.set_outputs.outputs.digest-amd64 }}
-      digest-arm64: ${{ steps.set_outputs.outputs.digest-arm64 }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+
+      - name: Deterministic timestamp
+        id: ts
+        run: echo "now=$(date -u +%Y-%m-%dT%H:%M:%SZ)" >> "$GITHUB_OUTPUT"
```
M0NsTeRRR: Use `echo "TIMESTAMP=$(git log -1 --pretty=%ct)" >> "$GITHUB_OUTPUT"` to get docker reproductible build https://docs.docker.com/build/ci/github-actions/reproducible-builds/#git-commit-timestamps

And we can add `ARG SOURCE_DATE_EPOCH` in the `Dockerfile`

.github/workflows/ci.yml
```diff
@@ -44,111 +52,109 @@ jobs:
...
-          push: true
+          push: false
           build-args: |
             COMMIT_TAG=${{ github.sha }}
             BUILD_VERSION=develop
-            BUILD_DATE=${{ github.event.repository.updated_at }}
-          outputs: |
-            type=image,push-by-digest=true,name=fallenbagel/jellyseerr,push=true
-            type=image,push-by-digest=true,name=ghcr.io/${{ env.OWNER_LC }}/jellyseerr,push=true
+            BUILD_DATE=${{ steps.ts.outputs.now }}
```
M0NsTeRRR: Use `SOURCE_DATE_EPOCH=${{ steps.ts.outputs.TIMESTAMP }}`
and configure output with `rewrite-timestamp=true` [rewrite-timestamp doc](https://github.com/moby/buildkit/blob/master/docs/build-repro.md#source_date_epoch)

.github/workflows/release.yml
```diff
@@ -1,120 +1,182 @@
...
           key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
           restore-keys: |
             ${{ runner.os }}-pnpm-store-
+
       - name: Install dependencies
         run: pnpm install
+
       - name: Release
+        id: release
+        uses: cycjimmy/semantic-release-action@v4
```
M0NsTeRRR: We should remove the dependency inside the `package.json` and use [extra_plugins](https://github.com/cycjimmy/semantic-release-action?tab=readme-ov-file#extra_plugins) ?

.github/workflows/ci.yml
```diff
@@ -44,111 +52,109 @@ jobs:
...
+          labels: |
+            org.opencontainers.image.source=${{ github.server_url }}/${{ github.repository }}
+            org.opencontainers.image.revision=${{ github.sha }}
+            org.opencontainers.image.version=develop
+          tags: |
+            fallenbagel/jellyseerr:develop
+            fallenbagel/jellyseerr:${{ github.sha }}
+            ghcr.io/${{ env.OWNER_LC }}/jellyseerr:develop
+            ghcr.io/${{ env.OWNER_LC }}/jellyseerr:${{ github.sha }}
```
M0NsTeRRR: We can use [docker/metadata-action](https://github.com/docker/metadata-action)

M0NsTeRRR:
I didn’t re-add some comments I previously made in #1905 that apply here.

.github/workflows/ci.yml
sudo-kraken: Ah, I went with `date -u` just to have a clean UTC timestamp, but tying it to git log makes more sense to tie the build date to the commit.

.github/workflows/ci.yml
sudo-kraken: I hadn’t considered that downstream tools can pick up `SOURCE_DATE_EPOCH` automatically, thats cleaner.

.github/workflows/ci.yml
sudo-kraken: My approach ismore explicit, so its obvious what metadata is being set, but I agree your way is leaner if we are happy with the defaults that are applied.

.github/workflows/release.yml
sudo-kraken: Assuming no one here actually runs releases locally. Am I right in thinking thats the case?

M0NsTeRRR: And, from a security perspective, especially if we ever implement attestations 😄 we probably shouldn’t. Thoughts, @fallenbagel @gauthier-th ?

M0NsTeRRR:
Also you can looks into [zizmor tool](https://github.com/zizmorcore/zizmor), I haven't run it yet but it's a security static analysis tool to find vulnetability in workflows

.github/workflows/release.yml
gauthier-th: No no we don't use it locally

sudo-kraken:
> Also you can looks into [zizmor tool](https://github.com/zizmorcore/zizmor), I haven't run it yet but it's a security static analysis tool to find vulnetability in workflows

looked into it and it works well but does throw a lot of errors about any action not pinned to a hash.

sudo-kraken:
> Also you can looks into [zizmor tool](https://github.com/zizmorcore/zizmor), I haven't run it yet but it's a security static analysis tool to find vulnetability in workflows

```sh
$ zizmor ./.github/workflows/ci.yml 
 INFO zizmor::registry: skipping impostor-commit: can't run without a GitHub API token          
 INFO zizmor::registry: skipping ref-confusion: can't run without a GitHub API token
 INFO zizmor::registry: skipping known-vulnerable-actions: can't run without a GitHub API token 
 INFO zizmor::registry: skipping stale-action-refs: can't run without a GitHub API token
 INFO audit: zizmor: 🌈 completed ./.github/workflows/ci.yml
error[unpinned-uses]: unpinned action reference
  --> ./.github/workflows/ci.yml:30:9
   |
30 |         uses: pnpm/action-setup@v4
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^ action is not pinned to a hash (required by blanket policy)
   |
   = note: audit confidence → High

error[unpinned-uses]: unpinned action reference
  --> ./.github/workflows/ci.yml:79:9
   |
79 |         uses: docker/setup-buildx-action@v3
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ action is not pinned to a hash (required by blanket policy)
   |
   = note: audit confidence → High

error[unpinned-uses]: unpinned action reference
  --> ./.github/workflows/ci.yml:82:9
   |
82 |         uses: docker/build-push-action@v6
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ action is not pinned to a hash (required by blanket policy)
   |
   = note: audit confidence → High

error[unpinned-uses]: unpinned action reference
   --> ./.github/workflows/ci.yml:115:9
    |
115 |         uses: docker/setup-qemu-action@v3
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ action is not pinned to a hash (required by blanket policy)
    |
    = note: audit confidence → High

error[unpinned-uses]: unpinned action reference
   --> ./.github/workflows/ci.yml:118:9
    |
118 |         uses: docker/setup-buildx-action@v3
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ action is not pinned to a hash (required by blanket policy)
    |
    = note: audit confidence → High

error[unpinned-uses]: unpinned action reference
   --> ./.github/workflows/ci.yml:121:9
    |
121 |         uses: docker/login-action@v3
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ action is not pinned to a hash (required by blanket policy)
    |
    = note: audit confidence → High

error[unpinned-uses]: unpinned action reference
   --> ./.github/workflows/ci.yml:127:9
    |
127 |         uses: docker/login-action@v3
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ action is not pinned to a hash (required by blanket policy)
    |
    = note: audit confidence → High

error[unpinned-uses]: unpinned action reference
   --> ./.github/workflows/ci.yml:139:9
    |
139 |         uses: docker/build-push-action@v6
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ action is not pinned to a hash (required by blanket policy)
    |
    = note: audit confidence → High

error[unpinned-uses]: unpinned action reference
   --> ./.github/workflows/ci.yml:171:9
    |
171 |         uses: technote-space/workflow-conclusion-action@v3
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ action is not pinned to a hash (required by blanket policy)
    |
    = note: audit confidence → High

error[unpinned-uses]: unpinned action reference
   --> ./.github/workflows/ci.yml:182:9
    |
182 |         uses: sarisia/actions-status-discord@v1
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ action is not pinned to a hash (required by blanket policy)
    |
    = note: audit confidence → High

11 findings (1 suppressed): 0 unknown, 0 informational, 0 low, 0 medium, 10 high
```

.github/workflows/ci.yml
```diff
@@ -7,6 +7,14 @@ on:
   push:
     branches:
       - develop
+  workflow_dispatch:
+permissions:
+  contents: read
+  packages: write
```
M0NsTeRRR: Here as discussed on the other PR

M0NsTeRRR:
Yup looks good, you forgot to push ?

sudo-kraken:
> Yup looks good, you forgot to push ?

I have pushed the others on PR-1905, still working on the release based ci flows.

sudo-kraken:
@M0NsTeRRR I have pushed the updates to clean up package.json, removed the local @semantic-release/git and @semantic-release/changelog since we already inject them via workflow extra_plugins. 

Adjusted Dockerfile to streamline build, removed the hard-coded OCI labels from the Dockerfile and shifted them into the metadata-action step.

Updated metadata steps to add standard OCI labels and ensure consistency across release/preview/CI builds.

I've been doing this between jobs today so please can you re-review and ensure I didnt miss anything obvious.

.github/workflows/ci.yml
```diff
@@ -44,111 +53,121 @@ jobs:
...
-          docker manifest create ghcr.io/${{ env.OWNER_LC }}/jellyseerr:develop \
-            --amend ghcr.io/${{ env.OWNER_LC }}/jellyseerr@${{ needs.build.outputs.digest-amd64 }} \
-            --amend ghcr.io/${{ env.OWNER_LC }}/jellyseerr@${{ needs.build.outputs.digest-arm64 }}
-          docker manifest push ghcr.io/${{ env.OWNER_LC }}/jellyseerr:develop
+      - name: Extract metadata
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: |
+            fallenbagel/jellyseerr
```
M0NsTeRRR: Would be better to use github context variable instead of owner and repository hardcoded value, in case we migrate this repo to an org one day or if we want to try the release process in a fork.

.github/workflows/ci.yml
```diff
@@ -7,6 +7,13 @@ on:
   push:
     branches:
       - develop
+  workflow_dispatch:
```
M0NsTeRRR, line 10: missing line break

.github/workflows/ci.yml
```diff
@@ -17,6 +24,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          persist-credentials: false
```
M0NsTeRRR, line 29: missing line break, if you can check if steps are separated for a line break (or we can do it in your other PR as you want)

.github/workflows/release.yml
```diff
@@ -1,120 +1,204 @@
...
       - name: Checkout
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
+          persist-credentials: false
+
       - name: Set up Node.js
         uses: actions/setup-node@v4
         with:
           node-version: 22
```
M0NsTeRRR: make more sense to use `node-version-file` with the package.json

.github/workflows/release.yml
```diff
@@ -1,120 +1,204 @@
...
-        uses: docker/login-action@v3
-        with:
-          registry: ghcr.io
-          username: ${{ github.repository_owner }}
-          password: ${{ secrets.GH_TOKEN }}
+
       - name: Pnpm Setup
         uses: pnpm/action-setup@v4
         with:
           version: 9
```
M0NsTeRRR: Maybe it makes sense to set `packageManager` in `package.json` so the pnpm version is configured only there. Unfortunately, we still have to duplicate it in the `engines` key because of https://github.com/pnpm/action-setup/issues/84

package.json
```diff
@@ -119,8 +119,6 @@
     "@codedependant/semantic-release-docker": "^5.1.0",
```
M0NsTeRRR: Should be configured in `extra_plugins` and `semantic-release` dev dependency can be removed also ?

.github/workflows/release.yml
```diff
@@ -1,120 +1,204 @@
...
+      - name: Extract metadata
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: |
+            fallenbagel/jellyseerr
+            ghcr.io/${{ env.OWNER_LC }}/jellyseerr
+          tags: |
+            type=raw,value=${{ needs.semantic-release.outputs.new_release_version }}
+          labels: |
```
M0NsTeRRR, line 161: Looks like we only need to overwrite `created` ?

https://github.com/docker/metadata-action/blob/master/src/meta.ts#L536

.github/workflows/release.yml
```diff
@@ -1,120 +1,204 @@
...
+
       - name: Install dependencies
         run: pnpm install
+
       - name: Release
+        id: release
+        uses: cycjimmy/semantic-release-action@v4
+        with:
+          extra_plugins: |
+            @semantic-release/git
```
M0NsTeRRR: does it make sense to pin the major at least ?

https://github.com/cycjimmy/semantic-release-action?tab=readme-ov-file#extra_plugins

package.json
```diff
@@ -208,28 +206,12 @@
     "plugins": [
       "@semantic-release/commit-analyzer",
       "@semantic-release/release-notes-generator",
-      [
```
M0NsTeRRR, line 211: From what I read here https://github.com/cycjimmy/semantic-release-action?tab=readme-ov-file#extra_plugins

We have to keep this configuration.

M0NsTeRRR (commented):
I’ve added some comments. After this review, could you test the release process in your fork? We want to avoid breaking anything with all the recent changes.

As for migrating from merge JSON to cache, I’m neutral—I’ve never thought about using a cache for this, but I don’t see any problems with it either, except maybe with cache poisonning.

sudo-kraken:
> I’ve added some comments. After this review, could you test the release process in your fork? We want to avoid breaking anything with all the recent changes.
> 
> As for migrating from merge JSON to cache, I’m neutral—I’ve never thought about using a cache for this, but I don’t see any problems with it either, except maybe with cache poisonning.

I can only validate that the CI jobs and builds run successfully, but not the actual release flow end-to-end in my fork.

On the cache poisoning, that’s a fair concern, but the cache is scoped to the repo and branch/ref, so the risk is minimal. The worst case would be a stale cache causing a rebuild, which the workflow handles. In practice it just speeds up dependency resolution without introducing any trust issues, since the lockfile still pins versions.

M0NsTeRRR:
> I can only validate that the CI jobs and builds run successfully, but not the actual release flow end-to-end in my fork.

Which part? Only Docker Hub will block, right? Maybe you can create another branch and just remove the blocking part so we can validate 80% of the release process better than nothing!

.github/workflows/release.yml
sudo-kraken: But since we still need it in engines for setup-node and again in the workflows, we’d just be duplicating config, just leave this one as is?

M0NsTeRRR: It's easier to manage version from a single file instead of managing inside the package.json and the workflow (so we don't forget it).
Also we can use the workaround here https://github.com/pnpm/action-setup/issues/84#issuecomment-1576586358, that would be sufficient in our case.

package.json
sudo-kraken: https://github.com/fallenbagel/jellyseerr/pull/1904/commits/4cfa65cefaa7ab77e384a7aaef8ce8a9e5a1891c

.github/workflows/release.yml
sudo-kraken: https://github.com/fallenbagel/jellyseerr/pull/1904/commits/4cfa65cefaa7ab77e384a7aaef8ce8a9e5a1891c

.github/workflows/release.yml
sudo-kraken: https://github.com/fallenbagel/jellyseerr/pull/1904/commits/4cfa65cefaa7ab77e384a7aaef8ce8a9e5a1891c

package.json
sudo-kraken: https://github.com/fallenbagel/jellyseerr/pull/1904/commits/4cfa65cefaa7ab77e384a7aaef8ce8a9e5a1891c

.github/workflows/release.yml
sudo-kraken: https://github.com/fallenbagel/jellyseerr/pull/1904/commits/4cfa65cefaa7ab77e384a7aaef8ce8a9e5a1891c

.github/workflows/release.yml
sudo-kraken: https://github.com/fallenbagel/jellyseerr/pull/1904/commits/4cfa65cefaa7ab77e384a7aaef8ce8a9e5a1891c

.github/workflows/ci.yml
sudo-kraken: https://github.com/fallenbagel/jellyseerr/pull/1904/commits/4cfa65cefaa7ab77e384a7aaef8ce8a9e5a1891c

.github/workflows/ci.yml
sudo-kraken: https://github.com/fallenbagel/jellyseerr/pull/1904/commits/4cfa65cefaa7ab77e384a7aaef8ce8a9e5a1891c

.github/workflows/ci.yml
sudo-kraken: https://github.com/fallenbagel/jellyseerr/pull/1904/commits/4cfa65cefaa7ab77e384a7aaef8ce8a9e5a1891c

sudo-kraken:
> > I can only validate that the CI jobs and builds run successfully, but not the actual release flow end-to-end in my fork.
> 
> Which part? Only Docker Hub will block, right? Maybe you can create another branch and just remove the blocking part so we can validate 80% of the release process better than nothing!

That commit https://github.com/fallenbagel/jellyseerr/pull/1904/commits/4cfa65cefaa7ab77e384a7aaef8ce8a9e5a1891c should resolve everything once you are happy with this one, I will try test it out in my fork. Dont want to test until we are happy with the workflows.

sudo-kraken:
That should be the last one now @M0NsTeRRR had to fix shell inside the node container to use `sh` not `bash` like preview and release workflows use.

sudo-kraken:
Closed PR-1904 and merged into PR-1905.

PR #1916, opened 2025-09-22
fix(ui): remove duplicate download items in manage slide over

InterN0te:
## Description
Prevents duplicate download items from being displayed by filtering based on title uniqueness. This improves UI clarity when the same content appears multiple times in the download status.

### Summary
- Adds `getUniqueByTitle` function to filter duplicate download items based on title
- Applies filtering to both regular and 4K download status lists
- Improves UI clarity by preventing duplicate entries in the manage slide over

### Screenshot

<details><summary>Before </summary><p>
image
</p></details> 

<details><summary>After </summary><p>
image
</p></details> 

### Test plan
- [X] Verify that duplicate download items no longer appear in the manage slide over
- [X] Check that both regular and 4K download lists are properly deduplicated
- [X] Ensure filtering doesn't affect other functionality in the component

### DIFF
--- a/src/components/ManageSlideOver/index.tsx
+++ b/src/components/ManageSlideOver/index.tsx
@@ -35,10 +35,8 @@ import { useIntl } from 'react-intl';
 import useSWR from 'swr';
 
 const filterDuplicateDownloads = (
-  items: DownloadingItem[] | undefined
+  items: DownloadingItem[] = []
 ): DownloadingItem[] => {
-  if (!items || items.length === 0) return [];
-
   const seen = new Set<string>();
   return items.filter((item) => {
     if (seen.has(item.downloadId)) return false;


src/components/ManageSlideOver/index.tsx
```diff
@@ -69,6 +70,50 @@ const isMovie = (movie: MovieDetails | TvDetails): movie is MovieDetails => {
   return (movie as MovieDetails).title !== undefined;
 };
 
+const groupDownloadBySeason = (items: DownloadingItem[] | undefined) => {
+  if (!items || items.length === 0) return [];
+
+  const groupedItems: DownloadingItem[] = [];
+
+  items.forEach(item => {
+    // Grouper par titre ET saison (pas juste par titre)
```
Gauvino: Make the comment in English and not French please

Also apply to: 88, 101

Otherwise, great work simple bug fix but make the app a little bit better over time

gauthier-th (commented):
I can't reproduce your issue in the first place.
This is something #927 should already have fixed. How are you items downloaded?

InterN0te:
It's almost like https://github.com/fallenbagel/jellyseerr/pull/927, but not in the same location: my PR is for the "Manage series", not the "Status Badge"

To reproduce, simply download a torrent for an entire season and go to "Manage Series".

This is a test a just did on develop branch, without my PR :

The status badge only shows one download:
image

But in "Manage Series," I have the same download ten times (I also tried with a season of 8 episodes and it only shows 8 times the download so I think it take only the 10 firsts downloads of a TV show) :
image

gauthier-th:
Ah, I didn't see I forgot to implement it there too.
I'm likely biased, but I find my previous implementation way more clean and less hacky.

Could you please implement the same logic as I did on the #927 PR?

gauthier-th:
Re-opening this since it's an interesting QOL feature.

fallenbagel:
> Ah, I didn't see I forgot to implement it there too.
> I'm likely biased, but I find my previous implementation way more clean and less hacky.
> 
> Could you please implement the same logic as I did on the #927 PR?

@InterN0te any update?

InterN0te:
Hi @gauthier-th @fallenbagel ! Sorry for the delay. I've updated the PR to use the same logic as #927 - filtering by ```downloadId``` instead of title. The implementation now uses a ```filterDuplicateDownloads``` function that keeps only one entry per ```downloadId```, which properly handles season packs where multiple episodes share the same download.

```
const filterDuplicateDownloads = (
  items: DownloadingItem[] | undefined
): DownloadingItem[] => {
  if (!items || items.length === 0) return [];

  const seen = new Set<string>();
  return items.filter((item) => {
    if (!item.downloadId) return true;
    if (seen.has(item.downloadId)) return false;
    seen.add(item.downloadId);
    return true;
  });
};
```

This is consistent with your approach in #927 and should be cleaner than the previous title-based filtering. Let me know if you'd like any changes!

InterN0te:
The result is the same as before, but with (significantly) fewer lines of code.

<details><summary>Screenshot</summary>
<p>

With deduplication:
image


Without deduplication:
image

</p>
</details>

src/components/ManageSlideOver/index.tsx
```diff
@@ -33,6 +34,20 @@ import Link from 'next/link';
...
+const filterDuplicateDownloads = (
+  items: DownloadingItem[] | undefined
+): DownloadingItem[] => {
+  if (!items || items.length === 0) return [];
+
+  const seen = new Set<string>();
+  return items.filter((item) => {
+    if (!item.downloadId) return true;
+    if (seen.has(item.downloadId)) return false;
+    seen.add(item.downloadId);
```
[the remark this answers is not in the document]
fallenbagel: Not relevant. PR description was the old behaviour this PR implemented that was requested to be changed.

src/components/ManageSlideOver/index.tsx
```diff
@@ -33,6 +34,19 @@ import Link from 'next/link';
 import { useIntl } from 'react-intl';
 import useSWR from 'swr';
 
+const filterDuplicateDownloads = (
+  items: DownloadingItem[] | undefined
+): DownloadingItem[] => {
+  if (!items || items.length === 0) return [];
+
```
gauthier-th: I find it cleaner like this:
```suggestion
const filterDuplicateDownloads = (
  items: DownloadingItem[] = []
): DownloadingItem[] => {
```

gauthier-th (approved):

fallenbagel (approved):

PR #1938, opened 2025-09-29
removed the node pnpm cache in favour of the cache we setup

sudo-kraken:
#### Description

Removed the `cache: pnpm` from the node setup step, this was used previously but now that we have a cache built into the workflow that we store and restore, this step is legacy and not required.

### DIFF
--- a/.github/workflows/docs-deploy.yml
+++ b/.github/workflows/docs-deploy.yml
@@ -29,7 +29,6 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version-file: package.json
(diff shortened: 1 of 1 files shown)


fallenbagel (changes requested):
This is there in:
https://github.com/fallenbagel/jellyseerr/blob/764f7b4270861c2efe660284a80dca850012a58d/.github/workflows/release.yml#L33

as well. So releases will fail until its removed. Can you check every workflow and remove it in this pr, please?

fallenbagel (approved):

PR #1948, opened 2025-10-01
feat(webhook): add API key support for webhook notifications

0xSysR3ll:
#### Description

This PR introduces an optional API key field in the webhook settings, allowing users to authenticate to the webhook using the `X-API-Key` header.

image

#### To-Dos

- [ ] Disclosed any use of AI (see our [policy](https://github.com/fallenbagel/jellyseerr/blob/develop/CONTRIBUTING.md#ai-assistance-notice))

#### Issues Fixed or Closed

- Fixes #XXXX

### DIFF
--- a/docs/using-jellyseerr/notifications/webhook.md
+++ b/docs/using-jellyseerr/notifications/webhook.md
@@ -22,6 +22,14 @@ This is typically not needed. Please refer to your webhook provider's documentat
 
 This value will be sent as an `Authorization` HTTP header.
 
+### API Key (optional)
+
+:::info
+Uses the `X-API-Key` header, a de facto standard for webhook authentication.
+:::
+
+This value will be sent as an `X-API-Key` HTTP header.
+
 ### JSON Payload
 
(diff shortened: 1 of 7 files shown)


M0NsTeRRR:
Should we rework this feature and allow users to configure the headers they need? That way, it would cover all future feature requests regarding this.

0xSysR3ll:
> Should we rework this feature and allow users to configure the headers they need? That way, it would cover all future feature requests regarding this.

Since the `X-API-Key` is a de facto standard, I thought it would be better to keep it like this. But yeah, maybe we could implement a custom headers feature; allowing to put literally everything we want, like you said.

gauthier-th:
> > Should we rework this feature and allow users to configure the headers they need? That way, it would cover all future feature requests regarding this.
> 
> Since the `X-API-Key` is a de facto standard, I thought it would be better to keep it like this. But yeah, maybe we could implement a custom headers feature; allowing to put literally everything we want, like you said.

Yeah, it's probably better, some 3rd party apps use other ways to authenticate. That's also what I implemented in *arr.

0xSysR3ll:
> That's also what I implemented in *arr.

Oh, I will check this out !

0xSysR3ll:
Closed in favor of https://github.com/seerr-team/seerr/pull/2230

PR #1959, opened 2025-10-03
 feat(webhook): add jellyfinMediaId and jellyfinUserId to webhook notification

### LINKED ISSUE #1949: Feature Request: Enable marking items from fulfilled request as favorites for the requesting user in jellyfin
### Description

Users that are requesting (too much) a lot might have issues finding their content in jellyfin. Especially if they have requested items a long time ago.

I first thought about creating personalized collections for users in jellyfin and then show their requests there, however there doesn't seem to be private collections in jellyfin. When I looked into it I found out that favorites could be used for that instead since they are specific to a user.

### Desired Behavior

Mark fulfilled requests in jellyseerr as favorites for the user in jellyfin.
Persist the items that have been set as favorites to a store, perhaps the disk or jellyseerr database, so that it is possible for users to remove a requested entry from jellyfin without having it readded during the favorites sync. An alternative solution could be to only trigger this when a request transitions to fulfilled since an item is not expected to transition into the fulfilled states multiple times.

### Additional Context

I have implemented this as an external service using C#, using the jellyseerr and jellyfin APIs, but I think that this could be a feature that could interest more people. I think that it could make sense to include this directly into jellyseerr since jellyseerr already have the context that is required (jellyfin user id, item id).

The relevant endpoint in Jellyfin API to mark an item as favorite is:
POST to /Users/<jellyfin_user_id>/FavoriteItems/<jellyfin_item_id>

### Code of Conduct

Jerra94:
#### Description
Add jellyfinMediaId and jellyfinUserId to webhook notification in order to enable a webhook to be configured to mark an item as favorite for the requesting user via the jellyfin API.

I consulted Gemini to understand the codebase but the solution was fully authored manually by myself.

#### To-Dos

- [X] Disclosed any use of AI (see our [policy](https://github.com/fallenbagel/jellyseerr/blob/develop/CONTRIBUTING.md#ai-assistance-notice))
- [X] Database migration (if required)

#### Issues Fixed or Closed

- Fixes #1949

### DIFF
--- a/server/lib/notifications/agents/webhook.ts
+++ b/server/lib/notifications/agents/webhook.ts
@@ -28,11 +28,14 @@ const KeyMap: Record<string, string | KeyMapFunction> = {
   media_tmdbid: 'media.tmdbId',
   media_tvdbid: 'media.tvdbId',
   media_type: 'media.mediaType',
+  media_jellyfinMediaId: (payload) =>
+    payload.media?.jellyfinMediaId ?? payload.media?.jellyfinMediaId4k ?? '',
   media_status: (payload) =>
     payload.media ? MediaStatus[payload.media.status] : '',
   media_status4k: (payload) =>
     payload.media ? MediaStatus[payload.media.status4k] : '',
   request_id: 'request.id',
+  requestedBy_jellyfinUserId: 'request.requestedBy.jellyfinUserId',
   requestedBy_username: 'request.requestedBy.displayName',
   requestedBy_email: 'request.requestedBy.email',
   requestedBy_avatar: 'request.requestedBy.avatar',
--- a/docs/using-seerr/notifications/webhook.md
+++ b/docs/using-seerr/notifications/webhook.md
@@ -84,13 +84,14 @@ The `{{media}}` will be `null` if there is no relevant media object for the noti
 
 These following special variables are only included in media-related notifications, such as requests.
 
-| Variable             | Value                                                                                                          |
-| -------------------- | -------------------------------------------------------------------------------------------------------------- |
-| `{{media_type}}`     | The media type (`movie` or `tv`)                                                                               |
-| `{{media_tmdbid}}`   | The media's TMDB ID                                                                                            |
-| `{{media_tvdbid}}`   | The media's TheTVDB ID                                                                                         |
-| `{{media_status}}`   | The media's availability status (`UNKNOWN`, `PENDING`, `PROCESSING`, `PARTIALLY_AVAILABLE`, or `AVAILABLE`)    |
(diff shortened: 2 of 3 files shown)


server/lib/notifications/agents/webhook.ts
```diff
@@ -28,11 +28,16 @@ const KeyMap: Record<string, string | KeyMapFunction> = {
   media_tmdbid: 'media.tmdbId',
   media_tvdbid: 'media.tvdbId',
   media_type: 'media.mediaType',
+  media_jellyfinMediaId: (payload) =>
+    payload?.media?.jellyfinMediaId
+      ? payload.media.jellyfinMediaId
+      : payload?.media?.jellyfinMediaId ?? '',
```
gauthier-th: I think it would be cleaner that way:
```suggestion
  media_jellyfinMediaId: (payload) => payload.media?.jellyfinMediaId ?? '',
```

gauthier-th (changes requested):

server/lib/notifications/agents/webhook.ts
Jerra94: Hi,
Sorry, it was not intended. My intention was to use either jellyfinMediaId or jellyfinMediaId4k as the media_jellyfinMediaId and then fallback to an empty string if none of them are set. I have updated this to use that fallback logic now in the PR.

The reasoning behind it is that in the API:s at jellyfin there is no distinction between them and during a jellyseerr fulfilled request either of them would be set. Do you think this makes sense or would you prefer if it would only be jellyfinMediaId?

server/lib/notifications/agents/webhook.ts
```diff
@@ -28,11 +28,16 @@ const KeyMap: Record<string, string | KeyMapFunction> = {
   media_tmdbid: 'media.tmdbId',
   media_tvdbid: 'media.tvdbId',
   media_type: 'media.mediaType',
+  media_jellyfinMediaId: (payload) =>
+    payload?.media?.jellyfinMediaId
+      ? payload.media.jellyfinMediaId
+      : payload?.media?.jellyfinMediaId4k ?? '',
```
gauthier-th:
```suggestion
  media_jellyfinMediaId: (payload) =>
    payload.media?.jellyfinMediaId ?? payload.media?.jellyfinMediaId4k ?? '',
```

gauthier-th (changes requested):

server/lib/notifications/agents/webhook.ts
gauthier-th: Ah yes ok, it makes sense.

server/lib/notifications/agents/webhook.ts
Jerra94: I have updated the PR according to this now

gauthier-th (dismissed):

gauthier-th (approved):

M0NsTeRRR (approved):

PR #1960, opened 2025-10-04
ci: add stale bot

M0NsTeRRR:
#### Description
1. Check for any issues and PRs with the label **"pending author's response"**. I think we can reuse the label **"more information needed"** and rename it to **"pending author's response"** so we can use the same label for both issues and PRs.
2. Add a **"stale"** label to issues/PRs after **30 days** of inactivity.
3. Issues/PRs with the **"stale"** label for **30 days** will be automatically closed.
4. Issues with the **"confirmed"** label will be ignored. I don’t think we need something like that for PRs.

If an issue/PR is updated (e.g., a new comment is added), the **"stale"** label is removed.

@sudo-kraken feedback ?


gauthier-th:
For 3. I think we shouldn't close PRs when there is no activity. We're sometimes slow to merge them but we eventually do it. 
And I think we could increase the 30 days delay? Seems a bit short to me.

M0NsTeRRR:
> For 3. I think we shouldn't close PRs when there is no activity. We're sometimes slow to merge them but we eventually do it. And I think we could increase the 30 days delay? Seems a bit short to me.

The stale workflow only works when the PR has the **"pending author's response"** label. In my opinion, this label should be added when you review someone’s work and, after **3 months of inactivity**, you add it so the PR will eventually be auto-closed.

The **30 days** refers to the time after the **"stale"** label is added. The **entire process** before closing takes **60 days**:
- **First 30 days**: No activity → **"stale"** label is added.
- **Next 30 days**: If still inactive → PR is auto-closed.

Which one you want to increase ?

gauthier-th:
> The stale workflow only works when the PR has the **"pending author's response"** label. In my opinion, this label should be added when you review someone’s work and, after **3 months of inactivity**, you add it so the PR will eventually be auto-closed.

Ok it makes sense, this looks ok to me.

> The **entire process** before closing takes **60 days**

Same, I didn't understand it that way, 60 days for the whole process is fine.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,31 @@
...
+concurrency:
+  group: close-stale-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  stale:
+    name: Close stale issues and PRs
+    runs-on: ubuntu-24.04
+    permissions:
+      actions: write
```
sudo-kraken, line 20: is actions: write needed here

.github/workflows/stale.yml
```diff
@@ -0,0 +1,31 @@
...
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v10.1.0
+        with:
+          any-of-labels: "pending author's response"
+          days-before-stale: 30
+          days-before-close: 30
+          stale-issue-label: 'stale'
+          exempt-issue-labels: 'confirmed'
+          stale-pr-label: 'stale'
```
sudo-kraken, line 33: Do we want to add some messages when we mark as stale etc
[stale-issue-message](https://github.com/actions/stale#stale-issue-message)
[stale-pr-message](https://github.com/actions/stale#stale-pr-message)

.github/workflows/stale.yml
M0NsTeRRR: Unfortunately yes
https://github.com/actions/stale/issues/1159

.github/workflows/stale.yml
M0NsTeRRR: I didn’t add them because I think when someone adds the `pending author's response` label, they will also include a message explaining what is missing.

gauthier-th: We should, because no notification is sent when adding/removing labels

sudo-kraken:
Other than what I commented on it all looks good to me, good work @M0NsTeRRR

.github/workflows/stale.yml
sudo-kraken: That is annoying we will have to leave it in, and review later in a few months time

.github/workflows/stale.yml
gauthier-th: > I didn’t add them because I think when someone adds the `pending author's response` label, they will also include a message explaining what is missing.

Just in case someone forgot, wouldn't it be better to send the message every time?

sudo-kraken: It wouldnt hurt to add them, if it becomes to noisy we can remove later. @M0NsTeRRR what do you think.

M0NsTeRRR: I added a message on each event, does it looks good for both of you ?

sudo-kraken: Yep looks good 👍

PR #1961, opened 2025-10-04
build(docker): enable corepack so we don't need to pin pnpm version

M0NsTeRRR:
#### Description

Enable corepack in dockerfile so we don't to install pnpm and pin the version inside the dockerfile (less maintenance) :)

https://pnpm.io/docker

```
STEP 7/8: RUN pnpm install
! Corepack is about to download https://registry.npmjs.org/pnpm/-/pnpm-10.17.1.tgz
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,14 @@
 FROM node:22-alpine AS BUILD_IMAGE
 
-WORKDIR /app
-
 ARG SOURCE_DATE_EPOCH
 ARG TARGETPLATFORM
 ARG COMMIT_TAG
 ENV TARGETPLATFORM=${TARGETPLATFORM:-linux/amd64}
 ENV COMMIT_TAG=${COMMIT_TAG}
+
+ENV PNPM_HOME="/pnpm"
+ENV PATH="$PNPM_HOME:$PATH"
+RUN corepack enable
 
 RUN \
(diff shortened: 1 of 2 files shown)


gauthier-th:
Did you also see the remark about `pnpm fetch` in the example 3? https://pnpm.io/docker#example-3-build-on-cicd

M0NsTeRRR:
> Did you also see the remark about `pnpm fetch` in the example 3? https://pnpm.io/docker#example-3-build-on-cicd

No, I don't think we are affected as we use `gha` cache for the build on `docker/build-push-action`, @sudo-kraken you confirm ?

sudo-kraken:
> > Did you also see the remark about `pnpm fetch` in the example 3? https://pnpm.io/docker#example-3-build-on-cicd
> 
> No, I don't think we are affected as we use `gha` cache for the build on `docker/build-push-action`, @sudo-kraken you confirm ?

We are not affected we persist the image layers between runs so the pnpm layers are stored and reused between runs from gha cache.

M0NsTeRRR:
I've also fixed a warning detected in the CI ctrl f "Annotations" inside https://github.com/fallenbagel/jellyseerr/actions/runs/18246232969

https://docs.docker.com/reference/build-checks/stage-name-casing/

gauthier-th (approved):
LGTM

PR #1979, opened 2025-10-06
ci: fix seerr reference in workflow run

sudo-kraken:
ci: fix seerr reference in workflow run


fallenbagel:
Wait why did cypress fail

PR #2019, opened 2025-10-08
feat: Overseerr to Jellyseerr migration

gauthier-th:
#### Description

This PR includes the migration script to convert an Overseerr instance into a Jellyseerr one.
The database migration is included in the `server/lib/overseerrMerge.ts` file, and has been generated using TypeORM. This migration will always be run after the `UpdateWebPush1745492372230` migration, so that Overseerr users migrating much later doesn't run a new Seerr migration before the Overseerr->Seerr migration.

It also includes some renames of Jellyseerr -> Seerr I wrote when creating the first version of the migration.

This is what have been used by beta-testers to migrate their Overseerr instance.

#### To-Dos

### DIFF
--- a/gen-docs/src/components/SeerrVersion/index.tsx
+++ b/gen-docs/src/components/SeerrVersion/index.tsx
@@ -7,7 +7,7 @@ export const SeerrVersion = () => {
     async function fetchVersion() {
       try {
         const response = await fetch(
-          'https://raw.githubusercontent.com/fallenbagel/jellyseerr/main/package.json'
+          'https://raw.githubusercontent.com/seerr-team/seerr/main/package.json'
         );
 
         const data = await response.json();


charts/seerr-chart/templates/deployment.yaml
```diff
@@ -0,0 +1,121 @@
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: {{ include "seerr.fullname" . }}
+  labels:
+    {{- include "seerr.labels" . | nindent 4 }}
+spec:
+  serviceName: {{ include "jellyseerr.fullname" . }}
```
fallenbagel:
```suggestion
  serviceName: {{ include "seerr.fullname" . }}
```

docs/README.md
```diff
@@ -26,16 +26,10 @@ Welcome to the Jellyseerr Documentation.
 - Easily **Watchlist** or **Blacklist** media.
 - More features to come!
 
-## Motivation
-
-The primary motivation for starting Jellyseerr was to bring Jellyfin and Emby support to Overseerr. However, over time, **Jellyseerr** has evolved into its own distinct application with unique features. Designed as a one-stop shop for media requests, it offers a simple, easy-to-use experience for managing requests on Jellyfin, Emby, and Plex servers.
-
 ## We need your help!
 
 [Jellyseerr](https://github.com/fallenbagel/jellyseerr) is an ambitious project where developers/contributors poured a lot of work into, and that builds on top of [Overseerr](https://github.com/sct/overseerr). And we have a lot more to do as well.
```
fallenbagel: This needs to be reworked. We can say that seerr is the merge if these two projects etc

docs/using-seerr/backups.md
```diff
@@ -19,20 +19,20 @@ Apart from the settings, all other data—including user accounts, media request
 
 ### SQLite
 
-If your backup system uses filesystem snapshots (such as Kubernetes with Volsync), you can directly back up the Jellyseerr data folder.  
-Otherwise, you need to stop the Jellyseerr application and back up the `config` folder.
+If your backup system uses filesystem snapshots (such as Kubernetes with Volsync), you can directly back up the JellSeerryseerr data folder.
```
fallenbagel:
```suggestion
If your backup system uses filesystem snapshots (such as Kubernetes with Volsync), you can directly back up the seerr data folder.  
```

fallenbagel (changes requested):
Noticed some places where there were typos. 

Also one thing I'm concerned with is, with this merge, the docs will reference images that are currently not out there yet. Maybe we should add a temporary thing until then? We could add tabs in the installation docs for `jellyseerr` and `overseerr` and `seerr`. We can remove the first two tabs once the merge is fully completed

Wdyt? @sct @gauthier-th

charts/seerr-chart/templates/deployment.yaml
M0NsTeRRR, line 1: This file should not exist anymore, I think you did something wrong on rebase

M0NsTeRRR (changes requested):

charts/seerr-chart/templates/deployment.yaml
gauthier-th: Yup, I did a lot of rebase, may have made a mistake somewhere

M0NsTeRRR:
> Also one thing I'm concerned with is, with this merge, the docs will reference images that are currently not out there yet

Yes that’s why I proposed releasing the 2.7.3 image (in seerr-team/seerr) on the same commit as Jellyseerr, so everything can be updated prior to the next (first) release without breaking anything for existing jellyseerr users.

fallenbagel:
> > Also one thing I'm concerned with is, with this merge, the docs will reference images that are currently not out there yet
> 
> Yes that’s why I proposed releasing the 2.7.3 image (in seerr-team/seerr) on the same commit as Jellyseerr, so everything can be updated prior to the next (first) release without breaking anything for existing jellyseerr users.
> 
> 
2.7.3 won't work though as it wont have the overseerr migration

gauthier-th:
> Also one thing I'm concerned with is, with this merge, the docs will reference images that are currently not out there yet. Maybe we should add a temporary thing until then? We could add tabs in the installation docs for `jellyseerr` and `overseerr` and `seerr`. We can remove the first two tabs once the merge is fully completed

Would disabling the CI to publish the docs be a solution for now, until the merge is completed?

fallenbagel:
> > Also one thing I'm concerned with is, with this merge, the docs will reference images that are currently not out there yet. Maybe we should add a temporary thing until then? We could add tabs in the installation docs for `jellyseerr` and `overseerr` and `seerr`. We can remove the first two tabs once the merge is fully completed
> 
> Would disabling the CI to publish the docs be a solution for now, until the merge is completed?

That would work too

M0NsTeRRR:
> 2.7.3 won't work though as it wont have the overseerr migration

No issue for me this image isn’t meant to handle the migration. The 3.0.0 release could be the first to do so. Currently, there’s no information confirming that the Seerr image is compatible with Overseerr, except for the preview-seerr tag from Jellyseerr.

fallenbagel:
> > 2.7.3 won't work though as it wont have the overseerr migration
> 
> No issue for me this image isn’t meant to handle the migration. The 3.0.0 release could be the first to do so. Currently, there’s no information confirming that the Seerr image is compatible with Overseerr, except for the preview-seerr tag from Jellyseerr.

That reminds me @gauthier-th we should merge this pr as a "breaking change" commit with that added. This would allow semantic version bot to release next version as 3.0.0

M0NsTeRRR:
> > > 2.7.3 won't work though as it wont have the overseerr migration
> > 
> > 
> > No issue for me this image isn’t meant to handle the migration. The 3.0.0 release could be the first to do so. Currently, there’s no information confirming that the Seerr image is compatible with Overseerr, except for the preview-seerr tag from Jellyseerr.
> 
> That reminds me @gauthier-th we should merge this pr as a "breaking change" commit with that added. This would allow semantic version bot to release next version as 3.0.0

@sudo-kraken and me gonna rework the release part. As we use convetionnal commit you just need to commit with a `!` and the changelog (git-cliff used on the other repo, e.g mobile app) will handle it :)

docs/README.md
TheCatLady:
```suggestion
[Seerr](https://github.com/seerr-team/seerr) is an ambitious project developers/contributors poured a lot of work into, and we still have a lot more to do.
```

gauthier-th: I previously reworked it like this:

> [Seerr](https://github.com/seerr-team/seerr) is an ambitious project where developers/contributors poured a lot of work into, and we have a lot more to do as well. Seerr is the result of a collaborative effort between the original Overseerr project and the Jellyseerr fork, created to deliver an excellent request management solution for both Plex and Jellyfin users.

WDYT?

fallenbagel: Jellyin and emby users*

TheCatLady: @gauthier-th LGTM! Could you apply my edits to the first sentence though? The grammar is a bit odd in the original.

gauthier-th: Updated

server/lib/overseerrMerge.ts
```diff
@@ -0,0 +1,535 @@
...
+    });
+    process.exit(1);
+  }
+
+  // Manually run the migration to update the database schema
+  if (process.env.NODE_ENV === 'production') {
+    await dbConnection.query('PRAGMA foreign_keys=OFF');
+    await dbConnection.runMigrations();
+    await dbConnection.query('PRAGMA foreign_keys=ON');
+  }
```
OwsleyJr, line 90: Do we want to run a full checkpoint here and update the main db file if Overseerr users decide to move to postgres

gauthier-th: I didn't handle it for Postgres since it's not possible to use Postgres in Overseerr.
If Overseeerr users want to use Postgres, they should do:

- Overseerr (sqlite) -> migration -> Seerr (sqlite) -> Seerr (postgres)

So nothing to worry about that, since the migration will not see any Postgres database.

M0NsTeRRR: We will add to the migration documentation (I need to submit the PR) that this is the only supported migration path.

gauthier-th:
Not again

M0NsTeRRR (dismissed):

sudo-kraken (approved):

server/datasource.ts
```diff
@@ -74,10 +74,10 @@ const postgresDevConfig: DataSourceOptions = {
     : parseInt(process.env.DB_PORT ?? '5432'),
   username: process.env.DB_USER,
   password: process.env.DB_PASS,
-  database: process.env.DB_NAME ?? 'jellyseerr',
+  database: process.env.DB_NAME ?? 'seerr',
   ssl: buildSslConfig(),
   synchronize: false,
-  migrationsRun: true,
```
fallenbagel, line 80: Just curious, why was this changed?

fallenbagel (commented):
Wanted to ask about the migrations part. @gauthier-th why was the devconfig for postgres changed?

server/datasource.ts
gauthier-th: I changed it while messing around with the migrations, and didn't change it back. I believe this was a misconfiguration introduced with postgres support (see the 3 other configs for sqlite dev, sqlite prod and postgres prod having the value at `false`).

fallenbagel (approved):
LGTM

server/lib/overseerrMerge.ts
OwsleyJr: I meant for Overseerr users that have already migrated to Seerr but we probably don't need to worry about that for the initial migration. Can just add to the documentation as @M0NsTeRRR mentioned.

sudo-kraken:
@M0NsTeRRR once this merges we will need to raise a pr to update the verification docs link in the docker.mdx, or we can push that with the helm verification docs after the helm pr merges.

OwsleyJr (approved):
LGTM as well

server/datasource.ts
0xSysR3ll: Actually, I think it's needed in a dev env when using postgres. Just confirmed it.

gauthier-th: Wdym it's needed? It's needed to be enabled or disabled?

fallenbagel: > Wdym it's needed? It's needed to be enabled or disabled?

I think it needs to be enabled. I have not tested it myself but that's what I understood from that comment

0xSysR3ll: > Wdym it's needed? It's needed to be enabled or disabled?

Sorry, what I meant by "needed" is that `migrationsRun` should be set to `true` - as it was before.

gauthier-th: Ok, my bad then, I'll revert this. But why is this not the same for SQLite though?

0xSysR3ll: From what I understand, in production mode we work with a "pre-initialized" database, so there's no need for either `synchronize` or `migrationsRun`.
In development mode, however, we need to build the schema - for sqlite, `synchronize: true` is enough, whereas postgres requires `migrationsRun: true` instead.

gauthier-th: The thing I don't understand is why that difference between Postgres and Sqlite? Doesn't makes sense to run the migration in dev only for Postgres.

0xSysR3ll: > The thing I don't understand is why that difference between Postgres and Sqlite? Doesn't makes sense to run the migration in dev only for Postgres.

I guess that's another typeORM mystery 🤷‍♂️

gauthier-th: > Actually, I think it's needed in a dev env when using postgres. Just confirmed it.

I don't think it's needed. I think `synchronize: true` is needed on a dev env:
- `synchronize`: whether TypeORM will modify the schema whenever something doesn't match. Unsafe for production, but useful for development. ([source](https://typeorm.io/docs/advanced-topics/migrations/#how-migrations-work))
- `runMigrations`: whether we should run the migrations when TypeORM is initialized ([source](https://github.com/typeorm/typeorm/blob/master/src/data-source/BaseDataSourceOptions.ts#L106-L110)). We need this to be `false` in production, because we already run it when initializing the connection [here](https://github.com/seerr-team/seerr/blob/develop/server/index.ts#L65-L73). In a development environment, we don't need the migrations because `synchronize` will do it.

Let me make this change in another PR, this is not relevant in this one.

0xSysR3ll: > In a development environment, we don't need the migrations because synchronize will do it.

That's the thing, from what I tested, `synchronize` seems to only work with SQLite.

But yeah let's keep this modification for another PR.

gauthier-th: [This](https://github.com/typeorm/typeorm/issues/7992#issuecomment-1332816345) could be a solution if the issue persists.

0xSysR3ll (dismissed):

docs/getting-started/docker.mdx
```diff
@@ -40,15 +40,15 @@ For details on the Docker CLI, please [review the official `docker run` document
...
+  --name seerr \
   --init \
   -e LOG_LEVEL=debug \
   -e TZ=Asia/Tashkent \
   -e PORT=5055 \
   -p 5055:5055 \
   -v /path/to/appdata/config:/app/config \
   --restart unless-stopped \
-  ghcr.io/fallenbagel/jellyseerr:latest
+  ghcr.io/seerr-team/seerr
```
sudo-kraken:
```suggestion
  ghcr.io/seerr-team/seerr:latest
```

sudo-kraken: as this is what we suggest in the compose should we not keep it uniform and include the latest tag? like [here](https://github.com/seerr-team/seerr/pull/2019/files#diff-65c3f75b10f128d81ff0f4b985972f886e8a79b341f9df224f5f25ef9b5ae1c3R98)

docs/getting-started/docker.mdx
```diff
@@ -68,19 +68,19 @@ To run the container as a specific user/group, you may optionally add `--user=[
 
 Stop and remove the existing container:
 ```bash
-docker stop jellyseerr && docker rm jellyseerr
+docker stop seerr && docker rm seerr
 ```
 Pull the latest image:
 ```bash
-docker pull ghcr.io/fallenbagel/jellyseerr:latest
+docker pull ghcr.io/seerr-team/seerr
```
sudo-kraken:
```suggestion
docker pull ghcr.io/seerr-team/seerr:latest
```

sudo-kraken: and here

docs/getting-started/docker.mdx
```diff
@@ -142,31 +142,31 @@ Please refer to the [Docker Desktop for Windows user manual](https://docs.docker
...
   --init \
   -e LOG_LEVEL=debug \
   -e TZ=Asia/Tashkent \
   -e PORT=5055 \
   -p 5055:5055 \
-  -v jellyseerr-data:/app/config \
+  -v seerr-data:/app/config \
   --restart unless-stopped \
-  ghcr.io/fallenbagel/jellyseerr:latest
+  ghcr.io/seerr-team/seerr
```
sudo-kraken:
```suggestion
  ghcr.io/seerr-team/seerr:latest
```

sudo-kraken: and here

docs/getting-started/docker.mdx
fallenbagel: I agree

docs/getting-started/docker.mdx
fallenbagel: I agree

docs/getting-started/docker.mdx
fallenbagel: I agree

server/lib/overseerrMerge.ts
```diff
@@ -0,0 +1,535 @@
...
+
+  // MediaStatus.Blacklisted was added before MediaStatus.Deleted in Jellyseerr
+  try {
+    const mediaRepository = getRepository(Media);
+    const mediaToUpdate = await mediaRepository.find({ where: { status: 6 } });
+    for (const media of mediaToUpdate) {
+      media.status = 7;
+      await mediaRepository.save(media);
+    }
```
OwsleyJr, line 99: Need to account for 4K media as well.

gauthier-th: Good catch, thanks

gen-docs/src/components/SeerrVersion/index.tsx
sct: This will need to be updated.

TheCatLady (approved):

PR #2032, opened 2025-10-12
build(docker): setup rootless image

M0NsTeRRR:
#### Description

Setup rootless docker image


docs/getting-started/docker.mdx
```diff
@@ -31,6 +31,7 @@ For details on the Docker CLI, please [review the official `docker run` document
 ```bash
 docker run -d \
   --name jellyseerr \
```
sudo-kraken, line 33:
```suggestion
  --name seerr\
```

docs/getting-started/docker.mdx
```diff
@@ -86,6 +87,7 @@ services:
   jellyseerr:
     image: fallenbagel/jellyseerr:latest
```
sudo-kraken, line 88:
```suggestion
    image: ghcr.io/seerr-team/seerr:latest
```

docs/getting-started/docker.mdx
```diff
@@ -86,6 +87,7 @@ services:
...
     container_name: jellyseerr
```
sudo-kraken, line 90:
```suggestion
    container_name: seerr
```

docs/getting-started/docker.mdx
```diff
@@ -156,6 +158,7 @@ Then, create and start the Jellyseerr container:
 ```bash
 docker run -d \
   --name jellyseerr \
```
sudo-kraken, line 160:
```suggestion
  --name seerr \
```

docs/getting-started/docker.mdx
```diff
@@ -194,6 +197,7 @@ services:
   jellyseerr:
     image: fallenbagel/jellyseerr:latest
```
sudo-kraken, line 198:
```suggestion
    image: ghcr.io/seerr-team/seerr:latest
```

docs/getting-started/docker.mdx
```diff
@@ -86,6 +87,7 @@ services:
...
     container_name: jellyseerr
+    user: 1000:1000
```
sudo-kraken:
```suggestion
```

sudo-kraken: Dockerfile already specifies `USER 1000:1000`?

docs/getting-started/docker.mdx
```diff
@@ -31,6 +31,7 @@ For details on the Docker CLI, please [review the official `docker run` document
 ```bash
 docker run -d \
   --name jellyseerr \
+  --user 1000:1000 \
```
sudo-kraken:
```suggestion
```

sudo-kraken: Dockerfile already specifies `USER 1000:1000`?

package.json
```diff
@@ -22,7 +22,7 @@
     "typecheck": "pnpm typecheck:server && pnpm typecheck:client",
     "typecheck:server": "tsc --project server/tsconfig.json --noEmit",
     "typecheck:client": "tsc --noEmit",
-    "prepare": "husky install",
+    "prepare": "husky install || true",
```
gauthier-th: Why?

docs/getting-started/docker.mdx
M0NsTeRRR: It's handled by PR #2019

package.json
M0NsTeRRR: it's needed because `pnpm install --prod` use prepare step and ignore-script cannot be used as it's used for sqlite package

See this bug https://github.com/pnpm/pnpm/issues/7068

gauthier-th: Yes, but then the prepare is going to succeed even if husky fails on development setup?

M0NsTeRRR: I agree, it’s not the best approach. One thing I had in mind but didn’t include in this PR is to completely remove Husky from the prepare and document its setup instead in `CONTRIBUTING.md`. Husky/pre-commit is a dev tool and should be opt-in, not forced. We should still enforce best practices in CI, of course.

Personally, I struggle with Git hooks, so I’d rather avoid forcing them on everyone.

For this, I see two options:

- Leave it as is and address it in a separate PR (separate topic).
- I can add a `NODE_ENV=production` in the dockerfile and add a condition instead of  `| true` in the `package.json`

gauthier-th: > is to completely remove Husky from the prepare and document its setup instead in `CONTRIBUTING.md`

I'm not sure removing Husky and git hooks would really help us.
Git hooks have their caveats, but at least it's forcing people to run the dev tools before sending a PR. For instance I would be the first to often forgot to run the `pnpm format` command before sending my PR.
And it would also remove the commit message helper.

What's the struggle you have with git hooks?

> I can add a NODE_ENV=production in the dockerfile and add a condition instead of | true in the package.json

Anyway, we have to keep it mind that it should also supports Windows and not only Unix. I don't know Windows scripting very well but I don't think `command | true` is a valid thing.

M0NsTeRRR: > I'm not sure removing Husky and git hooks would really help us.

When I say 'remove', I mean removing the default automatic installation. It will be opt-in, not forced. The user will need to manually run husky install to get the same setup.

I don’t like the conventional commit setup because it slows me down. I prefer using my Vim setup to directly write commit messages, and I’d rather run pnpm format at the end of the PR just before the review instead of on every commit.

gauthier-th: > When I say 'remove', I mean removing the default automatic installation. It will be opt-in, not forced. The user will need to manually run husky install to get the same setup.

Same thing. I'm pretty sure most of users will forget to run it.

> I don’t like the conventional commit setup because it slows me down.

I get that, but you can just ctrl+c the husky form and go directly into the git commit message editor, with your configured editor of choice.

M0NsTeRRR: > Same thing. I'm pretty sure most of users will forget to run it.

The CI is here to check that, you forgot to run the linter, the CI is red we know it's not ready for review, simple as that.

For the message yeah for the other part no ^^

danshilm: I really want to not remove husky, the downsides of removing it far outweight the upsides imho.

In my personal projects, I do something like this to handle the issue being faced:

```javascript
#!/usr/bin/env node

/**
 * Do not run husky in CI environments
 */
const isCi = process.env.CI !== undefined;
if (!isCi) {
  require('husky').install();
}
```

And I use this JS script in the npm script `prepare` command.

```
"prepare": "node bin/prepare.js",
```

I found that snippet a long while ago on the issues in the husky repo and use it in several projects now.

gauthier-th: I also prefer the snippet solution instead of removing husky completely. Let's see what the rest of the team thinks about this

sudo-kraken (dismissed):
LGTM. Great work!

PR #2039, opened 2025-10-13
refactor(notifications): replace webhook KeyMap with LiquidJS templating

### LINKED ISSUE #1876: Email customization
### Description

Hello,

Would it be possible to add a section for customizing email notifications (and notifications in general)?
The idea would be to be able to use dynamic variables (e.g., user name, title of the requested movie, etc.) to personalize the content of the message sent to customers.

This would allow the email to be adapted to the context and provide a better user experience.

Thank you in advance!

### Desired Behavior

Description
Add a section for customizing email notifications (and ideally notifications in general).

Example of use
Be able to insert dynamic variables into the message content, such as:

{{user_name}} → User name
{{requested_movie}} → Title of the requested movie
etc.

Benefits
Offer a more personalized experience to customers
Improve the clarity and relevance of emails sent

### Additional Context

_No response_

### Code of Conduct

mmenanno:
## Description

This PR refactors the webhook notification agent to use a shared **LiquidJS**-based templating engine instead of the bespoke `KeyMap` + `parseKeys` substitution system, providing a foundation for standardized templating across all notification agents.

This replaces the original Handlebars approach from earlier in this PR, following the maintainer discussion ([review thread](https://github.com/seerr-team/seerr/pull/2039#discussion_r3391800589)): Liquid is a better fit for an end-user template engine because it ships the logic/filters users need out of the box, so we don't have to implement and maintain custom helpers.

This is part of the larger effort to enable custom templates across notification agents; it lays the groundwork for extending template support to all agents in future PRs. No specific issue to link.

### Key Changes

- **New `TemplateEngine`** (`server/lib/notifications/templateEngine.ts`) wrapping a shared LiquidJS engine.
- **No custom helpers needed** — Liquid provides conditionals, comparisons (`==`, `>`, …), and filters (`default`, `upcase`, `downcase`, `capitalize`, `truncate`, `json`, `url_encode`) natively. This also resolves several review points for free: numeric comparison coercion, `default` on empty strings, and no HTML-escaping (apostrophes/quotes render verbatim).
- **Circular-safe `json` filter** — the built-in `json` filter is overridden so circular TypeORM relations (e.g. `Media` ↔ `Issue`) can't crash rendering.
- **`buildContext`** exposes the same scalar variables as the old `KeyMap` (including `discordIds`, `media_imdbid`, `jellyfinMediaId`, `plexRatingKey`, `requestedBy_jellyfinUserId`) plus the full `media`/`request`/`issue`/`comment`/`extra` values for conditionals.
- **Simpler payload build** — the stored template is rendered to a string and handed to `JSON.parse`; the previous regex "cleanup" of the rendered payload was removed.
- **Relaxed validation** (client + server) — the payload is now a Liquid template that needn't be valid JSON until rendered, so the strict `JSON.parse` checks on save/test were removed. Templates are validated by using the **Test** button.
- **`customHeaders` support retained.**
- **Unit tests** added for `TemplateEngine` (escaping, optional-section `null` handling, circular-safe `json`, and variable mapping).

### New Templating Capabilities

Webhook payloads and URLs can now use Liquid:

- **Conditionals:** `{% if media %}…{% else %}…{% endif %}`, `{% if media_status == "AVAILABLE" %}…{% endif %}`
- **Comparisons:** `{% if pendingRequestsCount > 10 %}…{% endif %}`
- **Filters:** `{{ notifyuser_username | capitalize }}`, `{{ media_type | upcase }}`, `{{ subject | truncate: 50 }}`, `{{ subject | json }}`, `{{ value | url_encode }}`
- **Defaults:** `{{ notifyuser_username | default: "Unknown User" }}`

The default payload now uses `{% if %}…{% else %}null{% endif %}` so optional sections (media/request/issue/comment) render as `null` when absent (matching prior behavior) while staying valid JSON, and emits free-text values (titles, overviews, display names, comments) through the `json` filter so quotes/newlines can't break the surrounding JSON.

### Compatibility Notes

- Simple `{{ variable }}` substitution continues to work unchanged.
- This is **not** a drop-in for every existing custom payload: templates are now Liquid, so the old "magic" object keys (`"{{media}}"`, `"{{request}}"`, …) should be rewritten as `{% if %}` blocks (the **Reset to Default** button produces an up-to-date example). Filter names follow Liquid (`upcase`/`downcase` rather than `upper`/`lower`).
- URL variable substitution renders raw; use the `url_encode` filter where encoding is needed.

### Review Feedback Addressed

- Switched the engine from Handlebars to **LiquidJS** (gauthier-th / M0NsTeRRR).
- Removed the regex JSON "cleanup"; the render output is parsed directly (M0NsTeRRR).
- A non-JSON template that renders to valid JSON is now allowed end-to-end (relaxed client + server validation) (M0NsTeRRR).
- Removed the frontend payload-validation test that could no longer verify templates (M0NsTeRRR).
- Dropped the `notifyUser` → requester fallback so `notifyuser_*` maps to `notifyUser` only, matching the previous `KeyMap` (M0NsTeRRR).
- Restored an escaping path for free-text values, and `url_encode` is available for URL templating.

## How Has This Been Tested?

Tested locally with `pnpm dev` against a [webhook.site](https://webhook.site) endpoint, plus automated unit tests.

Manual:
1. Configure the webhook agent against the test endpoint, enable notification types, and **Save**.
2. Click **Test** and confirm the webhook is received with a valid JSON body and variables substituted.
3. Trigger real notifications (request, auto-approve, issue, comment) and confirm optional sections render as objects when relevant and `null` otherwise.
4. Edge cases verified: a media title containing quotes, a multi-line issue/comment message, and templates using conditionals/filters — all produce valid JSON.

Automated: `pnpm test server/lib/notifications/templateEngine.test.ts` (escaping, optional-section `null` handling, circular-safe `json`, and `buildContext` variable mapping). `pnpm build`, `pnpm typecheck`, and `pnpm lint` all pass.

N/A — backend refactor (frontend default payload/validation updated, UI unchanged).

## Checklist:

- [x] Disclosed any use of AI (see our [policy](https://github.com/seerr-team/seerr/blob/develop/CONTRIBUTING.md#ai-assistance-notice)) — **AI (Claude Code) used for implementation assistance**
- [ ] I have updated the documentation accordingly. — docs for the new Liquid template syntax to follow
- [ ] Translation keys `pnpm i18n:extract` — no new translation keys added
- [ ] Database migration (if required) — not required (uses existing settings system)

## Summary by CodeRabbit

* **New Features**
  * Webhook notifications now support Liquid templates for JSON payloads and webhook URLs.
  * Templates can use notification details, related media, requests, issues, comments, and extra data.

* **Bug Fixes**
  * Improved validation for template-based and standard JSON payloads.
  * Webhook settings are now handled more consistently during retrieval, updates, and upgrades.

* **Documentation**
  * Expanded webhook documentation with Liquid examples, supported filters, conditionals, error handling, and available template objects.

### DIFF
--- a/docs/using-seerr/notifications/webhook.md
+++ b/docs/using-seerr/notifications/webhook.md
@@ -47,13 +47,13 @@ Payloads and webhook URLs are rendered with [LiquidJS](https://liquidjs.com/), s
 
 Render a section only when the relevant object is present, falling back to `null` otherwise:
 
-```
+```liquid
 "media": {% if media %}{ "tmdbId": "{{ media_tmdbid }}" }{% else %}null{% endif %}
 ```
 
 You can also branch on a value:
 
-```
+```liquid
 {% if media_status == "AVAILABLE" %}...{% else %}...{% endif %}
 ```
 
--- a/docs/using-seerr/notifications/webhook.md
+++ b/docs/using-seerr/notifications/webhook.md
@@ -35,9 +35,56 @@ You cannot configure both the **Authorization Header** field and a custom `Autho
 
 ### JSON Payload
 
-Customize the JSON payload to suit your needs. Seerr provides several [template variables](#template-variables) for use in the payload, which will be replaced with the relevant data when the notifications are triggered.
+Customize the JSON payload to suit your needs. The payload is a [Liquid](https://liquidjs.com/) template that is rendered when a notification is triggered and then sent as the request body. Seerr exposes a set of [template variables](#template-variables), plus the full `media`, `request`, `issue`, and `comment` objects, for use in [conditionals and filters](#template-syntax).
+
+The template only needs to _render_ to valid JSON — it does not need to be valid JSON itself (for example, optional sections can be wrapped in `{% if %}` blocks). Use the **Test** button to confirm your template renders correctly; template errors are also written to the server logs.
+
+## Template Syntax
+
+Payloads and webhook URLs are rendered with [LiquidJS](https://liquidjs.com/), so in addition to simple `{{ variable }}` substitution you can use Liquid tags and filters.
+
+### Conditionals
+
+Render a section only when the relevant object is present, falling back to `null` otherwise:
+
+```
+"media": {% if media %}{ "tmdbId": "{{ media_tmdbid }}" }{% else %}null{% endif %}
+```
+
+You can also branch on a value:
+
+```
+{% if media_status == "AVAILABLE" %}...{% else %}...{% endif %}
+```
+
+### Filters
+
+- **`json`** — `{{ subject | json }}`: serializes a value as JSON, quoting and escaping strings so characters like `"` or newlines can't break the payload. Circular-safe.
+- **`default`** — `{{ notifyuser_username | default: "Unknown User" }}`: falls back to a value when the input is nil or empty.
+- **`upcase` / `downcase`** — `{{ media_type | upcase }}`: changes case.
+- **`capitalize`** — `{{ notifyuser_username | capitalize }}`: capitalizes the first character.
+- **`truncate`** — `{{ subject | truncate: 50 }}`: shortens a string.
+- **`url_encode`** — `{{ subject | url_encode }}`: percent-encodes a value for use in a URL.
+
+See the [LiquidJS filter reference](https://liquidjs.com/filters/overview.html) for the full list.
+
+:::info
+Free-text values (titles, overviews, usernames, comments) should be emitted through the `json` filter so quotes and newlines can't produce invalid JSON, e.g. `"subject": {{ subject | json }}`.
+:::
+
+:::warning
+Webhook URL values are rendered raw. When substituting a value into a URL, use the `url_encode` filter (e.g. `...?user={{ notifyuser_username | url_encode }}`) so characters like `&`, `?`, or spaces don't break the URL.
+:::
+
+### Advanced object access
+
+In addition to the flat variables documented below, the full `media`, `request`, `issue`, and `comment` objects are available for conditionals and advanced templates (for example `{% if issue %}` or `{{ request.id }}`). These map to the [`NotificationPayload` interface](https://github.com/seerr-team/seerr/blob/develop/server/lib/notifications/agents/agent.ts).
 
 ## Template Variables
+
+:::warning Deprecated
+The flat template variables listed below (e.g. `{{ media_tmdbid }}`, `{{ notifyuser_username }}`) are deprecated and will be replaced in a future release. They continue to work for now, and a migration will be provided to convert existing templates automatically.
+:::
 
 ### General
 
@@ -77,21 +124,21 @@ On the other hand, the `notifyuser` variables _will_ be replaced with the reques
 If you would like to use the requesting user's information in your webhook, please instead include the relevant variables from the [Request](#request) section below.
 :::
 
-### Special
+### Objects
 
-The following variables must be used as a key in the JSON payload (e.g., `"{{extra}}": []`).
+The following values are the full objects for each notification. They are `null` when not relevant to the notification, and are primarily useful in conditionals (e.g. `{% if media %}`) or for accessing fields not covered by the flat variables below. Serialize an object into the payload with the `json` filter, e.g. `"extra": {{ extra | json }}`.
 
-| Variable      | Value                                                                                                                          |
-| ------------- | ------------------------------------------------------------------------------------------------------------------------------ |
-| `{{media}}`   | The relevant media object                                                                                                      |
-| `{{request}}` | The relevant request object                                                                                                    |
-| `{{issue}}`   | The relevant issue object                                                                                                      |
-| `{{comment}}` | The relevant issue comment object                                                                                              |
-| `{{extra}}`   | The "extra" array of additional data for certain notifications (e.g., season/episode numbers for series-related notifications) |
+| Variable    | Value                                                                                                                          |
+| ----------- | ------------------------------------------------------------------------------------------------------------------------------ |
+| `media`     | The relevant media object                                                                                                      |
+| `request`   | The relevant request object                                                                                                    |
+| `issue`     | The relevant issue object                                                                                                      |
+| `comment`   | The relevant issue comment object                                                                                              |
+| `extra`     | The "extra" array of additional data for certain notifications (e.g., season/episode numbers for series-related notifications) |
 
 #### Media
 
-The `{{media}}` will be `null` if there is no relevant media object for the notification.
+`media` will be `null` if there is no relevant media object for the notification.
 
 These following special variables are only included in media-related notifications, such as requests.
 
@@ -109,7 +156,7 @@ These following special variables are only included in media-related notificatio
 
 #### Request
 
-The `{{request}}` will be `null` if there is no relevant media object for the notification.
+`request` will be `null` if there is no relevant request object for the notification.
 
 The following special variables are only included in request-related notifications.
 
@@ -125,7 +172,7 @@ The following special variables are only included in request-related notificatio
 
 #### Issue
 
-The `{{issue}}` will be `null` if there is no relevant media object for the notification.
+`issue` will be `null` if there is no relevant issue object for the notification.
 
 The following special variables are only included in issue-related notifications.
 
@@ -140,7 +187,7 @@ The following special variables are only included in issue-related notifications
 
 #### Comment
 
-The `{{comment}}` will be `null` if there is no relevant media object for the notification.
+`comment` will be `null` if there is no relevant comment object for the notification.
 
 The following special variables are only included in issue comment-related notifications.
 
--- a/docs/using-seerr/notifications/webhook.md
+++ b/docs/using-seerr/notifications/webhook.md
@@ -35,7 +35,7 @@ You cannot configure both the **Authorization Header** field and a custom `Autho
 
 ### JSON Payload
 
-Customize the JSON payload to suit your needs. The payload is a [Liquid](https://liquidjs.com/) template that is rendered when a notification is triggered and then sent as the request body. Seerr exposes a set of [template variables](#template-variables), plus the full `media`, `request`, `issue`, and `comment` objects, for use in [conditionals and filters](#template-syntax).
+Customize the JSON payload to suit your needs. The payload is a [Liquid](https://liquidjs.com/) template that is rendered when a notification is triggered and then sent as the request body. Seerr exposes a set of [template variables](#template-variables), plus the full `media`, `request`, `issue`, and `comment` objects, for use in [conditionals and filters](#template-syntax). If you are new to Liquid, see the [introduction to Liquid tutorial](https://liquidjs.com/tutorials/intro-to-liquid.html).
 
 The template only needs to _render_ to valid JSON — it does not need to be valid JSON itself (for example, optional sections can be wrapped in `{% if %}` blocks). Use the **Test** button to confirm your template renders correctly; template errors are also written to the server logs.
 
--- a/server/lib/notifications/agents/webhook.ts
+++ b/server/lib/notifications/agents/webhook.ts
@@ -1,61 +1,11 @@
-import { IssueStatus, IssueType } from '@server/constants/issue';
-import { MediaStatus } from '@server/constants/media';
+import { TemplateEngine } from '@server/lib/notifications/templateEngine';
 import type { NotificationAgentWebhook } from '@server/lib/settings';
 import { getSettings } from '@server/lib/settings';
 import logger from '@server/logger';
 import axios from 'axios';
-import { get } from 'lodash';
 import { hasNotificationType, Notification } from '..';
 import type { NotificationAgent, NotificationPayload } from './agent';
 import { BaseAgent } from './agent';
-
-type KeyMapFunction = (
-  payload: NotificationPayload,
-  type: Notification
-) => string;
-
-const KeyMap: Record<string, string | KeyMapFunction> = {
-  notification_type: (_payload, type) => Notification[type],
-  event: 'event',
-  subject: 'subject',
-  message: 'message',
-  image: 'image',
-  notifyuser_username: 'notifyUser.displayName',
-  notifyuser_email: 'notifyUser.email',
-  notifyuser_avatar: 'notifyUser.avatar',
-  notifyuser_settings_discordId: 'notifyUser.settings.discordId',
-  notifyuser_settings_telegramChatId: 'notifyUser.settings.telegramChatId',
-  media_tmdbid: 'media.tmdbId',
-  media_tvdbid: 'media.tvdbId',
-  media_type: 'media.mediaType',
-  media_status: (payload) =>
-    payload.media ? MediaStatus[payload.media.status] : '',
-  media_status4k: (payload) =>
-    payload.media ? MediaStatus[payload.media.status4k] : '',
-  request_id: 'request.id',
-  requestedBy_username: 'request.requestedBy.displayName',
-  requestedBy_email: 'request.requestedBy.email',
-  requestedBy_avatar: 'request.requestedBy.avatar',
-  requestedBy_settings_discordId: 'request.requestedBy.settings.discordId',
-  requestedBy_settings_telegramChatId:
-    'request.requestedBy.settings.telegramChatId',
-  issue_id: 'issue.id',
-  issue_type: (payload) =>
-    payload.issue ? IssueType[payload.issue.issueType] : '',
-  issue_status: (payload) =>
-    payload.issue ? IssueStatus[payload.issue.status] : '',
-  reportedBy_username: 'issue.createdBy.displayName',
-  reportedBy_email: 'issue.createdBy.email',
-  reportedBy_avatar: 'issue.createdBy.avatar',
-  reportedBy_settings_discordId: 'issue.createdBy.settings.discordId',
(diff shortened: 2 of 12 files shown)


src/components/Settings/Notifications/NotificationsWebhook/index.tsx
```diff
@@ -134,11 +134,23 @@ const NotificationsWebhook = () => {
...
+          if (!value) return false;
+
+          // Check if the payload contains Handlebars template syntax
+          const hasHandlebarsContent = /\{\{[^}]+\}\}/.test(value);
+
+          if (hasHandlebarsContent) {
+            // For templates with Handlebars syntax, just check if it's not empty
+            // Backend will validate Handlebars syntax and provide error logs
+            return value.trim().length > 0;
+          }
```
mmenanno: More extensive validation would be if I imported Handlebars in the frontend to try test the render and see if we get errors but that would triple the bundle size for this page in my testing. If there are any other suggestions for simple frontend tests we should add, let me know

M0NsTeRRR:
@coderabbitai review

server/lib/notifications/agents/webhook.ts
```diff
@@ -71,81 +21,42 @@ class WebhookAgent
...
+    try {
+      // Parse the outer JSON string to get the template
+      const templateString = JSON.parse(payloadString);
+
+      // Render the template with Handlebars
+      let renderedString = TemplateEngine.render(templateString, payload, type);
+
+      // Clean up common JSON issues from template rendering
+      // Remove trailing commas before closing braces/brackets
+      renderedString = renderedString
```
M0NsTeRRR: I would let JSON.parse handle the parsing and avoid doing tricky things with the payload.

server/lib/notifications/agents/webhook.ts
```diff
@@ -71,81 +21,42 @@ class WebhookAgent
...
       this.getSettings().options.jsonPayload,
       'base64'
     ).toString('utf8');
 
-    const parsedJSON = JSON.parse(JSON.parse(payloadString));
-
-    return this.parseKeys(parsedJSON, payload, type);
+    try {
+      // Parse the outer JSON string to get the template
+      const templateString = JSON.parse(payloadString);
```
M0NsTeRRR: The render function only needs a string. The final goal is to produce a valid JSON. It is perfectly acceptable to have a non-valid JSON template with a payload that ultimately generates a valid JSON.

server/lib/notifications/agents/webhook.ts
```diff
@@ -71,81 +21,42 @@ class WebhookAgent
...
+        .replace(/([[{])\s*,/g, '$1');
+
+      // Parse the rendered template into an object
+      return JSON.parse(renderedString);
+    } catch (error) {
+      logger.error('Error rendering webhook payload template', {
+        label: 'Notifications',
+        errorMessage: error instanceof Error ? error.message : 'Unknown error',
+      });
+      // Fallback: try to parse as double-encoded JSON
```
M0NsTeRRR: I don't understand what is the goal of this fallback, do you have an example ?

server/lib/notifications/agents/webhook.ts
```diff
@@ -180,19 +91,18 @@ class WebhookAgent
     let webhookUrl = settings.options.webhookUrl;
 
     if (settings.options.supportVariables) {
```
M0NsTeRRR, line 65: Is supportVariables still needed @0xSysR3ll ?

server/lib/notifications/templateEngine.ts
```diff
@@ -0,0 +1,244 @@
...
+import Handlebars from 'handlebars';
+import { get } from 'lodash';
+import type { NotificationPayload } from './agents/agent';
+import { Notification } from './index';
+
+/**
+ * TemplateEngine provides Handlebars-based templating for notification agents.
+ * Supports variable substitution, conditionals, loops, and custom helpers.
+ */
+export class TemplateEngine {
```
M0NsTeRRR, line 14: Sorry to discover this only now (I don’t do much JS), but after reading this file, I searched for other template engines that support the helpers you implemented. I found Nunjucks. For our use case (an end-user template engine, so we need logic inside), it looks like a better fit. It would save us from having to implement every helper users might need.

https://mozilla.github.io/nunjucks/api.html

wdyt @seerr-team/seerr-core ?

server/lib/notifications/templateEngine.ts
```diff
@@ -0,0 +1,244 @@
...
+   */
+  static buildContext(
+    payload: NotificationPayload,
+    type: Notification
+  ): Record<string, unknown> {
+    // For auto-approved/auto-requested, use the requesting user as the notify user
+    const effectiveNotifyUser =
+      payload.notifyUser || payload.request?.requestedBy;
+
+    const context: Record<string, unknown> = {
```
M0NsTeRRR: Thanks for avoiding a breaking change we definitely want this! :)

src/components/Settings/Notifications/NotificationsWebhook/index.tsx
```diff
@@ -134,11 +134,23 @@ const NotificationsWebhook = () => {
         otherwise: Yup.string().nullable(),
       })
       .test(
-        'validate-json',
+        'validate-json-or-template',
         intl.formatMessage(messages.validationJsonPayloadRequired),
         (value) => {
+          if (!value) return false;
```
M0NsTeRRR: I think we can remove the entire test, since we can no longer verify it every payload will now be a template. Users can still use the test function to validate their templates.

M0NsTeRRR (commented):
Hello,
Sorry for the delay here’s some feedback before I test it! :)

server/lib/notifications/templateEngine.ts
gauthier-th: I'd prefer something like https://liquidjs.com/tutorials/intro-to-liquid.html
Nunjucks looks almost abandoned, and we only need some basic logic (which Liquid handles well)

mmenanno: Ok I'll see if I can get this spun back up in liquid instead as soon as I have a chance

server/lib/notifications/agents/webhook.ts
```diff
@@ -77,81 +21,18 @@ class WebhookAgent
...
   private buildPayload(type: Notification, payload: NotificationPayload) {
     const payloadString = Buffer.from(
       this.getSettings().options.jsonPayload,
       'base64'
     ).toString('utf8');
 
-    const parsedJSON = JSON.parse(JSON.parse(payloadString));
+    // The payload is stored as a JSON-encoded string; decode it to the raw
+    // template, render it, then parse the result as the final JSON body.
+    const template = JSON.parse(payloadString);
```
M0NsTeRRR: The render function only needs a string. The final goal is to produce a valid JSON. It is perfectly acceptable to have a non-valid JSON template with a payload that ultimately generates a valid JSON.

server/lib/notifications/templateEngine.ts
```diff
@@ -0,0 +1,160 @@
...
+        'comment.user.settings.discordIds',
+        ''
+      ),
+      commentedBy_settings_telegramChatId: get(
+        payload,
+        'comment.user.settings.telegramChatId',
+        ''
+      ),
+
+      // Full values, exposed for conditionals and advanced templates
```
M0NsTeRRR, line 137: It would be nice to include documentation for this in [the webhook notifications guide](https://github.com/seerr-team/seerr/blob/develop/docs/using-seerr/notifications/webhook.md), with a link to payload interfaces so users can see which fields are available.

Also, with the new template system, I don’t think we’ll keep the old variable system. We should mark it as deprecated now. In a separate PR, we’ll create a migration to convert existing notification templates to the new variable system.
wdyt @seerr-team/seerr-core ?

server/lib/notifications/agents/webhook.ts
```diff
@@ -186,19 +67,18 @@ class WebhookAgent
...
-            ? 'test'
-            : typeof keymapValue === 'function'
-              ? keymapValue(payload, type)
-              : get(payload, keymapValue) || 'test';
-        webhookUrl = webhookUrl.replace(
-          new RegExp(`{{${keymapKey}}}`, 'g'),
-          encodeURIComponent(variableValue)
+      try {
+        webhookUrl = TemplateEngine.render(webhookUrl, payload, type);
+      } catch (error) {
```
M0NsTeRRR, line 68: I don’t think we should move forward if there’s a problem. When would it be reasonable to continue?

server/lib/notifications/templateEngine.test.ts
```diff
@@ -0,0 +1,171 @@
...
+
+      const media = { tmdbId: 1 } as never;
+      const present = TemplateEngine.buildContext(
+        makePayload({ media }),
+        Notification.MEDIA_AVAILABLE
+      );
+      assert.equal(present.media, media);
+    });
+  });
+});
```
M0NsTeRRR, line 131: It would be good to add a test with undefined variables used in template.

server/lib/notifications/agents/webhook.ts
```diff
@@ -77,81 +21,18 @@ class WebhookAgent
...
     const payloadString = Buffer.from(
       this.getSettings().options.jsonPayload,
       'base64'
     ).toString('utf8');
 
-    const parsedJSON = JSON.parse(JSON.parse(payloadString));
+    // The payload is stored as a JSON-encoded string; decode it to the raw
+    // template, render it, then parse the result as the final JSON body.
+    const template = JSON.parse(payloadString);
+    const rendered = TemplateEngine.render(template, payload, type);
```
M0NsTeRRR: We should catch errors that occur inside the template (e.g., undefined variables) and log them clearly so users can diagnose the problem.

M0NsTeRRR (commented):
Hey,

An other review before testing out your PR let me know if something isn't clear or does not make sense.

Rombond:
Hey guys,

Any timeline for merging and releasing this PR?
Several other PRs were closed in favor of this one, and it's been open for a while. 
Which release version or milestone is this targeting? 3.4.0 ?

fallenbagel:
> Hey guys,
> 
> Any timeline for merging and releasing this PR?
> Several other PRs were closed in favor of this one, and it's been open for a while. 
> Which release version or milestone is this targeting? 3.4.0 ?

No ETA. If this is already reviewed, then it will be included in 3.4.0 and you can find it in the roadmap issue that links to the milestones.

mmenanno:
> Hey guys,
> 
> Any timeline for merging and releasing this PR? Several other PRs were closed in favor of this one, and it's been open for a while. Which release version or milestone is this targeting? 3.4.0 ?

This PR recently became active again, during the seerr transition this PR was waiting on a review for a bit. I'm currently working on the feedback I got on the PR yesterday and should have that up soon. So there is active effort on this currently to get it over the finish line

server/lib/notifications/agents/webhook.ts
mmenanno: I think this one's already covered. The `JSON.parse(payloadString)` here isn't validating the template as JSON, it's unwrapping the storage envelope: the save route stores the payload as `Buffer.from(JSON.stringify(templateString))`, so the decoded string is a quoted JSON string literal. `JSON.parse` there just returns the raw template string and always succeeds, whatever the template contains. The template is then rendered, and only the render output is parsed as JSON; so a non-JSON template that renders to valid JSON already works end-to-end.

server/lib/notifications/templateEngine.ts
mmenanno: On the docs: agreed, I'll add the Liquid syntax section + a link to the NotificationPayload interface.

On deprecating the flat variables now: I'd hold off. They're not just aliases; many do conversions the raw objects don't: enum→name (`media_status → "AVAILABLE"`, not `5`), fallbacks (`media_jellyfinMediaId`), renames (`notifyuser_username ← displayName`), and some aren't on any entity (`notification_type`). The nested objects are raw TypeORM entities, so switching couples templates to our DB schema and loses those conversions. I'd rather land a replacement that preserves them before marking the current ones deprecated - otherwise the follow-up migration has nothing to map to. wdyt?

server/lib/notifications/agents/webhook.ts
M0NsTeRRR: JSON.parse need a valid JSON string. You can test this by removing a quote inside a string on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse.

Anyway, what is the purpose of this `JSON.parse`? The `template.render` method expects a string as input (see https://github.com/seerr-team/seerr/pull/2039/changes#diff-423da8f9ae6ee96ec6c518029279fc89276a4b4642049021f358d7c6aaccf8baR152). You can directly pass `payloadString` to it there’s no need for this line.

server/lib/notifications/templateEngine.ts
M0NsTeRRR: For the last paragraph, it was mainly intended for other maintainers. But anyway, I was just asking to mark it as deprecated in the documentation nothing more. Another PR will handle this.

server/lib/notifications/agents/webhook.ts
mmenanno: You are right, context switching had my brain looking at this from the wrong angle 🙃

M0NsTeRRR (changes requested):

server/lib/settings/migrations/0009_migrate_webhook_payload_encoding.ts
```diff
@@ -0,0 +1,41 @@
+import type { AllSettings } from '@server/lib/settings';
```
M0NsTeRRR, line 1: 1. What is the purpose of this migration? Your AI seems to have hallucinated again. Please reread what was done [here](https://github.com/seerr-team/seerr/pull/2039/changes#diff-f1c662c873c720aa9876ac6a47795f0007de0ad8665656c0bac065c9ce8aad3cL184). Is this migration needed?
2. Please read `CONTRIBUTING.md` this is not the way to generate db migration file if you need it.

docs/using-seerr/notifications/webhook.md
```diff
@@ -35,10 +35,57 @@ You cannot configure both the **Authorization Header** field and a custom `Autho
 
 ### JSON Payload
 
-Customize the JSON payload to suit your needs. Seerr provides several [template variables](#template-variables) for use in the payload, which will be replaced with the relevant data when the notifications are triggered.
+Customize the JSON payload to suit your needs. The payload is a [Liquid](https://liquidjs.com/) template that is rendered when a notification is triggered and then sent as the request body. Seerr exposes a set of [template variables](#template-variables), plus the full `media`, `request`, `issue`, and `comment` objects, for use in [conditionals and filters](#template-syntax).
```
M0NsTeRRR: Add a link to their tutorial https://liquidjs.com/tutorials/intro-to-liquid.html

server/lib/notifications/templateEngine.test.ts
```diff
@@ -0,0 +1,183 @@
...
+  overrides: Partial<NotificationPayload> = {}
+): NotificationPayload =>
+  ({
+    subject: 'Test Subject',
+    notifySystem: true,
+    notifyAdmin: false,
+    ...overrides,
+  }) as unknown as NotificationPayload;
+
+describe('TemplateEngine', () => {
```
M0NsTeRRR, line 19: We shouldn’t test the LiquidJS engine that’s not our responsibility. We should assume it works. Instead, we should test, `Webhook.send` (with the appropriate mock). Our goal is to verify that:

- true is returned when the payload is properly rendered.
- false is returned if the payload fails to render.
- false is returned if the webhook URL is invalid.

server/lib/notifications/templateEngine.ts
```diff
@@ -0,0 +1,160 @@
...
+   * @throws if the template contains invalid Liquid syntax
+   */
+  static render(
+    template: string,
+    payload: NotificationPayload,
+    type: Notification
+  ): string {
+    const context = this.buildContext(payload, type);
+    // parseAndRenderSync is typed `any`; coerce to honor the `string` return.
+    return String(this.getEngine().parseAndRenderSync(template, context));
```
M0NsTeRRR: Why you do not use async function and update function signature `WebhookAgent.send` is aync ?

M0NsTeRRR (changes requested):

server/lib/settings/migrations/0009_migrate_webhook_payload_encoding.ts
mmenanno: The migration unwraps a storage-format change in the save route, not anything about the removed frontend validation: on `develop`, `POST /webhook` stores `Buffer.from(JSON.stringify(req.body.options.jsonPayload))`, while this PR stores the raw text. Without it, existing installs decode to a quoted JSON string literal and the agent would post a string body instead of the rendered object. 

On the second point this is a `settings.json` migration following the existing `server/lib/settings/migrations/0001`–`0008` pattern (auto-run by `migrator.ts`), not a TypeORM db migration, so the `CONTRIBUTING.md` `migration:generate` flow doesn’t apply. I’ve expanded the file’s header comment to make this explicit.

mmenanno:
Hey @M0NsTeRRR maybe you are just busy but thought I'd mention this is read for review in case the last notification didn't come through. I noticed the pr still has the `pending author's response` label but I think I've responded to everything

M0NsTeRRR:
Hey, yes, we’re currently focusing on bug fixes, but your PR will continue to be reviewed once we’re done with these. :)

PR #2042, opened 2025-10-14
ci(release.yml): update to release process 

sudo-kraken:
### Description
Release changelog: git-cliff generates release notes.

Sign: cosign signs the final image using GitHub OIDC.

SBOM and attestation: Trivy produces a CycloneDX SBOM for the final manifest digest. cosign attaches an SBOM attestation to the image.

Verify: cosign verifies the image signature and the SBOM attestation against the workflow identity. Then attaches to the release.

### DIFF
--- a/compose.postgres.yaml
+++ b/compose.postgres.yaml
@@ -23,7 +23,7 @@ services:
     links:
       - postgres
   postgres:
-    image: postgres
+    image: postgres:18
     environment:
       POSTGRES_USER: jellyseerr
       POSTGRES_PASSWORD: jellyseerr
@@ -31,6 +31,6 @@ services:
     ports:
       - '5432:5432'
     volumes:
-      - postgres:/var/lib/postgresql/data
+      - postgres:var/lib/postgresql/18/docker
 volumes:
   postgres:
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -104,7 +104,7 @@
     "swagger-ui-express": "4.6.2",
     "swr": "2.2.5",
     "tailwind-merge": "^2.6.0",
-    "typeorm": "0.3.12",
```
gauthier-th, line 107: Unwanted change

pnpm-lock.yaml
gauthier-th, line 1: Same thing here, there is no package update, so no reason to have a package lock update

.github/workflows/release.yml
```diff
@@ -172,37 +167,169 @@ jobs:
...
+        run: |
+          cosign sign --recursive "ghcr.io/${{ github.repository }}@${{ needs.publish.outputs.image_digest }}"
+          cosign sign --recursive "${{ github.repository }}@${{ needs.publish.outputs.image_digest }}"
+
+      - name: Generate SBOMs
+        run: |
+          trivy image --format cyclonedx --output seerr-ghcr-image-${{ env.VERSION }}.sbom \
+            "ghcr.io/${{ github.repository }}@${{ needs.publish.outputs.image_digest }}"
+
+          trivy image --format cyclonedx --output seerr-dockerhub-image-${{ env.VERSION }}.sbom \
```
M0NsTeRRR: The sbom should be identical ?

sudo-kraken: resolved in new commit

PR #2046, opened 2025-10-14
docs: postgres 18 documentation

M0NsTeRRR:
#### Description

Migrate existing compose file to postgres 18

### DIFF
--- a/compose.postgres.yaml
+++ b/compose.postgres.yaml
@@ -23,7 +23,7 @@ services:
     links:
       - postgres
   postgres:
-    image: postgres
+    image: postgres:18
     environment:
       POSTGRES_USER: jellyseerr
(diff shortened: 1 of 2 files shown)


gauthier-th (dismissed):

gauthier-th (approved):

Gylesie:
Great note, but this just defeats the purpose of the change that was made in the postgres image.

It should be mounted as `postgres:/var/lib/postgresql`. Also, I am pretty sure the the change removed the first / at the start there by mistake.

This way it makes migration to newer postgres versions easier.

fallenbagel:
> Also, I am pretty sure the the change removed the first / at the start there by mistake.
> 
Good catch!

PR #2049, opened 2025-10-15
fix: path in docs and compose for postgres 18

sudo-kraken:
#### Description
fix: path in docs and compose for postgres 18

### DIFF
--- a/compose.postgres.yaml
+++ b/compose.postgres.yaml
@@ -31,6 +31,6 @@ services:
     ports:
       - '5432:5432'
     volumes:
-      - postgres:var/lib/postgresql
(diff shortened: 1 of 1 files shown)


compose.postgres.yaml
```diff
@@ -31,6 +31,6 @@ services:
     ports:
       - '5432:5432'
     volumes:
-      - postgres:var/lib/postgresql/18/docker
+      - postgres:var/lib/postgresql
```
fallenbagel:
```suggestion
      - postgres:/var/lib/postgresql
```

fallenbagel (approved):

0xSysR3ll (approved):

PR #2066, opened 2025-10-18
fix: preserve deleted status when processing movies

OwsleyJr:
#### Description

Small PR that fixes a case where the DELETED status was being unintentionally overwritten for movies. Also added download progress if a movie is processing after deleted:

- When the media is marked as deleted but still exists in radarr, the status was changing back to requested.
- Download progress added to the deleted status badge.

Screenshot 2025-10-18 at 12 00 44 AM

#### To-Dos

### DIFF
--- a/src/components/StatusBadge/index.tsx
+++ b/src/components/StatusBadge/index.tsx
@@ -412,14 +412,27 @@ const StatusBadge = ({
               </span>
               {inProgress && (
                 <>
-                  {mediaType === 'tv' && downloadItem[0].episode && (
-                    <span className="ml-1">
-                      {intl.formatMessage(messages.seasonepisodenumber, {
-                        seasonNumber: downloadItem[0].episode.seasonNumber,
-                        episodeNumber: downloadItem[0].episode.episodeNumber,
-                      })}
-                    </span>
-                  )}
+                  {mediaType === 'tv' &&
+                    downloadItem[0].episode &&
+                    (downloadItem.length > 1 &&
+                    downloadItem.every(
+                      (item) =>
+                        item.downloadId &&
+                        item.downloadId === downloadItem[0].downloadId
+                    ) ? (
+                      <span className="ml-1">
+                        {intl.formatMessage(messages.seasonnumber, {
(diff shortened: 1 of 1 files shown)


server/lib/scanners/baseScanner.ts
gauthier-th, line 1: I think you should also edit the Jellyfin scanner to add this behavior.
This is something that needs refactor since a while, but the `JellyfinScanner` class is currently not inheriting from the `BaseScanner` class.

OwsleyJr: Ah okay I'll take a look when I have a chance

fallenbagel: > I think you should also edit the Jellyfin scanner to add this behavior.  This is something that needs refactor since a while, but the `JellyfinScanner` class is currently not inheriting from the `BaseScanner` class.

No need to edit the jellyfin scanner as #2226 refactors `JellyfinScanner` to inhering from the `BaseScanner` class

gauthier-th (approved):

server/lib/scanners/baseScanner.ts
```diff
@@ -115,9 +115,11 @@ class BaseScanner<T> {
...
 
         if (existing[is4k ? 'status4k' : 'status'] !== MediaStatus.AVAILABLE) {
-          existing[is4k ? 'status4k' : 'status'] = processing
-            ? MediaStatus.PROCESSING
-            : MediaStatus.AVAILABLE;
+          existing[is4k ? 'status4k' : 'status'] = !processing
+            ? MediaStatus.AVAILABLE
+            : existing[is4k ? 'status4k' : 'status'] === MediaStatus.DELETED
+            ? MediaStatus.DELETED
+            : MediaStatus.PROCESSING;
```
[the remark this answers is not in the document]
fallenbagel: Bit of nit/optional: 
It could be slightly clearer by checking the DELETED case first, or in my opinion, using an if else block would make it the best (readability wise). But this works fine as is too

PR #2069, opened 2025-10-18
docs: add migration guide

M0NsTeRRR:
#### Description

add migration guide

### DIFF
--- a/docs/migration-guide.mdx
+++ b/docs/migration-guide.mdx
@@ -31,10 +31,7 @@ If you're migrating from a previous installation, you may need to update the own
 sudo chown -R 1000:1000 /path/to/appdata/config
 ```
 
-Alternatively, you can use `chmod` to grant permissions:
-```bash
-sudo chmod -R 755 /path/to/appdata/config
-```
+This ensures the `node` user (UID 1000) owns the config directory and can read and write to it.
 :::
 
 ### Unix


docs/migration-guide.mdx
```diff
@@ -0,0 +1,34 @@
...
+Then, restart all services defined in the Compose file:
+```bash
+docker compose up -d
+```
+
+## Kubernetes
+Refer to [Seerr Kubernetes Documentation](/getting-started/kubernetes), all of our example have been updated to reflect the below change.
+
+As our container image can be run without root permission we have updated the default `securityContext` and `podSecurityContext` to match this.
+Every reference to jellyseerr have been renamed to seerr in manifests.
```
sudo-kraken:
```suggestion
Every reference to jellyseerr has been renamed to seerr in manifests.
```

sudo-kraken: Only small change in this PR, once corrected I will approve

docs/migration-guide.mdx
```diff
@@ -0,0 +1,34 @@
...
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+This documentation is for users who want to migrate their existing Overseerr or Jellyseerr instance to Seerr.
+Jellyseerr serves as the codebase for Seerr, but Overseerr users do not need to perform any manual migration steps.
+When you start your Seerr instance, a script automatically detects if your instance is from Overseerr and handles the database migration.
+
+:::warning
+Before doing anything you should backup your existing instance so that you can rollback in case something goes wrong.
+See [Configuring Databases](/using-jellyseerr/backups) for details on how to properly backup your instance.
```
gauthier-th:
```suggestion
See [Backups](/using-jellyseerr/backups) for details on how to properly backup your instance.
```

docs/migration-guide.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+## Docker
+Refer to [Seerr Docker Documentation](/getting-started/docker), all of our example have been updated to reflect the below change.
+
+Down your compose stack
+```bash
+docker compose down
+```
+Copy paste the new compose file (don't forget to update the volume path in the new compose file).
+Then, restart all services defined in the Compose file:
```
gauthier-th: To make things easier for users, we could say here that they just need to change the image to `seerr/seerr` and also give the new volume path?

docs/migration-guide.mdx
```diff
@@ -0,0 +1,34 @@
+---
+title: Migration guide
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+This documentation is for users who want to migrate their existing Overseerr or Jellyseerr instance to Seerr.
+Jellyseerr serves as the codebase for Seerr, but Overseerr users do not need to perform any manual migration steps.
+When you start your Seerr instance, a script automatically detects if your instance is from Overseerr and handles the database migration.
```
gauthier-th: I find this a bit difficult to understand when I put myself in the shoes of a user with limited knowledge.
Maybe say something like this instead?
> Whether you come from Overseerr or Jellyseerr, you don't need to perform any manual migration steps, your instance will automatically be migrated to Seerr. An additional migration will happen for Overseerr users, to migrate their configuration to the new codebase.

docs/migration-guide.mdx
M0NsTeRRR: They also need to add `init: true`. I think it’s easier for users with less experience to copy and paste the new file and update the `volumePath` (one step) rather than manually adding keys to the YAML (multiple steps). What do you think?

docs/migration-guide.mdx
M0NsTeRRR: I don’t find it difficult, and it provides users with more information about what will happen. However, for less experienced users, it still states: 'Users do not need to perform any manual migration steps.'
I can add your sentence if you think it makes things clearer I don’t mind.

Gylesie: I aggree with @gauthier-th , 

> Jellyseerr serves as the codebase for Seerr, but Overseerr users do not need to perform any manual migration steps.

Was confusing for me too, especially the latter part felt out of place.

Why do the users need to know about this? Its technical and there shouldnt be any input from the users. Migrations are transparent to the users.

M0NsTeRRR: I’ll update it accordingly, then.

> Why do the users need to know about this? Its technical and there shouldnt be any input from the users. Migrations are transparent to the users.

Because it’s their data, and I believe transparency in how we handle it is at the core of open source

Gylesie:
This guide is low value for users with existing Jellyseerr and Overserr instances. It essentially tells them to scratch whatever they had previously in e.g. docker compose and replace it with whatever is in the general getting started guide which is geared towards new users.

M0NsTeRRR:
> This guide is low value for users with existing Jellyseerr and Overserr instances. It essentially tells them to scratch whatever they had previously in e.g. docker compose and replace it with whatever is in the general getting started guide which is geared towards new users.

1) It’s still in draft, I need to confirm whether additional migration steps are required.
2) This is for users with less experience, and it’s meant to prevent support tickets like, 'Is it really as easy as updating the Docker image reference?' This way, we can simply redirect them here. Users with more experience already know how to handle the new image and what it does for them.

docs/migration-guide.mdx
Gylesie: Mentioning only the necessary changes is better in my opinion. The user might have added additional volume mounts, env. variables, labels, custom container name, etc.

This just makes it annoying and would leave the user to figure it out for themselves from the getting started docs linked.

Gylesie:
I understand its in draft, but you requested a review. So thats why I wanted to give some feedback.

I just dont find it helpful overall. It literally says to replace the whole compose file with new one. But there is a good chance that the user is running different services in a single compose file. What should they do then?

I dont think majority of users would need this much hand holding to migrate. Showing an example of an old docker compose file and then the new one with specific changes that were made highlighted would be better.

docs/migration-guide.mdx
gauthier-th: Also agree for saying what changes needs to be made (but sure we should also say "refer to the new one here").
A diff with the old compose and the new one could also help.

docs/migration-guide.mdx
```diff
@@ -0,0 +1,156 @@
...
+:::warning
+Before doing anything you should backup your existing instance so that you can rollback in case something goes wrong.
+See [Backups](/using-jellyseerr/backups) for details on how to properly backup your instance.
+:::
+
+## Docker
+Refer to [Seerr Docker Documentation](/getting-started/docker), all of our example have been updated to reflect the below change.
+
+Changes :
+- Rename all references from `overseerr` or `jellyseerr` to `seerr`.
```
0xSysR3ll: Rename**d** ?

0xSysR3ll (changes requested):

docs/migration-guide.mdx
```diff
@@ -0,0 +1,156 @@
+---
+title: Migration guide
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+Whether you come from Overseerr or Jellyseerr, you don't need to perform any manual migration steps, your instance will automatically be migrated to Seerr.
```
gshpychka, line 8: Might be worth clarifying that this will happen when you first start Seerr (assuming that's what this is saying), so you do need to do that much (switch your package/image/etc to seerr).

docs/migration-guide.mdx
```diff
@@ -0,0 +1,156 @@
...
+An additional migration will happen for Overseerr users, to migrate their configuration to the new codebase.
+
+:::warning
+Before doing anything you should backup your existing instance so that you can rollback in case something goes wrong.
+See [Backups](/using-jellyseerr/backups) for details on how to properly backup your instance.
```
gauthier-th:
```suggestion
See [Backups](/using-seerr/backups) for details on how to properly backup your instance.
```

sudo-kraken (approved):
LGTM

docs/migration-guide.mdx
```diff
@@ -0,0 +1,157 @@
...
+This migration will run automatically the first time you start your instance using the Seerr codebase (Docker image or source build or Kubernetes, etc.).
+An additional migration will happen for Overseerr users, to migrate their configuration to the new codebase.
+
+:::warning
+Before doing anything you should backup your existing instance so that you can rollback in case something goes wrong.
+See [Backups](/using-seerr/backups) for details on how to properly backup your instance.
+:::
+
+## Docker
+Refer to [Seerr Docker Documentation](/getting-started/docker), all of our example have been updated to reflect the below change.
```
Gauvino:
```suggestion
Refer to [Seerr Docker Documentation](/getting-started/docker), all of our examples have been updated to reflect the below change.
```

docs/migration-guide.mdx
```diff
@@ -0,0 +1,157 @@
...
+    -p 5055:5055 \
+    -v seerr-data:/app/config \
+    --restart unless-stopped \
+    ghcr.io/seerr-team/seerr:latest
+  ```
+  </TabItem>
+</Tabs>
+
+## Kubernetes
+Refer to [Seerr Kubernetes Documentation](/getting-started/kubernetes), all of our example have been updated to reflect the below change.
```
Gauvino:
```suggestion
Refer to [Seerr Kubernetes Documentation](/getting-started/kubernetes), all of our examples have been updated to reflect the below change.
```

Gauvino (commented):
Only two minor typo I see otherwise seem's very nice

0xSysR3ll:
Should we also write some documentation about the new rights on the `config` folder ?

0xSysR3ll:
> Should we also write some documentation about the new rights on the `config` folder ?

Done. LGTM now !

PR #2092, opened 2025-10-27
feat: use tmdb with anidb on jellyfin

ale183:
#### Description
This PR changes the metadata provider to TMDB when using anidb for better season recognition

#### To-Dos

- [X] Disclosed any use of AI (see our [policy](../CONTRIBUTING.md#ai-assistance-notice))

### DIFF
--- a/server/api/animelist.ts
+++ b/server/api/animelist.ts
@@ -48,11 +48,11 @@ interface AnimeList {
 
 export interface AnidbItem {
   tvdbId?: number;
-  tmdbtv?: number;
+  tmdbTv?: number;
   tmdbId?: number;
   imdbId?: string;
   tvdbSeason?: number;
-  tmdbseason?: number;
+  tmdbSeason?: number;
 }
 
 class AnimeListMapping {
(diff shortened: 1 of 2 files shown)


server/api/animelist.ts
```diff
@@ -46,9 +48,11 @@ interface AnimeList {
 
 export interface AnidbItem {
   tvdbId?: number;
+  tmdbtv?: number;
```
gauthier-th: You should call this variable `tmdbTv` and not `tmdbtv`.

server/api/animelist.ts
```diff
@@ -46,9 +48,11 @@ interface AnimeList {
...
   tmdbId?: number;
   imdbId?: string;
   tvdbSeason?: number;
+  tmdbseason?: number;
```
gauthier-th: Same here, call this variable `tmdbSeason` and not `tmdbseason`.

gauthier-th (changes requested):
Could you please explain a bit further why this change from TVDB to TMDB helps?
And shouldn't this be done depending on the current metadata provider (we now support TMDB and TVDB)?

ale183:
I kinda wrote the wrong thing in the description, the provider remains the same:
Before the change the provider was TMDB (more precisely TMDB using TVDB id) and the reference season from `animelist.ts` was `tvdbSeason`, this could generate some mismatches.
After this change the provider is still TMDB (but it uses the TMDB id from `animelist.ts`) and it references the correct season (`tmdbSeason` from `animelist.ts`).

I could use the method `getMetadataProvider`, but the variable `tvShow` in `processShow` has the type `TmdbTvDetails | null`, i think that this change should be done in another pr, but correct me if i'm wrong.

gauthier-th:
@ale183 sorry for the late review. Can you please address / comment the comments above?
Can you also rebase this on develop and you edit your PR description to match the new PR template ([available here](https://github.com/seerr-team/seerr/blob/develop/.github/PULL_REQUEST_TEMPLATE.md)).

PR #2140, opened 2025-11-11
fix(webpush): improve iOS push subscription endpoint cleanup

0xSysR3ll:
#### Description

This series fixes several recurring WebPush issues: transient vs permanent delivery failures, iOS key/endpoint rotation, duplicate/stale subscriptions, race conditions during registration, and frontend state desynchronisation that forced users to re-enable push repeatedly.

##### What has been done here ?

-  As per [RFC 8030](https://datatracker.ietf.org/doc/html/rfc8030), only permanent failures (e.g `HTTP 410 Gone` and `404 Not Found`) cause subscription deletion. So transient errors like rate limiting, `5xx`, network blips, `429`, etc. are logged but the subscription is retained - preventing unnecessary loss of subscriptions when the push service is temporarily unhappy.

- Handle iOS/Safari key rotation (`auth`/`p256dh`) without dropping subscriptions - if the backend receives the same endpoint but a changed `auth`/`p256dh` (common on iOS), we update the stored keys (and `userAgent`) instead of creating duplicates or deleting the existing record. This avoids forcing users to re-subscribe after valid key refreshes.

- Stale-subscription cleanup for the same device (`userAgent`) is limited to entries with a different endpoint (`endpoint != req.body.endpoint`) so active subscriptions are not removed accidentally (and aggressively).

- Subscription registration and stale cleanup are wrapped in a DB transaction; we re-check existence inside the transaction to avoid duplicate inserts when concurrent requests occur.

- `unsubscribeToPushNotifications` now returns the current endpoint when available; frontend uses that to ensure we delete the correct backend subscription.

- `verifyPushSubscription` and `localStorage` handling were also improved to avoid spurious toggles.

#### To-Dos

- [ ] Disclosed any use of AI (see our [policy](../CONTRIBUTING.md#ai-assistance-notice))

#### Issues Fixed or Closed

- Fixes #XXXX

### DIFF
--- a/src/components/UserProfile/UserSettings/UserNotificationSettings/UserNotificationsWebPush/index.tsx
+++ b/src/components/UserProfile/UserSettings/UserNotificationSettings/UserNotificationsWebPush/index.tsx
@@ -185,7 +185,7 @@ const UserWebPushSettings = () => {
           (device) => device.userAgent === currentUserAgent
         );
 
-        if (hasMatchingDevice || dataDevices.length === 1) {
+        if (hasMatchingDevice) {
           isEnabled = true;
         }
       }


gauthier-th:
@OwsleyJr

fallenbagel (approved):

0xSysR3ll:
Sorry, just had to fix 2 small UI things

fallenbagel:
> Sorry, just had to fix 2 small UI things

Is this PR now ready for review?

0xSysR3ll:
> > Sorry, just had to fix 2 small UI things
> 
> 
> 
> Is this PR now ready for review?

Yes!

gauthier-th (approved):

PR #2146, opened 2025-11-12
fix(docker): pass `COMMIT_TAG` to build stage for custom image builds

0xSysR3ll:
#### Description

This PR fixes the issue where custom images built with `--build-arg COMMIT_TAG` would fail because the client bundle didn't receive the commit tag value.

#### To-Dos

~~- [ ] Disclosed any use of AI (see our [policy](../CONTRIBUTING.md#ai-assistance-notice))~~
~~- [ ] Successful build `pnpm build`~~
~~- [ ] Translation keys `pnpm i18n:extract`~~
~~- [ ] Database migration (if required)~~

#### Issues Fixed or Closed

- Fixes #XXXX

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,9 @@ RUN --mount=type=cache,id=pnpm,target=/pnpm/store CI=true pnpm install --prod --
 
 FROM base AS build
 
+ARG COMMIT_TAG
+ENV COMMIT_TAG=${COMMIT_TAG}
+
 RUN \
   case "${TARGETPLATFORM}" in \
   'linux/arm64' | 'linux/arm/v7') \


fallenbagel (approved):

M0NsTeRRR:
As discussed with @0xSysR3ll,

Our current system need a bit of rework (in an other PR) :

- The commit tag should not be mandatory; we could use a default value (e.g `dev`) or generate a random one if needed (e.g `dev-<UUID>` or `dev-<DATE>` depend on what we want to achieve) .
- Instead of relying on both an environment variable and a JSON file, we should streamline this by using only environment variables, with the option to load them from an `.env` file.

0xSysR3ll:
> As discussed with @0xSysR3ll,
> 
> Our current system need a bit of rework (in an other PR) :
> 
> * The commit tag should not be mandatory; we could use a default value (e.g `dev`) or generate a random one if needed (e.g `dev-<UUID>` or `dev-<DATE>` depend on what we want to achieve) .
> * Instead of relying on both an environment variable and a JSON file, we should streamline this by using only environment variables, with the option to load them from an `.env` file.

Needless to say, this also affects CI builds, not only custom builds.

PR #2150, opened 2025-11-13
feat(api): load API key through systemd credentials with env fallback

blackxored:
#### Description

On Linux systems, be a well-behaved daemon and prefer systemd credential `api-key` if passed to the service. This is done through a file, access is checked by the kernel, and there's no process inheritance, TL;DR is more secure than env vars. See: [systemd Credentials](https://systemd.io/CREDENTIALS/). 

It would still fallback to the `API_KEY` env var as there's no need for this to be a breaking change.

#### To-Dos

- [ ] Disclosed any use of AI (see our [policy](../CONTRIBUTING.md#ai-assistance-notice))

### DIFF
--- a/docs/using-seerr/settings/general.md
+++ b/docs/using-seerr/settings/general.md
@@ -12,7 +12,18 @@ This is your Seerr API key, which can be used to integrate Seerr with third-part
 
 If you need to generate a new API key for any reason, simply click the button to the right of the text box.
 
-If you want to set the API key, rather than letting it be randomly generated, you can use the API_KEY environment variable. Whatever that variable is set to will be your API key.
+If you want to set the API key, rather than letting it be randomly generated,
+you can use the API_KEY environment variable. Whatever that variable is set to
+will be your API key.
+
+You can also use the `api-key` credential, if you're using a systemd service.
+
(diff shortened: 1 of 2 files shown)


M0NsTeRRR:
Hi @blackxored,

Instead of creating a new `apiKeyFromEnvOrCred` function, could you move the `stringOrReadFileFromEnv` function from `server/datasource.ts` to `server/utils/env.ts`? This way, we can reuse the same logic where needed.

Additionally, since this feature is similar to Docker secrets, it would be great to support both approaches simultaneously. By default, Docker secrets are mounted at `/run/secrets`. I think adding `ENV CREDENTIALS_DIRECTORY="/run/secrets"` to the Dockerfile would be enough to make them work out of the box. What do you think?

[Reference: Docker Secrets Documentation](https://docs.docker.com/engine/swarm/secrets/)

benja1006:
I agree, it would work nicely to place the systemd credentials into /run/secrets, then we can access them using _FILE appended to the DB username, password and name. This would integrate cleanly into the default postgres behavior as well

M0NsTeRRR:
Any news @blackxored ?

PR #2158, opened 2025-11-19
fix: most number types should be integer types in seerr-api.yml, and 48 additional commits; 49 commits to seerr-api.yml

daball:
#### Description

NSwag generates double types when number is provided. Seerr service returns 400 Bad Request when decimal numbers are sent from HttpClient. This fix changes most number types to integer types in seerr-api.yml, which can be used to generate NSwag API clients which after the change generate int types versus double types in C#.

N/A: Not UI related.

#### To-Dos

- [ ] I mostly used the find and replace tool in VS Code, some of it was edited by hand, some of it was generated in VS Code using AI auto-completion after it noticed what I was doing. I did review each line carefully before committing to my branch.
- [ ] Successfully built using `pnpm build` via Podman Compose on Windows 11. I was able to run, compile the pages, configure the Seerr service, connect to Jellyfin service, sync libraries, and perform all normal tasks following this commit.
- [ ] N/A: Translation keys `pnpm i18n:extract`
- [ ] N/A: Database migration (if required)

#### Issues Fixed or Closed

N/A: I didn't see any open issues.

### DIFF
--- a/seerr-api.yml
+++ b/seerr-api.yml
@@ -1896,6 +1896,47 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/MediaRequest'
+    SeasonsAll:
+      type: string
+      enum:
+        - all
+    PostRequest:
+      type: object
+      properties:
+        mediaType:
+          type: string
+          enum: [movie, tv]
+          example: movie
+        mediaId:
+          type: integer
+          example: 123
+        tvdbId:
+          type: integer
+          example: 123
+        seasons:
+          oneOf:
+            - type: array
+              items:
+                type: integer
+                minimum: 0
+            - $ref: '#/components/schemas/SeasonsAll'
+        is4k:
+          type: boolean
+          example: false
+        serverId:
+          type: integer
+        profileId:
+          type: integer
+        rootFolder:
+          type: string
+        languageProfileId:
+          type: integer
+        userId:
+          type: integer
+          nullable: true
+      required:
+        - mediaType
+        - mediaId
     RequestsCounts:
       type: object
       properties:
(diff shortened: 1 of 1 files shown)


0xSysR3ll:
Would be preferable to make another separate PR for the schemas declarations - will be easier to review :)

daball:
bug: paths: /settings/cache: get: responses: '200': dnsCache: properties: stats: properties: hitRate should be 'number' and not 'integer', rates such as hitRate are fractional

daball:
NSwag is no longer generating Anonymous#, Response#, and Data# classes. There were many opportunities to consolidate the number of classes by reusing similar classes and to destructure complex object schemas into flatter schemas. And many good candidates for integer types have been converted from number types (and in one case changed back).

I fixed a few unidentified bugs, particularly in commits:

+ 839d6586e7a6ef110cd9d3462e23667908c8f132 fix: most number types should be integer types in seerr-api.yml
+ 5b46f1db5a4391e957d8f2a5205aa84d031580c6 fix: fix semantic error at paths./settings/cache.get.responses.200.content.application/json.schema.properties.dnsCache.properties.entries, schemas with 'type: array', require a sibling 'items: ' field, resolved with refactor, introduced items: field and changed additionalProperties to properties
+ 5b46f1db5a4391e957d8f2a5205aa84d031580c6 fix: hitRate is a number, not an integer (fixing my own mistake, the one I changed back to number)
+ b43241fc2f009621f62995908872ba17c3decaa0 fix: change paths: /settings/network: get: responses: '200': content: application/json: schema: $ref: should be '#/components/schemas/NetworkSettings' instead of '#/components/schemas/MainSettings'
+ 5df773c0e8d068c7d59b354df39d064c6acd392d fix: WatchProviders named schema should be WatchProvider and should not be an array based on the usage from MovieDetails and the TvDetails named schemas, which are already declaring the array in each schema
+ c39c8da067f3a5ea28fc48053307206ccbf64de1 fix: /user/{userId}/pushSubscriptions endpoint returns an array of PushSubscription
+ 9c643dcbad79393ed79e48c6979bf20459a27b92 fix: added missing mediaType field to Watchlist
+ 1d9f50d72b240d3f63818b9433bb47491cec391d fix: added missing pageSize to PageInfo named schema
+ fc90017bb8a339c4b973f160b3350dcf18f0e161 fix: /tv/{tvId}/ratings endpoint missing audienceRating and audienceScore returned by service

I edited documentation in two places, particularly in commits:

- a018caf8cd04fb9e631c38e9b59cbb82f45678fa docs: add example /auth/jellyfin to seerr-api.yml description
- 1d9f50d72b240d3f63818b9433bb47491cec391d docs: added descriptions to fields in PageInfo named schema

And everything else is refactoring unnamed schemas to named schemas with no change in function.

Perhaps at some point I will look at adding description fields to everything, but that's another branch and PR for another time. This one is ready for review.

daball:
I had to make two more small changes. Collection was conflicting with .NET's Collection in generated source code from OpenApi Generator. I've now tested that the generated source code from both NSwag and OpenApi Generator source code builds. There was one line of generated source code that I had to fix pertaining to the example near:

```
paths:
  /settings/jobs/{jobId}/schedule:
...
                schedule:
                  type: string
                  example: '0 */5 * * * *'

```

This seems to confuse the OpenApi Generator logic and causes it to emit /* <example>0 */5 * * * *</example> */ instead of /// <example>0 */5 * * * *</example>, which I had to manually fix at each iteration. I consider this to be a problem with OpenApi Generator in this case.

PR #2164, opened 2025-11-21
feat(overriderules): apply override rules to advanced requests

### LINKED ISSUE #1506: Media requested with the advanced requests permision should have Override Rules applied
### Description

When requesting media with the advanced requests permision the root folder should be set to the folder set by the Override Rules

Selected:
Image

Should Be Selected:
Image

### Version

2.5.1

### Steps to Reproduce

1. Create an override rule that changes the root folder
2. Request a piece of media where the rule would normally apply without Advanced Requests
3. Notice that the Root Folder is set to the default instead of the one chosen by the override rule
4. You will now have to manually select your root folder instead of it being done by the Override Rule which is undesirable

### Screenshots

Image

Image

### Logs

```shell

```

### Platform

desktop

### Database

SQLite (default)

### Device

Computer

### Operating System

Windows 10

### Browser

Chrome

### Additional Context

_No response_

### Code of Conduct

gauthier-th:
## Description

This PR apply the override rules to the Advanced Request Modal.
Also moves the Override Rules logic to into its own file in `lib/` for more clarity.

- Fixes #1506

## How Has This Been Tested?

1. Setup an override rule
2. Open the request modal with advanced perms and check that the rule is applied in the modal.

## Checklist:

- [x] Disclosed any use of AI (see our [policy](https://github.com/seerr-team/seerr/blob/develop/CONTRIBUTING.md#ai-assistance-notice)): none

## New Features

- Advanced request forms now load matching override rules for movies and TV shows.
- Suggested root folders, quality profiles, and tags are applied automatically.
- Added API support for advanced override requests, including service and requester selection.
- Rules can be evaluated for different request users when permitted.

## Bug Fixes

- Override matching now better reflects media, user, language, genre, keyword, and service criteria.
- Matching rules are prioritized and merged more reliably.
- Advanced request settings update accurately for the selected title and server.

### DIFF
--- a/server/lib/overrideRules.ts
+++ b/server/lib/overrideRules.ts
@@ -1,0 +1,153 @@
+import { ANIME_KEYWORD_ID } from '@server/api/themoviedb/constants';
+import type {
+  TmdbKeyword,
+  TmdbMovieDetails,
+  TmdbTvDetails,
+} from '@server/api/themoviedb/interfaces';
+import { MediaType } from '@server/constants/media';
+import { getRepository } from '@server/datasource';
+import OverrideRule from '@server/entity/OverrideRule';
+import type { User } from '@server/entity/User';
+import { getSettings } from '@server/lib/settings';
+import logger from '@server/logger';
+
+export type OverrideRulesResult = {
+  rootFolder: string | null;
+  profileId: number | null;
+  tags: number[] | null;
+};
+
+async function overrideRules({
+  mediaType,
+  is4k,
+  tmdbMedia,
+  requestUser,
+}: {
+  mediaType: MediaType;
+  is4k: boolean;
+  tmdbMedia: TmdbMovieDetails | TmdbTvDetails;
+  requestUser: User;
+}): Promise<OverrideRulesResult> {
+  const settings = getSettings();
+
+  let rootFolder: string | null = null;
+  let profileId: number | null = null;
+  let tags: number[] | null = null;
+
+  const defaultRadarrId = is4k
+    ? settings.radarr.findIndex((r) => r.is4k && r.isDefault)
+    : settings.radarr.findIndex((r) => !r.is4k && r.isDefault);
+  const defaultSonarrId = is4k
+    ? settings.sonarr.findIndex((s) => s.is4k && s.isDefault)
+    : settings.sonarr.findIndex((s) => !s.is4k && s.isDefault);
+
+  const overrideRuleRepository = getRepository(OverrideRule);
+  const overrideRules = await overrideRuleRepository.find({
+    where:
+      mediaType === MediaType.MOVIE
+        ? { radarrServiceId: defaultRadarrId }
+        : { sonarrServiceId: defaultSonarrId },
+  });
+
+  const appliedOverrideRules = overrideRules.filter((rule) => {
+    const hasAnimeKeyword =
+      'results' in tmdbMedia.keywords &&
+      tmdbMedia.keywords.results.some(
+        (keyword: TmdbKeyword) => keyword.id === ANIME_KEYWORD_ID
+      );
+
+    // Skip override rules if the media is an anime TV show as anime TV
+    // is handled by default and override rules do not explicitly include
+    // the anime keyword
+    if (
+      mediaType === MediaType.TV &&
+      hasAnimeKeyword &&
+      (!rule.keywords ||
+        !rule.keywords.split(',').map(Number).includes(ANIME_KEYWORD_ID))
+    ) {
+      return false;
+    }
+
+    if (
+      rule.users &&
+      !rule.users.split(',').some((userId) => Number(userId) === requestUser.id)
+    ) {
+      return false;
+    }
+    if (
+      rule.genre &&
+      !rule.genre
+        .split(',')
+        .some((genreId) =>
+          tmdbMedia.genres.map((genre) => genre.id).includes(Number(genreId))
+        )
+    ) {
+      return false;
+    }
+    if (
+      rule.language &&
+      !rule.language
+        .split('|')
+        .some((languageId) => languageId === tmdbMedia.original_language)
+    ) {
+      return false;
+    }
+    if (
+      rule.keywords &&
+      !rule.keywords.split(',').some((keywordId) => {
+        let keywordList: TmdbKeyword[] = [];
+
+        if ('keywords' in tmdbMedia.keywords) {
+          keywordList = tmdbMedia.keywords.keywords;
+        } else if ('results' in tmdbMedia.keywords) {
+          keywordList = tmdbMedia.keywords.results;
+        }
+
+        return keywordList
+          .map((keyword: TmdbKeyword) => keyword.id)
+          .includes(Number(keywordId));
+      })
+    ) {
+      return false;
+    }
+    return true;
+  });
+
+  // hacky way to prioritize rules
+  // TODO: make this better
+  const prioritizedRule = appliedOverrideRules.sort((a, b) => {
+    const keys: (keyof OverrideRule)[] = ['genre', 'language', 'keywords'];
+
+    const aSpecificity = keys.filter((key) => a[key] !== null).length;
+    const bSpecificity = keys.filter((key) => b[key] !== null).length;
+
+    // Take the rule with the most specific condition first
+    return bSpecificity - aSpecificity;
+  })[0];
+
+  if (prioritizedRule) {
+    if (prioritizedRule.rootFolder) {
+      rootFolder = prioritizedRule.rootFolder;
+    }
+    if (prioritizedRule.profileId) {
+      profileId = prioritizedRule.profileId;
+    }
+    if (prioritizedRule.tags) {
+      tags = [
+        ...new Set([
+          ...(tags || []),
+          ...prioritizedRule.tags.split(',').map((tag) => Number(tag)),
+        ]),
+      ];
+    }
+
+    logger.debug('Override rule applied.', {
+      label: 'Media Request',
+      overrides: prioritizedRule,
+    });
+  }
+
+  return { rootFolder, profileId, tags };
+}
+
+export default overrideRules;
--- a/server/lib/overrideRules.ts
+++ b/server/lib/overrideRules.ts
@@ -135,7 +135,6 @@ async function overrideRules({
     if (prioritizedRule.tags) {
       tags = [
         ...new Set([
-          ...(tags || []),
           ...prioritizedRule.tags.split(',').map((tag) => Number(tag)),
         ]),
       ];
--- a/server/lib/overrideRules.ts
+++ b/server/lib/overrideRules.ts
@@ -1,0 +1,166 @@
+import { ANIME_KEYWORD_ID } from '@server/api/themoviedb/constants';
+import type {
+  TmdbKeyword,
+  TmdbMovieDetails,
+  TmdbTvDetails,
+} from '@server/api/themoviedb/interfaces';
+import { MediaType } from '@server/constants/media';
+import { getRepository } from '@server/datasource';
+import OverrideRule from '@server/entity/OverrideRule';
+import type { User } from '@server/entity/User';
+import { getSettings } from '@server/lib/settings';
+
+export type OverrideRulesResult = {
+  rootFolder: string | null;
+  profileId: number | null;
+  tags: number[] | null;
+};
+
+async function overrideRules({
+  mediaType,
+  is4k,
+  tmdbMedia,
+  requestUser,
+  tags = null,
+  serviceId,
+}: {
+  mediaType: MediaType;
+  is4k: boolean;
+  tmdbMedia: TmdbMovieDetails | TmdbTvDetails;
+  requestUser: User;
+  tags?: number[] | null;
+  serviceId?: number;
+}): Promise<OverrideRulesResult> {
+  const settings = getSettings();
+
+  let rootFolder: string | null = null;
+  let profileId: number | null = null;
+
+  const serviceIdField =
+    mediaType === MediaType.MOVIE ? 'radarrServiceId' : 'sonarrServiceId';
+  if (serviceId === undefined) {
+    const defaultRadarrId = is4k
+      ? settings.radarr.find((r) => r.is4k && r.isDefault)?.id
+      : settings.radarr.find((r) => !r.is4k && r.isDefault)?.id;
+    const defaultSonarrId = is4k
+      ? settings.sonarr.find((s) => s.is4k && s.isDefault)?.id
+      : settings.sonarr.find((s) => !s.is4k && s.isDefault)?.id;
+    serviceId =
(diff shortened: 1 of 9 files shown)


mysticalsoap:
Hit this exact gap today. I'm using override rules to route anime requests to a dedicated anime quality profile + tagging for Radarr. Confirmed the rule itself matches correctly for normal users, but skips when the request comes from an account with MANAGE_REQUESTS (including self-request and impersonation), landing on the default profile with no tag instead.

Built and tested out this branch with the same setup I have in the live environment and it works and solves the issue. Would be great to see this added.

server/lib/overrideRules.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  const appliedOverrideRules = overrideRules.filter((rule) => {
+    const hasAnimeKeyword =
+      'results' in tmdbMedia.keywords &&
+      tmdbMedia.keywords.results.some(
+        (keyword: TmdbKeyword) => keyword.id === ANIME_KEYWORD_ID
+      );
+
+    // Skip override rules if the media is an anime TV show as anime TV
+    // is handled by default and override rules do not explicitly include
```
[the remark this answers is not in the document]
gauthier-th: This is a copy/paste of the previous override rules.

server/routes/overrideRule.ts
```diff
@@ -61,6 +67,48 @@ overrideRuleRoutes.post<
   }
 });
 
+overrideRuleRoutes.post(
+  '/advancedRequest',
+  isAuthenticated([Permission.REQUEST_ADVANCED, Permission.MANAGE_REQUESTS], {
+    type: 'or',
+  }),
```
fallenbagel, line 92: This permission list never runs. The router is mounted behind an ADMIN check, so anything without the ADMIN bit gets a 403 before reaching here.

https://github.com/seerr-team/seerr/blob/48dd803e047204be4b34cf7f0bb30fe4ad3b9d20/server/routes/index.ts#L181-L185

The feature is admin-only as written, and the client's empty catch hides the 403, so REQUEST_ADVANCED and MANAGE_REQUESTS-without-ADMIN users just get server defaults with no indication of why. This route needs to live outside the admin-mounted router. This is from reading the code, not testing, so confirm it on a plain REQUEST_ADVANCED account and on a MANAGE_REQUESTS account without ADMIN.

server/routes/overrideRule.ts
```diff
@@ -61,6 +67,48 @@ overrideRuleRoutes.post<
...
+  async (req, res, next) => {
+    try {
+      const tmdb = new TheMovieDb();
+      const tmdbMedia =
+        req.body.mediaType === MediaType.MOVIE
+          ? await tmdb.getMovie({ movieId: req.body.tmdbId })
+          : await tmdb.getTvShow({ tvId: req.body.tmdbId });
```
fallenbagel: Declare the generics like the other routes in this file so `req.body` is typed. As written `mediaType` is unvalidated and anything that isn't movie silently takes the TV branch, and `tmdbId` goes into the TMDB call unchecked. I mean sure we only have TV/Movie but it would be cleaner and safer that way.

server/routes/overrideRule.ts
```diff
@@ -61,6 +67,48 @@ overrideRuleRoutes.post<
...
+
+      const userRepository = getRepository(User);
+      const requestUserId = req.user?.hasPermission([
+        Permission.MANAGE_REQUESTS,
+        Permission.MANAGE_USERS,
+      ])
+        ? req.body.requestUser
+        : req.user?.id;
```
fallenbagel: `hasPermission` defaults to `and`, so a MANAGE_REQUESTS user without MANAGE_USERS falls back to their own id. The edit modal passes the request owner as `requestUser`, so rules get evaluated against the wrong user there. The create path evaluates against the owner.

server/routes/overrideRule.ts
```diff
@@ -61,6 +67,48 @@ overrideRuleRoutes.post<
...
+      if (!user) {
+        return next({ status: 404, message: 'User not found.' });
+      }
+
+      const overrideRulesResult: OverrideRulesResult = await overrideRules({
+        mediaType: req.body.mediaType,
+        is4k: req.body.is4k,
+        tmdbMedia,
+        requestUser: user,
+        serviceId: req.body.serviceId,
```
fallenbagel, line 182: `serviceId` is taken verbatim with no check that it belongs to the given `mediaType` or to the 4K tier the caller is allowed to request against. Shouldn't we bind it to the configured servers for that `mediaType` and `is4k` before querying rules?

server/routes/overrideRule.ts
```diff
@@ -61,6 +67,48 @@ overrideRuleRoutes.post<
...
+      });
+
+      res.status(200).json(overrideRulesResult);
+    } catch {
+      next({ status: 404, message: 'Media not found' });
+    }
```
fallenbagel: Every failure in this block reports as `404 Media not found`. A TMDB 5xx or a DB error is not a missing item. Narrow the 404 to the TMDB lookup and let real failures surface as 500.

server/lib/overrideRules.ts
```diff
@@ -0,0 +1,172 @@
...
+          ...(tags || []),
+          ...prioritizedRule.tags.split(',').map((tag) => Number(tag)),
+        ]),
+      ];
+    }
+
+    logger.debug('Override rule applied.', {
+      label: 'Media Request',
+      overrides: prioritizedRule,
+    });
```
fallenbagel: `label: 'Media Request'` is wrong now that the route calls this. It fires on modal open, server switch and user switch, under a label that means request creation.

src/components/RequestModal/AdvancedRequester/index.tsx
```diff
@@ -330,6 +334,46 @@ const AdvancedRequester = ({
...
+          );
+          if (override.rootFolder) {
+            setSelectedFolder(override.rootFolder);
+          }
+          if (override.profileId) {
+            setSelectedProfile(override.profileId);
+          }
+          if (override.tags) {
+            setSelectedTags(override.tags);
+          }
```
fallenbagel: No `defaultOverrides` guard here, unlike the `serverData` effect above. The edit modal passes `tmdbId` alongside overrides loaded from the existing request.

https://github.com/seerr-team/seerr/blob/48dd803e047204be4b34cf7f0bb30fe4ad3b9d20/src/components/RequestModal/MovieRequestModal.tsx#L292-L306

So opening a pending request whose folder or profile was deliberately set to something other than the rule value resets those fields, and `onChange` fires with the rule values before the user touches anything. Skip the apply for any field `defaultOverrides` already covers.

src/components/RequestModal/AdvancedRequester/index.tsx
```diff
@@ -330,6 +334,46 @@ const AdvancedRequester = ({
...
+            '/api/v1/overrideRule/advancedRequest',
+            {
+              mediaType: type,
+              is4k,
+              requestUser:
+                selectedUser?.id ?? requestUser?.id ?? currentUser?.id,
+              tmdbId,
+              serviceId: selectedServer ?? undefined,
+            }
+          );
```
fallenbagel, line 358: No cancellation or stale-response guard. Switching server or user twice in quick succession can land the earlier response over the current selection. Capture the serviceId and user id the request was made for and drop the response if either has moved on.

src/components/RequestModal/AdvancedRequester/index.tsx
```diff
@@ -330,6 +334,46 @@ const AdvancedRequester = ({
...
+          }
+          if (override.profileId) {
+            setSelectedProfile(override.profileId);
+          }
+          if (override.tags) {
+            setSelectedTags(override.tags);
+          }
+        } catch {
+          /* empty */
+        }
```
fallenbagel, line 383: Swallowing this makes a failed lookup indistinguishable from no matching rule. Log the error.

src/components/RequestModal/AdvancedRequester/index.tsx
fallenbagel: This replaces the current selection, while the create path merges rule tags into the submitted tags. Send `selectedTags` with the request and let the helper do the union, so both paths agree.

fallenbagel (changes requested):
There are a few issues currently. More details are inline but here's a tldr:
- The `/overrideRule` endpoint is unreachable for the users this is meant to fix. `/overrideRule` is mounted behind an ADMIN check, so the route's own permission list never runs and every non-admin gets a 403 that the client silently swallows. 
- Also the new effect in `AdvancedRequester` has no `defaultOverrides` guard, so the edit modal resets deliberately-set folder, profile and tags to the rule values on open.
- `useOverrides` still excludes MANAGE_REQUESTS users:

https://github.com/seerr-team/seerr/blob/48dd803e047204be4b34cf7f0bb30fe4ad3b9d20/server/entity/MediaRequest.ts#L236-L239

That means rules only reach a request when it goes through the modal. Direct API requests from those accounts still bypass rules entirely, so the feature is only closed for the UI path. The feature here is meant to apply to advanced requests as a rule, so it belongs on the server, with the client pre-fill as presentation.

- The existing override rule cases in `server/routes/request.test.ts` cover the create path so the extraction is guarded there, but nothing covers the new endpoint.

server/routes/overrideRule.ts
gauthier-th: Ah yes good catch

server/routes/overrideRule.ts
gauthier-th: > The edit modal passes the request owner as requestUser, so rules get evaluated against the wrong user there. The create path evaluates against the owner.

Not sure about what you mean there? (I agree with the permission thing, i have to set it up as `type: 'or'`)

fallenbagel: > > The edit modal passes the request owner as requestUser, so rules get evaluated against the wrong user there. The create path evaluates against the owner.
> 
> Not sure about what you mean there? (I agree with the permission thing, i have to set it up as `type: 'or'`)

None of this can fire today, since everything reaching this route has ADMIN, and `hasPermission` returns true on the ADMIN bit before the array is even looked at, so the check always passes and the fallback never runs. What I meanf is this becomes an issue the moment the route moves out of the admin-mounted router, which is the other comment I mentioned.

Once this is reachable, a user with MANAGE_REQUESTS but no MANAGE_USERS opens someone else's pending request. The modal sends the owner as `requestUser`, since `selectedUser` is seeded from the `requestUser` prop in the edit flow. This route drops it and uses `req.user.id`, so the rules that come back are the caller's, not the owner's (if you have user specific rules). The modal pre-fills with them and PUT writes them straight onto a request that is still owned by the other user. 

That case is only true on the request side. PUT only rejects a `userId` that differs from the current owner, so passing the existing owner is fine without MANAGE_USERS:
 https://github.com/seerr-team/seerr/blob/48dd803e047204be4b34cf7f0bb30fe4ad3b9d20/server/routes/request.ts#L495-L513

And createRequest throws rather than falling back to the caller:
 https://github.com/seerr-team/seerr/blob/48dd803e047204be4b34cf7f0bb30fe4ad3b9d20/server/entity/MediaRequest.ts#L56-L72

So just swapping to `or` doesn't fix this. A MANAGE_REQUESTS-only caller has no MANAGE_USERS, so `createRequest` rejects any `userId` in the body and PUT only accepts the id already on the request. With `or` this endpoint would hand them rules for any user id they send, for a request they have no way to create or edit.

Match the request path instead:
- creating: only accept a `requestUser` the caller could legally pass to `createRequest`, otherwise 403 rather than falling back to `req.user.id`
- editing: the caller is allowed to pass the existing owner, so the endpoint needs the request id to confirm that's who `requestUser` is

server/routes/overrideRule.ts
gauthier-th: > `serviceId` is taken verbatim with no check that it belongs to the given `mediaType` or to the 4K tier the caller is allowed to request against. Shouldn't we bind it to the configured servers for that `mediaType` and `is4k` before querying rules?

True, but we don't check that anywhere else too (e.g. in [`MediaRequest.request`](https://github.com/seerr-team/seerr/blob/develop/server/entity/MediaRequest.ts#L408))

fallenbagel: > > `serviceId` is taken verbatim with no check that it belongs to the given `mediaType` or to the 4K tier the caller is allowed to request against. Shouldn't we bind it to the configured servers for that `mediaType` and `is4k` before querying rules?
> 
> True, but we don't check that anywhere else too (e.g. in [`MediaRequest.request`](https://github.com/seerr-team/seerr/blob/develop/server/entity/MediaRequest.ts#L408))

We should though but yeah that could be a separate pr then

server/routes/overrideRule.ts
gauthier-th: Since this endpoint doesn't perform any action, wdyt of not checking the perm for the requestUser and just use the one provided by the UI? Permission checks would already have been done on the UI side.

src/components/RequestModal/AdvancedRequester/index.tsx
gauthier-th: Yes this is a race condition that could happen, but i'm pretty sure we're not checking any of those elsewhere in our codebase?

fallenbagel: > Yes this is a race condition that could happen, but i'm pretty sure we're not checking any of those elsewhere in our codebase?

We have started to. Atleast on the quota and editing request part. Better to start somewhere no? Because i have personally reproduced this. We can slowly start properly fixing them so why not check here too? 🤔

server/routes/overrideRule.ts
gauthier-th: Should be fixed

src/components/RequestModal/AdvancedRequester/index.tsx
```diff
@@ -330,6 +338,69 @@ const AdvancedRequester = ({
...
+      }
+    })();
+    return () => {
+      cancelled = true;
+    };
+  }, [
+    tmdbId,
+    type,
+    is4k,
```
fallenbagel, line 392: `selectedTags` is read here and written at line 373 (https://github.com/seerr-team/seerr/pull/2164/changes#diff-228b85a3a341ef3d89dcd85a8802a38245c7cf95e2acc2e65bd7a04893930ba0R373), so every tag change refires the lookup. Wouldn't that mean deselecting a rule tag just puts it back on the next round trip? And since `defaultOverrides?.folder` is undefined on the create path, the refire would reset a manually picked folder and profile to the rule values too. So keeping it out of the deps and reading it from a ref for the payload should cover both no?

fallenbagel (changes requested):

src/components/RequestModal/AdvancedRequester/index.tsx
gauthier-th: For the tags it was ok during my tests.
But yes I didn't notice the profile selection change.

fallenbagel (approved):
LGTM. haven't tested

M0NsTeRRR (approved):

PR #2165, opened 2025-11-22
feat(imdb ratings): add IMDb ratings for TV Shows

samohtxotom:
#### Description

Currently TV Shows do not provide an IMDb rating, there is no official API and so we use Radarr's IMDb ratings for Movies, however they do not provide TV Ratings.

[Agregarr](https://github.com/agregarr/agregarr) hosts a free IMDb ratings proxy for both Movies and TV Shows. 

This PR adds IMDb ratings for TV Shows using the Agregarr API, keeping the original Radarr method for Movies due to its tested reliability. 

imdbratings

#### To-Dos

### DIFF
--- a/seerr-api.yml
+++ b/seerr-api.yml
@@ -6725,6 +6725,66 @@ paths:
                   criticsRating:
                     type: string
                     enum: ['Rotten', 'Fresh']
+  /tv/{tvId}/ratingscombined:
+    get:
+      summary: Get RT and IMDB TV ratings combined
+      description: Returns ratings from RottenTomatoes and IMDB based on the provided tvId in a JSON object.
+      tags:
+        - tv
+      parameters:
+        - in: path
+          name: tvId
+          required: true
+          schema:
+            type: number
+            example: 76479
+      responses:
+        '200':
+          description: Ratings returned
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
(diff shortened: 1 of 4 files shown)


M0NsTeRRR:
I reviewed the service, and I agree with Gauthier’s comment. According to Agregarr’s documentation, the service can be self-hosted. In the future, it would be beneficial to introduce configurable settings that allow users to choose between a self-hosted instance or the hosted version, this would address my concern about relying on a third-party service by default. This new service would be opt-in.

However, since the service is new and there hasn’t been any demand from other users, I'm also in favor of closing this PR for now. We can always reopen the discussion if there’s a need later of course.

agregarr:
I totally agree and I would have preferred anything other than creating my own service for this, but AFAIK there is no other way to gather IMDb ratings for TV shows other than through Plex for existing content, or through a paid API such as OMDB. 

It's worth noting that the [current Radarr proxy](https://github.com/seerr-team/seerr/blob/bde322de8e67c1c602edaafe646c816511a66bb7/server/api/rating/imdbRadarrProxy.ts#L151C1-L201C32) is also third party. Without that, Seerr would have no IMDb ratings at all. But it seems unreasonable to expect Radarrs proxy to provide TV ratings as well. Ideally, Sonarr would offer a similar service. Thanks for the quick review 👍

PR #2175, opened 2025-11-25
chore: improve PR template

M0NsTeRRR:
#### Description

Improve PR template to help triage.

#### To-Dos

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -19,17 +19,6 @@
 
 ## Screenshots / Logs (if applicable)
 
-## Types of Changes
-
-<!--- What types of changes does your code introduce? Put an `x` in all the boxes that apply: -->
-
-- [ ] Bug fix (non-breaking change which fixes an issue)
-- [ ] New feature (non-breaking change which adds functionality)
-- [ ] Breaking change (fix or feature that would cause existing functionality to change)
-- [ ] Documentation (no code change)
-- [ ] Refactor (refactoring production code)
-- [ ] Other <!--- (provide information) -->
-
 ## Checklist:
 
 <!--- Go over all the following points, and put an `x` in all the boxes that apply. -->
@@ -42,10 +31,3 @@
 - [ ] Successful build `pnpm build`
 - [ ] Translation keys `pnpm i18n:extract`
 - [ ] Database migration (if required)
-
-<!--- Thank you for keeping this note for the community --->
-
-### Community Note
-
-- Please vote on this pull request by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original pull request comment to help the community and maintainers prioritize this request
-- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for pull request followers and do not help prioritize the request


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,14 +1,51 @@
...
 - [ ] Database migration (if required)
 
-#### Issues Fixed or Closed
+<!--- Thank you for keeping this note for the community --->
 
-- Fixes #XXXX
+### Community Note
+
+- Please vote on this pull request by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original pull request comment to help the community and maintainers prioritize this request
+- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for pull request followers and do not help prioritize the request
```
gauthier-th: Do you think this is really necessary? I feel like this will mostly be useless and will only add some more text.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,14 +1,51 @@
...
+## Types of Changes
+
+<!--- What types of changes does your code introduce? Put an `x` in all the boxes that apply: -->
+
+- [ ] Bug fix (non-breaking change which fixes an issue)
+- [ ] New feature (non-breaking change which adds functionality)
+- [ ] Breaking change (fix or feature that would cause existing functionality to change)
+- [ ] Documentation (no code change)
+- [ ] Refactor (refactoring production code)
+- [ ] Other <!--- (provide information) -->
```
gauthier-th: We already have git hook that makes user choose this, and this belongs to the PR/commit title, not the description?

.github/PULL_REQUEST_TEMPLATE.md
M0NsTeRRR: Yes so we can hide message on PR without justify ourself on the why (lidarr, oidc etc... could benefit from this)

.github/PULL_REQUEST_TEMPLATE.md
M0NsTeRRR: Yeah that make sense :)

.github/PULL_REQUEST_TEMPLATE.md
gauthier-th: I personally have no problem hiding these messages even without justification, I don't really see what this will help.

M0NsTeRRR: For me it's a reminder and it's never stated (or I don't know where) somewhere we have to follow this guideline (even if that's common sense)

gauthier-th: We could add this in the code of conduct then? IMO it doesn't concern the PR itself or the PR author, just the users. And I don't think we should state everywhere what is already common sense.

M0NsTeRRR: > We could add this in the code of conduct then

I can add it there also yes.

> it doesn't concern the PR

I agree but we are going to have the same issue as AI it's why we put it in the checklist

gauthier-th: Same issue as AI? It's just a few comments, I really don't have any issue to mark them as spam whenever it's spam. No need to clutter each PR with that. But that's just my opinion.

0xSysR3ll (approved):

PR #2192, opened 2025-12-03
docs: fix chown command for windows users

M0NsTeRRR:
## Description

fix chown command for windows users

## Checklist:

### DIFF
--- a/docs/migration-guide.mdx
+++ b/docs/migration-guide.mdx
@@ -28,7 +28,7 @@ Changes :
 
 If you're migrating from a previous installation, you may need to update the ownership of your config folder:
 ```bash
-sudo chown -R 1000:1000 /path/to/appdata/config
+docker run --rm -v /path/to/appdata/config:/data alpine chown -R 1000:1000 /data
 ```
 
 This ensures the `node` user (UID 1000) owns the config directory and can read and write to it.


docs/migration-guide.mdx
```diff
@@ -28,7 +28,7 @@ Changes :
 
 If you're migrating from a previous installation, you may need to update the ownership of your config folder:
 ```bash
-sudo chown -R 1000:1000 /path/to/appdata/config
+docker run --rm -v /path/to/appdata/config:/data alpine chown -R 1000:1000 /data
```
gauthier-th, line 31: Couldn't we run this command in the existing container instead of creating a new one just for the chown?

M0NsTeRRR: The new Seerr container will not have the necessary permissions to perform this action, as the container now runs rootless. To simplify the process for users, it would be easier to provide a temporary container (using the `--rm` flag) that can be executed at any time. This approach also addresses cases where users have already migrated from Overseerr or Jellyseerr to the Seerr image (without reading the documentation :skull:) and encountered the permission issue we observed in the `#seerr-beta` channel. They can run the temporary container afterward, avoiding the need to revert to the Overseerr or Jellyseerr image and execute this command.

gauthier-th: Okay I get it. Makes sense.

PR #2225, opened 2025-12-13
fix(local-login): remove automatic plex linking & reduce logout log verbosity

fallenbagel:
## Description
This PR cleans up the local login authentication flow by removing the automatic Plex user discovery and linking logic that was causing unnecessary API calls and 401 errors for Jellyfin/Emby users. This functionality is already properly handled through the new Linked Accounts settings page where users can explicitly manage their external account connections. Also reduces log noise by changing the successful logout message from info to debug level since it's expected, routine behavior that doesn't require monitoring.

## How Has This Been Tested?
- Logged in with local credentials on a Jellyfin setup instance and verified no Plex API errors in logs
- Confirmed session is created correctly and user data is returned
- Logged out and verified session is destroyed
- Checked logs to confirm logout success message is only visible at debug level
 
## Checklist:

### DIFF
--- a/server/routes/auth.ts
+++ b/server/routes/auth.ts
@@ -626,76 +626,6 @@ authRoutes.post('/local', async (req, res, next) => {
       });
     }
 
-    const mainUser = await userRepository.findOneOrFail({
-      select: { id: true, plexToken: true, plexId: true },
-      where: { id: 1 },
-    });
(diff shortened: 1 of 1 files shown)


0xSysR3ll (approved):
LGTM !

PR #2261, opened 2025-12-29
fix(sonarr): search only requested seasons

Shlawpers:
## Description

What it does (summary):
Prevents seer from triggering a series-wide search in Sonarr, when not all seasons are requested.

Runs Sonarr SeasonSearch only for the requested season number.
Skips searching seasons that appear fully downloaded (via season statistics) to avoid wasted searches
Key file: sonarr.ts

AI disclosure: I used AI assistance in writing this patch, both OpenAI Codex (GPT-5.2) and Claude opus 4.5. I tested and reviewed.  manually.

- Fixes 
sonarr.ts

## How Has This Been Tested?

Requested a TV series with only Season 1 selected + search now; verified Sonarr only received SeasonSearch for that season (no full-series search). Did this with both existing series with missing seasons and a new series addition. Then Re-requested a fully-downloaded season; verified search is skipped.

Testing env was docker based setup with sonarr running in same stack.  Change was isolated to sonarr.ts so I don't think it impacts anything outside of this.

- [ ] Translation keys `pnpm i18n:extract`  #NA? 
- [ ] Database migration (if required) #NA?

apologies if I messed anything up this is my first pr

### DIFF
--- a/.github/renovate/docker.json5
+++ b/.github/renovate/docker.json5
@@ -1,10 +1,7 @@
 {
   $schema: 'https://docs.renovatebot.com/renovate-schema.json',
 
-  extends: [
-    'docker:enableMajor',
-    'docker:pinDigests'
-  ],
+  extends: ['docker:enableMajor', 'docker:pinDigests'],
 
   packageRules: [
     {
--- a/.github/renovate/pnpm.json5
+++ b/.github/renovate/pnpm.json5
@@ -7,5 +7,4 @@
   lockFileMaintenance: {
     enabled: true,
   },
-
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -20,8 +20,6 @@
   "files.associations": {
     "globals.css": "tailwindcss"
   },
-  "i18n-ally.localesPaths": [
-    "src/i18n/locale"
-  ],
+  "i18n-ally.localesPaths": ["src/i18n/locale"],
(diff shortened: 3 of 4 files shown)


fallenbagel:
I haven't really reviewed this but wouldn't this break anime type media as season searches do not work on them.

Shlawpers:
> wouldn't this break anime type media as season searches do not work on them.

ack, that may well be — the initial fix used Sonarr SeasonSearch per requested season, which I didn't realize could cause issues for anime. 

I pushed an update (commit dca857c7) so anime requests no longer use SeasonSearch.

For anime it'll now fetch episodes for the requested season(s) via GET /api/v3/episode?seriesId=...&seasonNumber=..., filter to missing + monitored episodes, and trigger EpisodeSearch with those episodeIds (chunked). Standard/daily series still use SeasonSearch. This keeps the fix (no series-wide searching when only one season is requested) and I think this should hopefully avoid breaking anime while still preventing whole-series searches when only a single season is requested.

zone-zero:
This would also solve the issue of search requests from Seerr causing sonarr to only download individual episodes, rather than season packs.

fallenbagel:
> This would also solve the issue of search requests from Seerr causing sonarr to only download individual episodes, rather than season packs.

Not necessarily. That pretty much iirc depends on your sonarr configuration (i say this as I always get season packs with the current behaviour as well).

PR #2282, opened 2026-01-09
Feat: No id in tag when tagging requester

### LINKED ISSUE #727: Per user tag added to sonarr & radarr includes user id number
### Description

Please remove "USER NUMBER" from the User tagging being sent to the *arr.

Feature of tagging requests in the *arr with the user who requested it ([https://github.com/Fallenbagel/jellyseerr/issues/201](https://github.com/Fallenbagel/jellyseerr/issues/201))seems to work well, thanks!!!

The minor issue/irritation with it is that it lists the user id # too.

The user number assigned to that user is not useful info. I'm not certain you can even have users of the same name on emby/jellyfin, but even if you could, the user number is not really helpful.

As an aside, I'm giving Jellyseerr a try over Ombi, see how it fares. So far, so good. But it did suprise me when I saw the tags included numbers for some odd reason suddenly.

### Desired Behavior

When user tagging is enabled, the tag sent to the arr should be ONLY the user_name.

### Additional Context

_No response_

### Code of Conduct

nodiaque:
## Description

Added a new checkbox in the Radarr and Sonarr option under the Tag Request to allow the creation of tag without an id.

Request made in the community

- Fixes #727 

## How Has This Been Tested?

Per the guide, ran in pnpm dev on ubuntu, connected to my running emby + sonarr + radarr, created a request in both and saw the new request with the proper tag.

ubuntu LXC running under Unraid. Docker running Emby, Sonarr and Radarr. I ran under my Ubuntu machine using the guide with pnpm dev from scratch, added the services for both radarr and sonarr, made request in both and saw the tag added.

Not the best code under MediaRequestSubscriber.ts but it work. Could be optimize.
image

## Checklist:

- [ x] I have read and followed the contribution [guidelines](https://github.com/seerr-team/seerr/blob/develop/CONTRIBUTING.md).
- [ x] Disclosed any use of AI (see our [policy](https://github.com/seerr-team/seerr/blob/develop/CONTRIBUTING.md#ai-assistance-notice))
- [ x] I have updated the documentation accordingly.
- [ x] All new and existing tests passed.
- [ x] Successful build `pnpm build`
- [ x] Translation keys `pnpm i18n:extract` (must I commit after running this?)
- [ x] Database migration (if required) (not required)


gauthier-th:
As said in #727, I would much prefer a feature to choose between 
- none
- userId
- userId - username
- username

Instead of just a checkbox to remove the userId.

nodiaque:
I understand but it's a step. The main issue is the fact that user migrating from ombi have 2 set of tag. This is a first step in the right direction. Implementing drop down with multiple choice is doable but require more code. This at least provide a solution for the issue.

fallenbagel:
> I understand but it's a step. The main issue is the fact that user migrating from ombi have 2 set of tag. This is a first step in the right direction. Implementing drop down with multiple choice is doable but require more code. This at least provide a solution for the issue. 

Better to implement it in this pr since its within scope. Shouldn't be too complex. Just need to add a dropdown and some conditions on when creating tags

nodiaque:
I'll be OK with it but there's 2 thing to keep in mind.
1) Make it backward compatible
2) I don't know how to make dropdown, I'll have to check other dropdown in that list to see how to do it. I'll check later this week how to change for that.

edit: With my new understanding of the project, I understand the "take the time to do it" since the release isn't for right now. I was making this quick change to fix and have a step, I'm used to quick release cadence ;)

fallenbagel:
> I'll be OK with it but there's 2 thing to keep in mind.
> 1) Make it backward compatible
> 2) I don't know how to make dropdown, I'll have to check other dropdown in that list to see how to do it. I'll check later this week how to change for that.
> 

It will be backward compatible. By default you keep it how it is. Only allow users to change it. 
For dropdown you can take a look at the dropdown component we have.

fallenbagel:
> edit: With my new understanding of the project, I understand the "take the time to do it" since the release isn't for right now. I was making this quick change to fix and have a step, I'm used to quick release cadence ;)

Quick release is not our policy. We strive for stability and reliability rather than "as long as it works" ;)

nodiaque:
> > edit: With my new understanding of the project, I understand the "take the time to do it" since the release isn't for right now. I was making this quick change to fix and have a step, I'm used to quick release cadence ;)
> 
> Quick release is not our policy. We strive for stability and reliability rather than "as long as it works" ;)

No problem. It was in fact to fix the main issue I did it like that. It was stable and functionnal. I'll check the other dropdown like you said to learn on it and push a new update.

nodiaque:
> > I'll be OK with it but there's 2 thing to keep in mind.
> > 
> > 1. Make it backward compatible
> > 2. I don't know how to make dropdown, I'll have to check other dropdown in that list to see how to do it. I'll check later this week how to change for that.
> 
> It will be backward compatible. By default you keep it how it is. Only allow users to change it. For dropdown you can take a look at the dropdown component we have.

The issue I was seeing is I was removing the checkbox from "tag requested" and put a dropdown with all the option including none (thus removing the new entry I just put). I could do simply like I did right now with a dropdown as another option (instead of the new checkbox I added). I just didn't find how to make the dropdown enabled/disabled based on the checkbox.

fallenbagel:
> > > I'll be OK with it but there's 2 thing to keep in mind.
> > > 
> > > 1. Make it backward compatible
> > > 2. I don't know how to make dropdown, I'll have to check other dropdown in that list to see how to do it. I'll check later this week how to change for that.
> > 
> > It will be backward compatible. By default you keep it how it is. Only allow users to change it. For dropdown you can take a look at the dropdown component we have.
> 
> The issue I was seeing is I was removing the checkbox from "tag requested" and put a dropdown with all the option including none (thus removing the new entry I just put). I could do simply like I did right now with a dropdown as another option (instead of the new checkbox I added). I just didn't find how to make the dropdown enabled/disabled based on the checkbox.

Never said to remove that option. That option should be kept with a child option that gets enabled when selected. Look at the settings > users page to understand what I mean.

nodiaque:
Yeah I guess you mean these?

image

This is a bit what I had in mind, which would make this (forget the value it's just a paint edit):
image

The problem I see here. The current code check the value of the checkbox is true or false. Now, it's a dropdown with value "none, userid-username,....". There's no way to translate that from current checkbox value.

Because of that, I'm forced to do something like this:
image

Or on 2 lines.

Unless there's a way I don't see right now to not lose the current config when removing the checkbox, I don't know all of the project code so maybe I'm missing a service helper or something.

fallenbagel:
> Yeah I guess you mean these?
> 
> image
> 
> This is a bit what I had in mind, which would make this (forget the value it's just a paint edit):
> image
> 
> The problem I see here. The current code check the value of the checkbox is true or false. Now, it's a dropdown with value "none, userid-username,....". There's no way to translate that from current checkbox value.
> 
> Because of that, I'm forced to do something like this:
> image
> 
> Or on 2 lines.
> 
> Unless there's a way I don't see right now to not lose the current config when removing the checkbox, I don't know all of the project code so maybe I'm missing a service helper or something.


No. None of that. I am talking about how there is a parent and a child. Look at the perms. You see how one tick enables the other fields. Like that have the selection as a SEPARATE field.

nodiaque:
Got ya. I'll check the code on that. Thanks

nodiaque:
I'm trying to understand how to do a dropdown box. It seems all dropdown box in the project use a different way and is overly complicated. I guess that's what happen when you aren't a UI dev, specially not web base

M0NsTeRRR:
Hello @nodiaque, is that okay for you if I close your PR in favor of #2504 (you can give feedback/comment there for sure) ?

MagicLegend:
Ah! @M0NsTeRRR is ahead of me, but yes - @nodiaque please do give feedback on the changes in #2504 :) I was lazy and cheated to get Sonnet to implement the dropdown for me ;) But it used your logic as a baseline!

nodiaque:
If there's a better change then what I did, I'm ok with id. I was struggling to get the html to display properly a dropdown box and haven't work anymore on it

M0NsTeRRR:
supersede by #2504

PR #2342, opened 2026-01-27
chore(deps): update dependencies and fix security vulnerabilities

fallenbagel:
## Description
Updates project dependencies to address security vulnerabilities and updated dependencies. 

TypeScript has been upgraded to 5.4 to enable Zod 4 support, which required the `const` type parameter syntax introduced in TS 5.0.  In addition, since `prettier-plugin-organize-imports` behaves differently with TypeScript 5.4
vs 4.9, causing CI formatting checks to fail, the imports have also been ran through prettier to match the ordering expected by the plugin with the upgraded TS version.

Direct dependencies with known vulnerabilities (nodemailer, xml2js, undici, lodash) have been updated to their patched versions.

For transitive dependencies that cannot be updated directly without breaking compatibility (tar, tar-fs, qs, jws, @babel/runtime, etc.), I have just left them untouched.

I did not update the tailwind-merge package from v2 as v3 changes class merging behavior for opacity utilities, which would require updating quite a lot of files across the codebase and I think it would be better if we leave that for a later update (after seerr release).

## How Has This Been Tested?
- Ran dev environment with no issues
- Built, and tested with no issues

## Checklist:


package.json
```diff
@@ -130,7 +130,7 @@
     "@types/mime": "3",
     "@types/node": "22.10.5",
     "@types/node-schedule": "2.1.8",
-    "@types/nodemailer": "6.4.7",
+    "@types/nodemailer": "7",
```
gauthier-th, line 133: Why not `7.0.12`?

fallenbagel: I see it as 7.0.12 on web. 7 as mobile. Locally it shows as 7.0.12 too. I think this is a github caching issue

PR #2347, opened 2026-01-27
ci(workflow): validate i18n locale files are synchronized

0xSysR3ll:
## Description

I don’t remember exactly when or where we discussed this, but we've noticed that contributors (CC @fallenbagel) sometimes forget to run `pnpm i18n:extract`.

This PR adds a i18n check job to the current CI to prevent outdated or missing translation keys from being merged. It does so by running `pnpm i18n:extract` and comparing the results, and raising an error if needed.

- Fixes #XXXX

## How Has This Been Tested?

- Verified the CI workflow runs successfully when locale files are in sync (see this [CI run](https://github.com/0xSysR3ll/jellyseerr/actions/runs/21548667896/job/62093976773))
- Tested that the check fails appropriately when locale files are out of sync (see this [CI run](https://github.com/0xSysR3ll/jellyseerr/actions/runs/21548639417/job/62093899036))

## Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # v5.0.0
         with:
-          node-version: '22'
+          node-version-file: 'package.json'
 
       - name: Get pnpm store directory
         shell: bash


fallenbagel:
> I don’t remember exactly when or where we discussed this, but we've noticed that contributors (CC @fallenbagel) sometimes forget to run `pnpm i18n:format`.
> 
~~Why you gotta expose me like that 😭~~
What he means is everyone else forgets, so CC'ng me because I agree 👀

.github/workflows/ci.yml
```diff
@@ -60,6 +60,17 @@ jobs:
...
+      - name: i18n Check
+        run: |
+          cp src/i18n/locale/en.json src/i18n/locale/en.json.bak
+          pnpm i18n:extract
+          if ! cmp -s src/i18n/locale/en.json src/i18n/locale/en.json.bak; then
+            echo "i18n messages are out of sync. Please run 'pnpm i18n:extract' and commit the changes."
+            rm src/i18n/locale/en.json.bak
+            exit 1
+          fi
+          rm src/i18n/locale/en.json.bak
```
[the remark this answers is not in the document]
0xSysR3ll: Damn typo

fallenbagel:
https://github.com/seerr-team/seerr/pull/2299#issuecomment-3756924831

.github/workflows/ci.yml
```diff
@@ -60,6 +60,17 @@ jobs:
       - name: Formatting
         run: pnpm format:check
 
+      - name: i18n Check
+        run: |
+          cp src/i18n/locale/en.json src/i18n/locale/en.json.bak
+          pnpm i18n:extract
+          if ! cmp -s src/i18n/locale/en.json src/i18n/locale/en.json.bak; then
+            echo "i18n messages are out of sync. Please run 'pnpm i18n:extract' and commit the changes."
```
gauthier-th: Could we also send a message in the related PR so that the author can understand more quickly what's the issue?

0xSysR3ll: Yep, could be worth.

0xSysR3ll:
Isn't it a bit too overkill now ?

.github/workflows/ci.yml
```diff
@@ -26,6 +26,9 @@ jobs:
     name: Lint & Test Build
     if: github.event_name == 'pull_request'
     runs-on: ubuntu-24.04
+    permissions:
+      contents: read
```
M0NsTeRRR, line 30: It's already configured at the workflow level

.github/workflows/ci.yml
```diff
@@ -60,6 +63,63 @@ jobs:
       - name: Formatting
         run: pnpm format:check
 
+      - name: i18n Check
+        if: github.event_name == 'pull_request'
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd
```
M0NsTeRRR: Can you add the tag at the end
`actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0`

.github/workflows/ci.yml
```diff
@@ -60,6 +63,63 @@ jobs:
...
+        with:
+          script: |
```
M0NsTeRRR: wouldn't make it sense to put this file inside `bin` and execute it inside the CI ?

.github/workflows/ci.yml
```diff
@@ -26,6 +26,9 @@ jobs:
     name: Lint & Test Build
     if: github.event_name == 'pull_request'
     runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+      pull-requests: write
```
M0NsTeRRR, line 31: is this permission needed ?

M0NsTeRRR (changes requested):

.github/workflows/ci.yml
```diff
@@ -60,6 +63,63 @@ jobs:
       - name: Formatting
         run: pnpm format:check
 
+      - name: i18n Check
+        if: github.event_name == 'pull_request'
```
M0NsTeRRR: This condition is already verified at the job level

.github/workflows/ci.yml
0xSysR3ll: I think it's needed for the bot to  edit and comment the PR yes. But I can test without it.

.github/workflows/ci.yml
0xSysR3ll: The script you mean ?

M0NsTeRRR: yes

.github/workflows/ci.yml
M0NsTeRRR: oh okay I didnd't see you were using github API so yes it's required.

.github/workflows/ci.yml
```diff
@@ -13,6 +13,7 @@ on:
 
 permissions:
   contents: read
+  pull-requests: write
```
M0NsTeRRR: Could you add the permission to the required job only ? (security best practice) otherwise every job in this PR will have `write` permission.
Maybe a dedicated job would be the best to limit this.

0xSysR3ll: Yeah I agree, but I don't really know what's best here. Dedicated job or dedicated workflow ?
@fallenbagel @gauthier-th wdyt ?

M0NsTeRRR: dedicated workflow looks like overkill to me, it's still `ci` business so independant job with the needed steps to make it works :)

gauthier-th: Yep, dedicated job seems good to me

.github/workflows/ci.yml
```diff
@@ -22,9 +22,73 @@ concurrency:
...
+        with:
+          persist-credentials: false
+
+      - name: Pnpm Setup
+        uses: pnpm/action-setup@41ff72655975bd51cab0327fa583b6e92b6d3061 # v4.2.0
+
+      - name: Set up Node.js
+        uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # v5.0.0
+        with:
+          node-version: '22'
```
M0NsTeRRR: Would be better to use `node-version-file`

.github/workflows/ci.yml
```diff
@@ -22,9 +22,73 @@ concurrency:
...
+            retry gh pr comment "$NUMBER" -R "$GH_REPO" -b "$BODY" || true
+          else
+            retry gh pr edit "$NUMBER" -R "$GH_REPO" --remove-label "i18n-out-of-sync" || true
+          fi
+          exit $check_failed
+
   test:
     name: Lint & Test Build
     if: github.event_name == 'pull_request'
+    needs: i18n
```
M0NsTeRRR: I think it’s better for CI jobs to continue running all tasks even if one fails. This way, you get all the information in a single run and avoid having to rerun the pipeline multiple times.

M0NsTeRRR (changes requested):

.github/workflows/ci.yml
0xSysR3ll: Didn't know about this one, thanks !

Issue #2315, opened 2026-01-19
[Bug] MediaRequest records created with NULL mediaId


fallenbagel:
### Description

I found an intermittent issue where `media_request` records are created with `mediaId = NULL`, causing a `TypeError: can't access property "tmdbId", P.media is null error` when viewing the user profile page.

The `MediaRequest.request()` static method in `server/entity/MediaRequest.ts` saves the media entity and then creates a `MediaRequest` with the media object reference (Ref: [Movie](https://github.com/seerr-team/seerr/blob/bb2120c14d4bf2582a8be7ca57a4db1310a947a6/server/entity/MediaRequest.ts#L333-L376), [Tv](https://github.com/seerr-team/seerr/blob/bb2120c14d4bf2582a8be7ca57a4db1310a947a6/server/entity/MediaRequest.ts#L443-L507)). TypeORM is expected to automatically extract the `media.id` and populate the `mediaId` foreign key column, but this implicit relation-to-foreign-key mapping intermittently fails sometimes.

I digged around my database for evidence:
| id | mediaId | type | createdAt |
|----|---------|------|-----------|
| 2544 | NULL | tv | 2025-07-30 05:49:25 |
| 2817 | NULL | movie | 2026-01-10 19:00:43 |
| 2836 | NULL | tv | 2026-01-16 12:29:04 |
| 2839 | NULL | tv | 2026-01-17 00:48:21 |

The media records **are** being created successfully (just milliseconds before the request), but the `mediaId` foreign key isn't being written:
| Request ID | Request Time | Media ID | Media Time | Delta |
|------------|--------------|----------|------------|-------|
| 2544 | 05:49:25.963 | 3264 | 05:49:25.947 | 16ms |
| 2817 | 19:00:43.017 | 3749 | 19:00:43.009 | 8ms |
| 2836 | 12:29:04.368 | 3768 | 12:29:04.361 | 7ms |
| 2839 | 00:48:21.048 | 3770 | 00:48:21.034 | 14ms |

### Version

develop (bb2120c14d4bf2582a8be7ca57a4db1310a947a6)

### Steps to Reproduce

1. Make media requests (issue is intermittent, not reliably reproducible)
2. Eventually a request will be created with `mediaId = NULL`
3. Go to /profile
4. See "Oops" error screen
5. Browser console shows: `TypeError: can't access property "tmdbId", P.media is null`

To check for orphaned requests:
```sql
[...]

PR #2356, opened 2026-01-29
fix(media-request-subscriber): prevent mediald nullification from cascade saves

fallenbagel:
## Description
The `mediaId` FK on `media_request` rows was being set to NULL shortly after insert. This was caused by `mediaRepository.save(media)` being called with the `requests` relation loaded, which triggered TypeORM's cascade `['insert', 'update']` reconciliation on all loaded request entities, corrupting the FK. The issue was compounded by these saves not being awaited in `updateParentStatus`, creating race conditions between concurrent subscriber methods (`updateParentStatus`, `sendToSonarr`, `sendToRadarr`) all firing from `afterInsert`.

Additionally, previously, declining a TV request did not reset the associated season statuses back to UNKNOWN, preventing those seasons from being re-requested. The pending request count check also did not exclude the request being declined or scope to the correct `is4k` variant, causing the media status to remain stuck at PENDING.

Related to #2316
```
[info][Media Request]: Sent request to Sonarr {"requestId":2871,"mediaId":3802}
[info][Media Request]: Sent request to Sonarr {"requestId":2872,"mediaId":3803}
[info][Media Request]: Sent request to Sonarr {"requestId":2873,"mediaId":3804}
```

Yet the database showed `NULL` for the `mediaId` column on those same records:
```
  id  | mediaId |           createdAt           
------+---------+-------------------------------
 2871 |         | 2026-01-28 04:35:16.162949+00
 2872 |         | 2026-01-28 04:37:23.039107+00
 2873 |         | 2026-01-28 04:45:12.229032+00
```

- Fixes https://github.com/seerr-team/seerr/issues/2315
- Fixes https://github.com/seerr-team/seerr/issues/2408

## How Has This Been Tested?
- Auto-approved request (movie & TV): `mediaId` not null
- Pending request → manual approve: sent to Radarr/Sonarr, notifications fire
- Decline movie request: media status resets to UNKNOWN
- Decline TV request (last pending): media status resets to UNKNOWN
- Decline TV request (other pending exists): media status stays PENDING
- Decline TV request: season statuses reset, can re-request
- Delete a request: media status resets appropriately

## Checklist:

### DIFF
--- a/server/subscriber/MediaRequestSubscriber.ts
+++ b/server/subscriber/MediaRequestSubscriber.ts
@@ -828,7 +828,6 @@ export class MediaRequestSubscriber implements EntitySubscriberInterface<MediaRe
       entity.status === MediaRequestStatus.DECLINED &&
       media[statusKey] === MediaStatus.PENDING
     ) {
-      const requestRepository = getRepository(MediaRequest);
       const pendingCount = await requestRepository.count({
         where: {
           media: { id: media.id },
@@ -868,7 +867,7 @@ export class MediaRequestSubscriber implements EntitySubscriberInterface<MediaRe
           (s) => s.seasonNumber === seasonRequest.seasonNumber
         );
 
-        if (season && season[statusKey] == MediaStatus.PENDING) {
+        if (season && season[statusKey] === MediaStatus.PENDING) {
           const otherActiveRequests = await requestRepository
             .createQueryBuilder('request')
             .leftJoinAndSelect('request.seasons', 'season')


gauthier-th:
Related to #2316

server/entity/MediaRequest.ts
```diff
@@ -533,11 +532,9 @@ export class MediaRequest {
     eager: true,
     onDelete: 'CASCADE',
   })
+  @JoinColumn({ name: 'mediaId' })
   public media: Media;
```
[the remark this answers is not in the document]
fallenbagel: Cleaning up orphaned rows (requests without media are invalid data anyway) and keep it non-nullable. Cleaner than adding null checks everywhere.

server/subscriber/MediaRequestSubscriber.ts
```diff
@@ -709,8 +709,10 @@ export class MediaRequestSubscriber implements EntitySubscriberInterface<MediaRe
           .catch(async () => {
             const requestRepository = getRepository(MediaRequest);
 
-            entity.status = MediaRequestStatus.FAILED;
-            requestRepository.save(entity);
+            if (entity.status !== MediaRequestStatus.FAILED) {
+              entity.status = MediaRequestStatus.FAILED;
+              await requestRepository.save(entity);
```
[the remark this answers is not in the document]
fallenbagel: Copilot seems to be getting dumber and dumber

server/subscriber/MediaRequestSubscriber.ts
```diff
@@ -399,8 +399,10 @@ export class MediaRequestSubscriber implements EntitySubscriberInterface<MediaRe
...
```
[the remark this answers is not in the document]
fallenbagel: dumbest.

server/subscriber/MediaRequestSubscriber.ts
```diff
@@ -768,26 +769,28 @@ export class MediaRequestSubscriber implements EntitySubscriberInterface<MediaRe
...
-      media[entity.is4k ? 'status4k' : 'status'] !== MediaStatus.PROCESSING
+      media[statusKey] !== MediaStatus.AVAILABLE &&
+      media[statusKey] !== MediaStatus.PARTIALLY_AVAILABLE &&
+      media[statusKey] !== MediaStatus.PROCESSING
     ) {
-      media[entity.is4k ? 'status4k' : 'status'] = MediaStatus.PROCESSING;
-      mediaRepository.save(media);
+      media[statusKey] = MediaStatus.PROCESSING;
+      await mediaRepository.save(media);
     }
```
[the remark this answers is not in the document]
fallenbagel: stupid.

server/subscriber/MediaRequestSubscriber.ts
```diff
@@ -830,54 +846,74 @@ export class MediaRequestSubscriber implements EntitySubscriberInterface<MediaRe
...
 
-    this.sendToRadarr(event.entity as MediaRequest);
-    this.sendToSonarr(event.entity as MediaRequest);
-
-    this.updateParentStatus(event.entity as MediaRequest);
+    try {
+      await this.sendToRadarr(event.entity as MediaRequest);
+      await this.sendToSonarr(event.entity as MediaRequest);
+      await this.updateParentStatus(event.entity as MediaRequest);
```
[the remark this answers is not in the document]
fallenbagel: stupid

server/subscriber/MediaRequestSubscriber.ts
```diff
@@ -830,54 +846,74 @@ export class MediaRequestSubscriber implements EntitySubscriberInterface<MediaRe
...
-      }
-      if (event.entity.media.mediaType === MediaType.TV) {
-        this.notifyAvailableSeries(event.entity as MediaRequest, event);
+      if (event.entity.status === MediaRequestStatus.COMPLETED) {
+        if (event.entity.media.mediaType === MediaType.MOVIE) {
+          await this.notifyAvailableMovie(event.entity as MediaRequest, event);
+        }
+        if (event.entity.media.mediaType === MediaType.TV) {
+          await this.notifyAvailableSeries(event.entity as MediaRequest, event);
+        }
```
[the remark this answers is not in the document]
fallenbagel: stupid

server/subscriber/MediaRequestSubscriber.ts
```diff
@@ -830,54 +846,74 @@ export class MediaRequestSubscriber implements EntitySubscriberInterface<MediaRe
...
     }
 
-    this.sendToRadarr(event.entity as MediaRequest);
-    this.sendToSonarr(event.entity as MediaRequest);
-
-    this.updateParentStatus(event.entity as MediaRequest);
+    try {
+      await this.sendToRadarr(event.entity as MediaRequest);
+      await this.sendToSonarr(event.entity as MediaRequest);
+      await this.updateParentStatus(event.entity as MediaRequest);
```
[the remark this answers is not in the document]
fallenbagel: same as before. Plain stupid

server/subscriber/MediaRequestSubscriber.ts
```diff
@@ -397,10 +398,23 @@ export class MediaRequestSubscriber implements EntitySubscriberInterface<MediaRe
             await mediaRepository.save(media);
           })
           .catch(async () => {
-            const requestRepository = getRepository(MediaRequest);
+            try {
+              const requestRepository = getRepository(MediaRequest);
 
-            entity.status = MediaRequestStatus.FAILED;
-            requestRepository.save(entity);
+              if (entity.status !== MediaRequestStatus.FAILED) {
```
gauthier-th, line 404: You should probably fetch this entity again before doing this? The entity is retrieved earlier and the values could have changed.

server/subscriber/MediaRequestSubscriber.ts
```diff
@@ -707,10 +719,23 @@ export class MediaRequestSubscriber implements EntitySubscriberInterface<MediaRe
...
```
gauthier-th, line 725: Same here

server/subscriber/MediaRequestSubscriber.ts
```diff
@@ -830,54 +915,74 @@ export class MediaRequestSubscriber implements EntitySubscriberInterface<MediaRe
...
+      await this.sendToSonarr(event.entity as MediaRequest);
+      await this.updateParentStatus(event.entity as MediaRequest);
 
-    if (event.entity.status === MediaRequestStatus.COMPLETED) {
-      if (event.entity.media.mediaType === MediaType.MOVIE) {
-        this.notifyAvailableMovie(event.entity as MediaRequest, event);
-      }
-      if (event.entity.media.mediaType === MediaType.TV) {
-        this.notifyAvailableSeries(event.entity as MediaRequest, event);
+      if (event.entity.status === MediaRequestStatus.COMPLETED) {
```
gauthier-th, line 952: Same here, shouldn't you fetch again the entity?

server/subscriber/MediaRequestSubscriber.ts
```diff
@@ -799,14 +826,72 @@ export class MediaRequestSubscriber implements EntitySubscriberInterface<MediaRe
...
+            .andWhere('season.seasonNumber = :seasonNumber', {
+              seasonNumber: season.seasonNumber,
+            })
+            .getCount();
+
+          if (otherActiveRequests === 0) {
+            season[statusKey] = MediaStatus.UNKNOWN;
+            await seasonRepository.save(season);
+          }
+        }
```
gauthier-th, line 892: Doesn't this belongs to another PR?

fallenbagel: Since its related to updateParentStatus function thats fixed in this PR, I thought of putting it together

server/subscriber/MediaRequestSubscriber.ts
fallenbagel: No we don't need. This is just a safeguard added. The status check is just a guard to prevent a redundant save. If something else already marked it failed, we skip. If not, we set it. Re-fetching adds a DB round trip for no real benefit, worst case we save FAILED twice, which is harmless.

server/subscriber/MediaRequestSubscriber.ts
fallenbagel: Same as above

server/subscriber/MediaRequestSubscriber.ts
fallenbagel: `event.entity` is literally the entity that TypeORM just persisted/saved. It's at It's freshest state possible. It is what triggered this subscriber. There's literally nothing newer to fetch.

server/subscriber/MediaRequestSubscriber.ts
gauthier-th: I get it, but that's a safeguard that is done on an entity whose values may have changed between the initial fetch and this check, since there are some async operation in-between.
What if the status of this entity has been changed somewhere else during the async operation?

fallenbagel: Even if we re-fetch before the check, there's always a gap between reading the value and acting on it since another operation could change the status between the fetch and the save anyways. Re-fetching just shrinks that window, it doesn't eliminate the race.

But more importantly, this guard isn't even strictly necessary. The original code just set `entity.status = FAILED` and saved unconditionally without any check which was more than fine and isn't related to why we had mediaId nullification occuring. The guard was only added as a minor optimization to avoid a redundant save if something else already marked it FAILED. Either way, this `.catch()` only runs when `.addMovie()` or `.addSeries()` fails. Meaning the external service rejected the request. At that point nothing else is going to be changing this entity's status, so the concern doesn't apply in practice, and marking it as FAILED regardless is the correct outcome.

gauthier-th: > Re-fetching just shrinks that window, it doesn't eliminate the race.

Sure, but it makes it way smaller.

> But more importantly, this guard isn't even strictly necessary. The original code just set `entity.status = FAILED` and saved unconditionally without any check which was more than fine and isn't related to why we had mediaId nullification occuring. The guard was only added as a minor optimization to avoid a redundant save if something else already marked it FAILED. Either way, this `.catch()` only runs when `.addMovie()` or `.addSeries()` fails. Meaning the external service rejected the request. At that point nothing else is going to be changing this entity's status, so the concern doesn't apply in practice, and marking it as FAILED regardless is the correct outcome.

Okay, makes more sense now. I was just asking to be sure that this couldn't introduce another issue.

server/subscriber/MediaRequestSubscriber.ts
gauthier-th: Yes, my bad

server/subscriber/MediaRequestSubscriber.ts
```diff
@@ -799,14 +826,72 @@ export class MediaRequestSubscriber implements EntitySubscriberInterface<MediaRe
...
-      media.requests.filter(
-        (request) => request.status === MediaRequestStatus.PENDING
-      ).length === 0 &&
-      media[entity.is4k ? 'status4k' : 'status'] === MediaStatus.PENDING &&
-      media[entity.is4k ? 'status4k' : 'status'] !== MediaStatus.DELETED
+      media[statusKey] === MediaStatus.PENDING
     ) {
-      media[entity.is4k ? 'status4k' : 'status'] = MediaStatus.UNKNOWN;
-      mediaRepository.save(media);
+      const requestRepository = getRepository(MediaRequest);
```
M0NsTeRRR: We can't reuse `requestRepository`  from line 798 ?

fallenbagel: Yes. We need a fresh instance here incase of state changes

gauthier-th: Retrieving the TypeORM repository again is different from fetching again an entity?

M0NsTeRRR (approved):

server/subscriber/MediaRequestSubscriber.ts
fallenbagel: > Retrieving the TypeORM repository again is different from fetching again an entity?

Actually no i just re-read what I did. This is a duplicate line leftover from my previously attempts. I moved it to the top and forgot to remove it

M0NsTeRRR (approved):

PR #2396, opened 2026-02-09
perf: add database indexes & minor frontend/backend improvements

fallenbagel:
## Description

This PR addresses performance bottlenecks in database queries and makes several small improvements across the codebase. 

The main changes this PR adds is the indexes to columns that are heavily filtered in the media request workflow such as status on both `MediaRequest` and `Media`, `status4k` on `Media`, and a composite index on `tmdbId+mediaType` which is used in nearly every media lookup. Migrations are included for both SQLite and PostgreSQL.

On the frontend, the discover hook now uses a 30-second deduping interval and disables revalidation on focus since TMDB results don't change frequently enough to justify refetching on every tab switch. 

The remaining changes are minor cleanup: fixing an any cast in the IP middleware, adding a missing `useEffect` dependency in the web push settings component, tree-shaking the `lodash` import in the Selector component, bumping the Next.js large page data threshold to 512 KB to suppress misleading warnings for TV pages with many seasons or /login with many translations loaded at once, and removing the unused `dayjs` dependency.

## How Has This Been Tested?
- `pnpm typecheck` was ran
- `pnpm lint` was ran
- `pnpm build` was ran successfully
- Tested a normal day-to-day pathway (works no regression)

## Checklist:
- [X] Database migration (if required)

### DIFF
--- a/server/entity/Media.ts
+++ b/server/entity/Media.ts
@@ -26,6 +26,7 @@ import { MediaRequest } from './MediaRequest';
 import Season from './Season';
 
 @Entity()
+@Index(['tmdbId', 'mediaType'])
 class Media {
   public static async getRelatedMedia(
     user: User | undefined,
@@ -101,9 +102,11 @@ class Media {
   public imdbId?: string;
 
   @Column({ type: 'int', default: MediaStatus.UNKNOWN })
+  @Index()
   public status: MediaStatus;
 
   @Column({ type: 'int', default: MediaStatus.UNKNOWN })
+  @Index()
   public status4k: MediaStatus;
 
   @OneToMany(() => MediaRequest, (request) => request.media, {
--- a/server/index.ts
+++ b/server/index.ts
@@ -165,12 +165,15 @@ app
       try {
         const descriptor = Object.getOwnPropertyDescriptor(req, 'ip');
         if (descriptor?.writable === true) {
-          (req as any).ip = getClientIp(req) ?? '';
+          Object.defineProperty(req, 'ip', {
+            ...descriptor,
+            value: getClientIp(req) ?? '',
+          });
         }
       } catch (e) {
         logger.error('Failed to attach the ip to the request', {
           label: 'Middleware',
-          message: e.message,
+          message: (e as Error).message,
         });
       } finally {
         next();
--- a/server/migration/postgres/1770627987304-AddPerformanceIndexes.ts
+++ b/server/migration/postgres/1770627987304-AddPerformanceIndexes.ts
@@ -1,0 +1,35 @@
+import type { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddPerformanceIndexes1770627987304 implements MigrationInterface {
+  name = 'AddPerformanceIndexes1770627987304';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE INDEX "IDX_4c696e8ed36ae34fe18abe59d2" ON "media_request" ("status") `
+    );
+    await queryRunner.query(
+      `CREATE INDEX "IDX_c730c2d67f271a372c39a07b7e" ON "media" ("status") `
+    );
+    await queryRunner.query(
+      `CREATE INDEX "IDX_5d6218de4f547909391a5c1347" ON "media" ("status4k") `
+    );
+    await queryRunner.query(
+      `CREATE INDEX "IDX_f8233358694d1677a67899b90a" ON "media" ("tmdbId", "mediaType") `
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `DROP INDEX "public"."IDX_f8233358694d1677a67899b90a"`
+    );
+    await queryRunner.query(
+      `DROP INDEX "public"."IDX_5d6218de4f547909391a5c1347"`
+    );
+    await queryRunner.query(
+      `DROP INDEX "public"."IDX_c730c2d67f271a372c39a07b7e"`
+    );
+    await queryRunner.query(
+      `DROP INDEX "public"."IDX_4c696e8ed36ae34fe18abe59d2"`
+    );
+  }
+}
--- a/server/migration/sqlite/1770627968781-AddPerformanceIndexes.ts
+++ b/server/migration/sqlite/1770627968781-AddPerformanceIndexes.ts
@@ -1,0 +1,339 @@
+import type { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddPerformanceIndexes1770627968781 implements MigrationInterface {
+  name = 'AddPerformanceIndexes1770627968781';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`DROP INDEX "UQ_6427d07d9a171a3a1ab87480005"`);
+    await queryRunner.query(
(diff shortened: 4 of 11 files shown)


server/migration/sqlite/1770627968781-AddPerformanceIndexes.ts
```diff
@@ -0,0 +1,339 @@
...
+    );
+    await queryRunner.query(`DROP TABLE "user_push_subscription"`);
+    await queryRunner.query(
+      `ALTER TABLE "temporary_user_push_subscription" RENAME TO "user_push_subscription"`
+    );
+    await queryRunner.query(`DROP INDEX "IDX_939f205946256cc0d2a1ac51a8"`);
+    await queryRunner.query(
+      `CREATE TABLE "temporary_watchlist" ("id" integer PRIMARY KEY AUTOINCREMENT NOT NULL, "ratingKey" varchar NOT NULL, "mediaType" varchar NOT NULL, "title" varchar NOT NULL, "tmdbId" integer NOT NULL, "createdAt" datetime NOT NULL DEFAULT (CURRENT_TIMESTAMP), "updatedAt" datetime NOT NULL DEFAULT (CURRENT_TIMESTAMP), "requestedById" integer, "mediaId" integer, CONSTRAINT "UNIQUE_USER_DB" UNIQUE ("tmdbId", "requestedById"), CONSTRAINT "FK_6641da8d831b93dfcb429f8b8bc" FOREIGN KEY ("mediaId") REFERENCES "media" ("id") ON DELETE CASCADE ON UPDATE NO ACTION, CONSTRAINT "FK_ae34e6b153a90672eb9dc4857d7" FOREIGN KEY ("requestedById") REFERENCES "user" ("id") ON DELETE CASCADE ON UPDATE NO ACTION)`
+    );
```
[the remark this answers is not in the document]
fallenbagel: stupid.

server/migration/postgres/1770627987304-AddPerformanceIndexes.ts
```diff
@@ -0,0 +1,35 @@
...
+export class AddPerformanceIndexes1770627987304 implements MigrationInterface {
+  name = 'AddPerformanceIndexes1770627987304';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE INDEX "IDX_4c696e8ed36ae34fe18abe59d2" ON "media_request" ("status") `
+    );
+    await queryRunner.query(
+      `CREATE INDEX "IDX_c730c2d67f271a372c39a07b7e" ON "media" ("status") `
+    );
```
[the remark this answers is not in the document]
fallenbagel: stupid.

server/index.ts
```diff
@@ -165,12 +165,15 @@ app
       try {
         const descriptor = Object.getOwnPropertyDescriptor(req, 'ip');
         if (descriptor?.writable === true) {
-          (req as any).ip = getClientIp(req) ?? '';
+          Object.defineProperty(req, 'ip', {
+            value: getClientIp(req) ?? '',
+            writable: true,
+          });
```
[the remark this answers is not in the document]
fallenbagel: stupid.

server/migration/sqlite/1770627968781-AddPerformanceIndexes.ts
```diff
@@ -0,0 +1,339 @@
...
+    await queryRunner.query(
+      `CREATE TABLE "temporary_user_push_subscription" ("id" integer PRIMARY KEY AUTOINCREMENT NOT NULL, "endpoint" varchar NOT NULL, "p256dh" varchar NOT NULL, "auth" varchar NOT NULL, "userId" integer, "userAgent" varchar, "createdAt" datetime DEFAULT (datetime('now')), CONSTRAINT "UQ_f90ab5a4ed54905a4bb51a7148b" UNIQUE ("auth"), CONSTRAINT "FK_03f7958328e311761b0de675fbe" FOREIGN KEY ("userId") REFERENCES "user" ("id") ON DELETE CASCADE ON UPDATE NO ACTION)`
+    );
+    await queryRunner.query(
+      `INSERT INTO "temporary_user_push_subscription"("id", "endpoint", "p256dh", "auth", "userId", "userAgent", "createdAt") SELECT "id", "endpoint", "p256dh", "auth", "userId", "userAgent", "createdAt" FROM "user_push_subscription"`
+    );
+    await queryRunner.query(`DROP TABLE "user_push_subscription"`);
+    await queryRunner.query(
+      `ALTER TABLE "temporary_user_push_subscription" RENAME TO "user_push_subscription"`
+    );
```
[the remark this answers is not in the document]
fallenbagel: stupid.

server/entity/Media.ts
```diff
@@ -101,9 +102,11 @@ class Media {
   public imdbId?: string;
 
   @Column({ type: 'int', default: MediaStatus.UNKNOWN })
+  @Index()
```
M0NsTeRRR, line 105: It would be better to use https://typeorm.io/docs/entity/entities/#enum-column-type.
The documentation appears incorrect regarding https://github.com/typeorm/typeorm/pull/3700, as SQLite is handled using `CHECK`. Could you test this?
If it works, we could update the typeorm documentation.

fallenbagel: That wouldn't work. The reason we use `int` is intentional as these are numerical enums that are stored and compared as integers throughput the entire codebase. TypeORM's enum column type stores values as strings, so switching would require a data migration for every existing row across both SQLite and Postgres. Postgres native ENUMs are also painful to modify later afaik (using ALTER TYPE). Also integer columns index more efficiently too, which is what this PR is about

gauthier-th: > The documentation appears incorrect regarding https://github.com/typeorm/typeorm/pull/3700, as SQLite is handled using CHECK.

I don't think it's incorrect, just missing for `simple-enum`. `simple-enum` is a workaround for DB when native `enum` are not supported (like SQLite).

Also `simple-enum` is storing a string, not an int, so it would require a migration of all enum values. I find it better like it is right now, since we already handle this with the value of our enum (which is an int).

fallenbagel: > I find it better like it is right now, since we already handle this with the value of our enum (which is an int).

Also int is way faster. Better perf. Especially when indexed for frequently queried ones like media status

M0NsTeRRR: Okay if this requires a major refactor 😅
It’s odd that we doesn’t serialize directly to the enum instead of mixing enums and integers.

> Postgres native ENUMs are also painful to modify later afaik (using ALTER TYPE)

The ORM is here for that :)


> Also integer columns index more efficiently too, which is what this PR is about

You can create index on enum column

> I don't think it's incorrect, just missing for `simple-enum`. `simple-enum` is a workaround for DB when native `enum` are not supported (like SQLite).

Ah yeah the doc is missing `simple-enum` not an update on this part :)

fallenbagel: > Okay if this requires a major refactor 😅
> It’s odd that we doesn’t serialize directly to the enum instead of mixing enums and integers.

Probably because when overseerr was created that was not supported by typeorm. Similar to how we are using expressjs instead of full stack nextjs.

> > Postgres native ENUMs are also painful to modify later afaik (using ALTER TYPE)
> 
> The ORM is here for that :)

I read that some people faced issues. We are already fighting with typeorm because it keeps improperly handling and many issues with its lifecycle hooks (actual bugs) that we have to workaround. I dont wanna fight typeorm for enums 😅

> 
> > Also integer columns index more efficiently too, which is what this PR is about
> 
> You can create index on enum column

Yeah. But afaik int is still faster. I could be wrong but this comes from my knowledge base I learnt back in 2021. Things might have changed now 😅

M0NsTeRRR: > Yeah. But afaik int is still faster. I could be wrong but this comes from my knowledge base I learnt back in 2021. Things might have changed now 😅

Enum use integer under the hood

fallenbagel: Ah even if saved as strings in the db?

gauthier-th (approved):

server/entity/Media.ts
M0NsTeRRR: PostgreSQL uses `pg_enum`, which relies on integer ordering and It takes 4 bytes on disk to store the enum.

PR #2467, opened 2026-02-16
chore(github): add docs and maintenance issue templates

fallenbagel:
## Description

Add Documentation and Maintenance/Chore issue forms for better triage and non-feature work tracking. Issue type has also been added.

## Checklist:

## Summary by CodeRabbit

* **Chores**
  * Updated issue templates: standardized labels and added explicit type fields for bug and enhancement reports.
  * Added new templates for documentation and maintenance tasks with structured form fields, validation, required checks, and guidance to capture location, area, problem, suggested fixes, and related links.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -1,6 +1,6 @@
 name: 🐛 Bug Report
 description: Report a problem
-labels: ['bug', 'awaiting triage']
+labels: ['awaiting triage']
 type: bug
 body:
   - type: markdown
--- a/.github/ISSUE_TEMPLATE/enhancement.yml
+++ b/.github/ISSUE_TEMPLATE/enhancement.yml
@@ -1,6 +1,6 @@
 name: ✨ Feature Request
 description: Suggest an idea
-labels: ['enhancement', 'awaiting triage']
+labels: ['awaiting triage']
 type: feature
 body:
   - type: markdown


.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -1,6 +1,7 @@
 name: 🐛 Bug Report
 description: Report a problem
 labels: ['bug', 'awaiting triage']
```
M0NsTeRRR: do we keep bug labels if we do that ?

.github/ISSUE_TEMPLATE/enhancement.yml
```diff
@@ -1,6 +1,7 @@
 name: ✨ Feature Request
 description: Suggest an idea
 labels: ['enhancement', 'awaiting triage']
```
M0NsTeRRR: same here for enhancement

.github/ISSUE_TEMPLATE/bug.yml
fallenbagel: Yes. Having both labels and type makes it easier to sort

.github/ISSUE_TEMPLATE/enhancement.yml
fallenbagel: Same as above. In my opinion

.github/ISSUE_TEMPLATE/bug.yml
M0NsTeRRR: It seems redundant to me. If we need to move a bug that is actually an enhancement, we need to do two things I don’t see the value in specifying it twice :thinking:

gauthier-th: I tend to agree. Let's keep only one type of category? IMO no matter which one.

PR #2473, opened 2026-02-17
fix(imageproxy): restrict to relative paths to prevent baseURL override

RinZ27:
## Description

- Problem: The `/imageproxy` route could be tricked into fetching absolute URLs, bypassing the intended `baseURL`.
- Why it matters: This allowed for potential SSRF where internal network resources or server-side metadata could be accessed through the proxy.
- What changed: Added a check to ensure `imagePath` is a relative path starting with `/` and does not contain protocol strings or double slashes.

## How Has This Been Tested?

- Verified that standard TMDB/TVDB paths (e.g., `/t/p/xxx.jpg`) still work.
- Verified that absolute URLs and double-slash bypasses are blocked.

## Checklist:

- [x] Disclosed any use of AI.
- [x] All new and existing tests passed. 🦞

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced input validation for the image proxy route to reject invalid image paths and return appropriate error responses. This prevents potential issues from malformed requests, improving system reliability and stability.

### DIFF
--- a/server/routes/imageproxy.ts
+++ b/server/routes/imageproxy.ts
@@ -32,6 +32,18 @@ function initTvdbImageProxy() {
 
 router.get('/:type/*', async (req, res) => {
   const imagePath = req.path.replace(/^\/\w+/, '');
+
+  if (
+    !imagePath.startsWith('/') ||
+    imagePath.startsWith('//') ||
+    imagePath.includes('://')
+  ) {
+    logger.error('Invalid image path detected', {
+      imagePath: imagePath.slice(0, 200),
+    });
+    return res.status(400).send('Invalid image path');
+  }
+
   try {
     let imageData;
     if (req.params.type === 'tmdb') {


gauthier-th:
I don't think I can reproduce this issue. When I try to inject an absolute URL to the image proxy, the URL is always prefixed with `/`. Could you please provide an example?

RinZ27:
Thanks for checking this. The bypass actually happens when using protocol-relative URLs (e.g., `//domain.com`). 

If I send a request like `GET /imageproxy/tmdb//google.com`, the `imagePath` variable becomes `//google.com` after the regex replacement. Since `axios` is configured with a `baseURL` in the `ImageProxy` class, passing a path that starts with `//` causes `axios` to treat it as an absolute URL and ignore the `baseURL` entirely. I noticed that while some environments might normalize the path, a raw request can often bypass this if it's not explicitly handled in the code.

You should be able to reproduce it using:
`curl -v "http://localhost:3000/imageproxy/tmdb//google.com"`

Without the fix, this would trigger a request to `https://google.com` instead of appending the path to the TMDB base URL.

@gauthier-th

fallenbagel:
> Thanks for checking this. The bypass actually happens when using protocol-relative URLs (e.g., `//domain.com`). 
> 
> If I send a request like `GET /imageproxy/tmdb//google.com`, the `imagePath` variable becomes `//google.com` after the regex replacement. Since `axios` is configured with a `baseURL` in the `ImageProxy` class, passing a path that starts with `//` causes `axios` to treat it as an absolute URL and ignore the `baseURL` entirely. I noticed that while some environments might normalize the path, a raw request can often bypass this if it's not explicitly handled in the code.
> 
> You should be able to reproduce it using:
> `curl -v "http://localhost:3000/imageproxy/tmdb//google.com"`
> 
> Without the fix, this would trigger a request to `https://google.com` instead of appending the path to the TMDB base URL.
> 
> @gauthier-th

I just tested exactly this. Returns nothing. Not with curl. Not with browser eithe. Not even with `curl -L`

RinZ27:
@fallenbagel That 'nothing' response (likely a 500 status with a blank body) actually confirms the bypass is happening. When `axios` sees a path starting with `//`, it treats it as a protocol-relative URL and ignores the `baseURL` entirely. Since it's trying to fetch an image but getting HTML from Google, or the request is failing, `getImage()` throws an error and the route's catch block sends back a 500.

Checking the server logs should show `axios` attempting to hit the external domain instead of TMDB. The fix ensures we only allow single-slash relative paths, preventing this from being used to probe internal infrastructure or cloud metadata endpoints.

fallenbagel:
> Checking the server logs should show `axios` attempting to hit the external domain instead of TMDB. The fix ensures we only allow single-slash relative paths, preventing this from being used to probe internal infrastructure or cloud metadata endpoints.

I'll let our security manager take a look at this as well @M0NsTeRRR 

But for future reference, please report security related things in accordance to:
https://github.com/seerr-team/seerr/blob/develop/SECURITY.md

RinZ27:
@fallenbagel understood, and thanks for the heads-up. I'll be sure to follow the `SECURITY.md` flow for future findings. Just let me know if @M0NsTeRRR needs more info.

0xSysR3ll:
@RinZ27
How did you come up with this ? Are you a security analyst ?

RinZ27:
@0xSysR3ll, I just have a keen interest in secure coding practices and enjoy digging into how different components, like Axios in this case, handle edge-case inputs. It's more of a hobbyist's curiosity combined with a bit of security auditing on projects I find interesting!

M0NsTeRRR:
Hello,

This path is not directly used on the host, but it relies on the hash signature in the cache, so I don’t see any issues with it. The Seerr frontend never uses this type of path, meaning it would only be relevant if an attacker attempted to exploit it. The only advantage of blocking it would be a minor performance optimization returning a 400 directly avoids executing the entire code. However, an attacker could simply use a regular path, so this doesn’t meaningfully improve performance during an attack.

I’m locking this issue because it’s not the appropriate way to report and discuss a security concern. If you believe my response is incorrect, please open a security report as described in `SECURITY.md` to avoid discussing such topics publicly.

PR #2531, opened 2026-02-20
fix(proxy): add path validation guardrail to imageproxy

RinZ27:
## Description

This PR adds a path validation guardrail to the `imageproxy` route. It prevents the processing of protocol-relative URLs (starting with `//`) or absolute URLs (containing `://`). 

While the underlying `ImageProxy` logic uses a `baseURL`, this additional check ensures that malformed or malicious paths that might attempt to bypass the intended destination are blocked early at the routing level.

## How Has This Been Tested?

Verified that standard image proxying for TMDB and TVDB remains functional. Confirmed that requests with absolute or protocol-relative paths in the `imagePath` parameter are now correctly caught and return a `403 Forbidden` response, with a corresponding error logged.

## Checklist:

- [x] Disclosed any use of AI: This PR was prepared with the assistance of an AI agent.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened image proxy security by adding validation to detect and block potentially malicious image requests that try to reference external or malformed resources. Such requests are now rejected with an appropriate forbidden response, preventing unintended access and improving overall safety.

### DIFF
--- a/server/routes/imageproxy.ts
+++ b/server/routes/imageproxy.ts
@@ -32,6 +32,12 @@ function initTvdbImageProxy() {
 
 router.get('/:type/*', async (req, res) => {
   const imagePath = req.path.replace(/^\/\w+/, '');
+
+  if (imagePath.startsWith('//') || imagePath.includes('://')) {
+    logger.error('Invalid URL for image proxy', { imagePath });
+    return res.status(403).send('Invalid URL for image proxy');
+  }
+
   try {
     let imageData;
     if (req.params.type === 'tmdb') {


server/routes/imageproxy.ts
```diff
@@ -32,7 +32,14 @@ function initTvdbImageProxy() {
 
 router.get('/:type/*', async (req, res) => {
   const imagePath = req.path.replace(/^\/\w+/, '');
+
+  if (imagePath.startsWith('//') || imagePath.includes('://')) {
+    logger.error('Potential SSRF attempt blocked', { imagePath });
```
gauthier-th: I'd prefer an error message like "Invalid URL for image proxy"

RinZ27: Updated the error message to "Invalid URL for image proxy" and cleaned up the blank line in the try block. @gauthier-th

gauthier-th: I don't see the change. Did you pushed it correctly?

server/routes/imageproxy.ts
```diff
@@ -32,6 +32,12 @@ function initTvdbImageProxy() {
...
```
gauthier-th:
```suggestion
    logger.error('Invalid URL for image proxy', { imagePath });
```

gauthier-th (changes requested):

RinZ27:
I actually kept it as "Potential SSRF" during the initial pass to explicitly flag the security intent for the review process, otherwise I would've stuck with the functional naming from the start. Glad we're on the same page that "Invalid URL" is better for the production logs. Updated!

gauthier-th (approved):

fallenbagel (approved):

PR #2605, opened 2026-02-28
fix(jellyfin-scanner): add TheMovieDb provider fallback for Jellyfin scanner

YakGravity:
## Description

Shokofin (a Jellyfin plugin for anime managed by Shoko Server) writes TMDB IDs using the legacy Jellyfin provider key `TheMovieDb` instead of `Tmdb`. This causes Seerr to miss all anime managed by Shokofin during library sync, showing them as unavailable despite being present in Jellyfin.

This PR adds a fallback in three places in the Jellyfin scanner:

1. **Movie scanner** (`extractMovieIds`): `metadata.ProviderIds.Tmdb || metadata.ProviderIds.TheMovieDb || null`
2. **TV show scanner** (`processJellyfinShow`): Added `|| metadata.ProviderIds.TheMovieDb` to the if-condition
3. **TV show scanner** (`processJellyfinShow`): `Number(metadata.ProviderIds.Tmdb || metadata.ProviderIds.TheMovieDb)` for the tmdbId lookup

No breaking changes: When `Tmdb` exists (standard Sonarr/Radarr), `||` short-circuits and behavior is unchanged. The fallback only activates for items using the legacy `TheMovieDb` key.

## How Has This Been Tested?

Tested on a live Jellyfin + Shokofin setup with multiple anime titles. Before the fix, all Shokofin-managed anime showed as unavailable in Seerr. After the fix, library sync correctly identifies them (e.g. "Frieren: Beyond Journey's End" returns `mediaInfo.status: 4` PARTIALLY_AVAILABLE). Non-anime content (Sonarr/Radarr managed) continues to work without regression.

N/A

## Checklist:

## AI Disclosure

The root cause analysis (Shokofin using `TheMovieDb` vs `Tmdb` provider key) was identified through manual debugging with log analysis. Claude was consulted for reviewing the fix approach and drafting the PR description. The code changes were authored and tested manually.

## Summary by CodeRabbit

* **Bug Fixes**
  * Jellyfin integration now recognizes alternate TMDb identifier keys for movies and TV shows, improving ID derivation and metadata matching.
  * Library scans are more reliable when metadata uses different TMDb provider fields, reducing missed or incorrect media matches.


fallenbagel:
Please update your pr description with the proper pr template. You're missing several things from the checklist.

fallenbagel (dismissed):

0xSysR3ll:
You forgot to run `pnpm format` before committing.

gauthier-th (changes requested):
Thanks for the contribution! While this addresses the sync issue with Shokofin, after looking into the Jellyfin source code (specifically ProviderIdsExtensions.cs and the MetadataProvider enum), it seems this "legacy" key is actually a side effect of how certain plugins or Jellyfin itself handle provider mapping.

We'd like to avoid adding specific fallbacks for external Jellyfin plugin inconsistencies. This issue should probably be fixed by Shokofin?

YakGravity:
> Thanks for the contribution! While this addresses the sync issue with Shokofin, after looking into the Jellyfin source code (specifically ProviderIdsExtensions.cs and the MetadataProvider enum), it seems this "legacy" key is actually a side effect of how certain plugins or Jellyfin itself handle provider mapping.
> 
> We'd like to avoid adding specific fallbacks for external Jellyfin plugin inconsistencies. This issue should probably be fixed by Shokofin?

Hi gauthier-th. Thanks for taking a look! I did some digging and I think this is actually a Jellyfin issue, not a Shokofin one.

`TheMovieDb` is defined as the official provider name directly in Jellyfin's own `TmdbUtils.cs` (`ProviderName = "TheMovieDb"`). The problem is that Jellyfin has two ways to set provider IDs. One via the enum (`MetadataProvider.Tmdb` which writes the key `Tmdb`) and one via the string based method (which uses `TheMovieDb`). So both keys can legitimately show up, depending on which path a plugin or Jellyfin itself takes.

Shokofin just happens to use the string based name, but any plugin could end up doing the same thing. Waiting for every plugin to fix this doesn't really solve the problem since the inconsistency lives in Jellyfin core.

Also worth noting: #2537 already added a similar fallback (TVDb to TMDB) for the Plex scanner, so there's precedent for this kind of defensive handling in Seerr.

Happy to discuss if you see it differently.

gauthier-th:
> > Thanks for the contribution! While this addresses the sync issue with Shokofin, after looking into the Jellyfin source code (specifically ProviderIdsExtensions.cs and the MetadataProvider enum), it seems this "legacy" key is actually a side effect of how certain plugins or Jellyfin itself handle provider mapping.
> > We'd like to avoid adding specific fallbacks for external Jellyfin plugin inconsistencies. This issue should probably be fixed by Shokofin?
> 
> Hi gauthier-th. Thanks for taking a look! I did some digging and I think this is actually a Jellyfin issue, not a Shokofin one.
> 
> `TheMovieDb` is defined as the official provider name directly in Jellyfin's own `TmdbUtils.cs` (`ProviderName = "TheMovieDb"`). The problem is that Jellyfin has two ways to set provider IDs. One via the enum (`MetadataProvider.Tmdb` which writes the key `Tmdb`) and one via the string based method (which uses `TheMovieDb`). So both keys can legitimately show up, depending on which path a plugin or Jellyfin itself takes.
> 
> Shokofin just happens to use the string based name, but any plugin could end up doing the same thing. Waiting for every plugin to fix this doesn't really solve the problem since the inconsistency lives in Jellyfin core.
> 
> Also worth noting: #2537 already added a similar fallback (TVDb to TMDB) for the Plex scanner, so there's precedent for this kind of defensive handling in Seerr.
> 
> Happy to discuss if you see it differently.

Thanks for the explanation.
In an ideal situation we wouldn't have to implement this, but given the small size of the changes and because it does not modify any logic in the app but only adds a little fallback, I'm OK with merging this.
CC @fallenbagel

gauthier-th (approved):

fallenbagel (approved):

PR #2502, opened 2026-02-18
feat: add script for SQLite to PostgreSQL migration

gauthier-th:
## Description

This PR creates a migration script to migrate Seerr from SQLite to PostgreSQL.

It uses a TypeORM instance with SQLite to load the existing schemas and data of the app, and another TypeORM instance connected to PostgreSQL to sync the schemas and upload the data. SQLite/PostgreSQL database configuration works the same way as described in [the docs](https://docs.seerr.dev/extending-seerr/database-config).

Because TypeORM methods `.save()` and `.update()` don't support manually setting the Primary Key, the `sequence` value of the table is set before inserting the data, so that the right value is used.

## How Has This Been Tested?

Create a PostgreSQL instance with Docker:
```sh
sudo docker run --name postgres-seerr -e POSTGRES_PASSWORD=postgres -d -p 127.0.0.1:5432:5432/tcp postgres:latest
```
Recreate an empty database:
```sh
PGPASSWORD=postgres sudo docker exec -it postgres-seerr /usr/bin/psql -h 127.0.0.1 -U postgres -c "DROP DATABASE IF EXISTS seerr;" -c "CREATE DATABASE seerr;"
```
Migrate the current SQLite database to PostgreSQL:
```sh
DB_USER=postgres DB_PASS=postgres pnpm db:migratetopostgres
```
Check that your PostgreSQL database is successfully cloned from your SQLite database.

## Checklist:

- [x] I have read and followed the contribution [guidelines](https://github.com/seerr-team/seerr/blob/develop/CONTRIBUTING.md).
- [x] Disclosed any use of AI (see our [policy](https://github.com/seerr-team/seerr/blob/develop/CONTRIBUTING.md#ai-assistance-notice))
- [ ] I have updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [ ] Successful build `pnpm build`
- [ ] Translation keys `pnpm i18n:extract`
- [ ] Database migration (if required)

PR #2627, opened 2026-03-04
feat(helm): Support sqlite to postgresql migration

damnmersault:
## Description

Adds explicit options for configuring a postgresql database as well migrating an existing sqlite database via initContainers.
There's also a trivial cleanup for determining the PVC name.

## How Has This Been Tested?

Tested locally with both sqlite database (default) and postgresql database with data migration.

## Checklist:

`pnpm` steps skipped as not applicable.

## AI Disclosure

Claude assisted with writing the bash scripts in `templates/configmap-migration.yaml` and adding the initContainers in `templates/statefulset.yaml`. The overall approach to migration was directed by myself, and all generated code was reviewed.

## Summary by CodeRabbit

* **New Features**
  * Added PostgreSQL database support with comprehensive configuration options including credentials, network settings, socket paths, and SSL certificates.
  * Added automated SQLite-to-PostgreSQL migration capability.
  * Added support for existing PVC claims in persistence configuration.

* **Documentation**
  * Updated configuration documentation with PostgreSQL parameters.

* **Chores**
  * Bumped Helm chart version to 3.4.0.

### DIFF
--- a/charts/seerr-chart/Chart.yaml
+++ b/charts/seerr-chart/Chart.yaml
@@ -3,7 +3,7 @@ kubeVersion: '>=1.23.0-0'
 name: seerr-chart
 description: Seerr helm chart for Kubernetes
 type: application
-version: 3.3.0
+version: 3.4.0
 # renovate: image=ghcr.io/seerr-team/seerr
 appVersion: 'v3.1.0'
 maintainers:
--- a/charts/seerr-chart/README.md
+++ b/charts/seerr-chart/README.md
@@ -73,6 +73,26 @@ If `replicaCount` value was used - remove it. Helm update should work fine after
 | podLabels | object | `{}` |  |
 | podSecurityContext.fsGroup | int | `1000` |  |
 | podSecurityContext.fsGroupChangePolicy | string | `"OnRootMismatch"` |  |
+| postgres.db_log_queries | bool | `false` |  |
+| postgres.db_name | string | `"seerr"` | Database name |
+| postgres.db_secret | string | `""` | Kubernetes basic-auth secret containing DB username and password |
+| postgres.enabled | bool | `false` | Specifies whether to use postgresql |
+| postgres.migration.enabled | bool | `false` | Enable SQLite to PostgreSQL migration initContainer using pgloader |
+| postgres.migration.image.pullPolicy | string | `"IfNotPresent"` |  |
+| postgres.migration.image.registry | string | `"ghcr.io"` |  |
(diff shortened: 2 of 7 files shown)


M0NsTeRRR:
Hello,

I’m not sure I want to accept this. First, there’s already a PR https://github.com/seerr-team/seerr/pull/2502 that implements a migration script using TypeORM, which is far better than using pgloader and could also work for users who don’t use Kubernetes.

If we were to support this feature, we’d first need to support a basic PostgreSQL setup with a Helm chart dependency for example.

Regarding the goal of this PR, I don’t see the added value, since it still requires manually enabling and disabling the migration step (even though it’s not necessary, the system will still attempt to run it and increase startup time).

damnmersault:
> Hello,
> 
> I’m not sure I want to accept this. First, there’s already a PR #2502 that implements a migration script using TypeORM, which is far better than using pgloader and could also work for users who don’t use Kubernetes.
> 
> If we were to support this feature, we’d first need to support a basic PostgreSQL setup with a Helm chart dependency for example.
> 
> Regarding the goal of this PR, I don’t see the added value, since it still requires manually enabling and disabling the migration step (even though it’s not necessary, the system will still attempt to run it and increase startup time).

The goal was purely personal - I wanted to migrate my database and PR #2502 isn't merged or released (also, I hadn't seen that PR). That PR is probably a better solution, but I am impatient. My data is migrated now, so it's no issue to me if this PR isn't accepted.

The semaphore will prevent the migration from executing more than once, but you're right that with migration enabled deployment will be slowed by having to pull the pgloader image and run the initContainers (even though they'll exit essentially immediately). Without that PR though there's no other way, since there's no native kubernetes support for conditionals or semaphores.

As for adding a dependent postgresql chart, I think you could add a note saying it's out of scope and linking to a few options. Anyone deploying Seerr to kubernetes via helm is likely going to be opinionated about their postgresql deployment.

You may want to grab the PVC change regardless, as it's just a little tidier than the current conditional in statefulset.yaml.

M0NsTeRRR:
> As for adding a dependent postgresql chart, I think you could add a note saying it's out of scope and linking to a few options. Anyone deploying Seerr to kubernetes via helm is likely going to be opinionated about their postgresql deployment.

Yes, that’s why this PostgreSQL option should be disabled by default (since SQLite is sufficient as a default choice), but it would still be useful to support an official, limited PostgreSQL instance for those who don’t want to manage an external database as a lot of charts do.

> You may want to grab the PVC change regardless, as it's just a little tidier than the current conditional in statefulset.yaml.

Yes would you mind opening an other PR for that ?

PR #2654, opened 2026-03-07
refactor(email): add default body fallback for test notification template

8emk10:
Description

This PR adds a small fallback for the test email notification template.

Currently the template renders `#{body}` directly.  
If the `body` variable is not provided, the email content may appear blank.

This change updates the template to:

#{body || 'Check check, 1, 2, 3. Are we coming in clear?'}

This ensures that a visible message is always rendered even if `body` is missing.

This change is intentionally minimal and only affects the **test notification template**.

It also helps prepare the template behavior for future improvements discussed around **multi-language email templates**, where template rendering may depend on localized content generation.

Modified file:

server/templates/email/test-email/html.pug

How Has This Been Tested?

- Template compiled successfully with `pug`
- Local development build succeeded
- Container started successfully
- Test email rendered correctly
- Verified that the fallback message appears when `body` is undefined

Screenshots / Logs (if applicable)

Test email rendered locally showing the fallback message.

Checklist:

- [x] Disclosed any use of AI (see our policy).

## Summary by CodeRabbit

* **Bug Fixes**
  * Test email template now displays a default message when the body field is empty, ensuring test emails always contain visible content instead of appearing blank.

### DIFF
--- a/server/templates/email/test-email/html.pug
+++ b/server/templates/email/test-email/html.pug
@@ -33,7 +33,7 @@ div(style='display: block; background-color: #111827; padding: 2.5rem 0;')
     tr
       td(style='text-align: center;')
         div(style='margin: 1rem 0 0; font-size: 1.25em;')
-          | #{body}
+          | #{body || 'Check check, 1, 2, 3. Are we coming in clear?'}
     if applicationUrl
       tr
         td


0xSysR3ll (changes requested):
Can you update your PR to use our template please ? https://github.com/seerr-team/seerr/blob/develop/.github/PULL_REQUEST_TEMPLATE.md

8emk10:
Updated the PR description to follow the project template. Thanks!

0xSysR3ll:
I'm not sure I understand in which scenario body would be `undefined`.

In the current test notification flow the message is already hard-coded:

```ts
const sendTestNotification = async (agent: NotificationAgent, user: User) =>
  await agent.send(Notification.TEST_NOTIFICATION, {
    notifySystem: true,
    notifyAdmin: false,
    notifyUser: user,
    subject: 'Test Notification',
    message: 'Check check, 1, 2, 3. Are we coming in clear?',
  });
```
So the template should always receive a value for the body.

Could you clarify how body might end up being `undefined` in this case? I'm trying to understand the scenario this change is protecting against.

8emk10:
Good point — you're right that in the current implementation the test notification flow always provides a `message` value:

message: 'Check check, 1, 2, 3. Are we coming in clear?'

So in the current code path the template should indeed always receive a `body` value.

The motivation for this change is mainly related to a longer-term goal around the email templates. Currently some email content is still partially defined in the notification sender. The idea is to gradually move email content fully into the templates so the templates become the single source of truth for rendering emails.

This is especially useful for future improvements around localization / multi-language email templates, where different template folders (e.g. `en`, `de`, etc.) would render the email content themselves.

In that scenario templates should ideally be able to render independently without relying on the sender always providing a `message` string.

That said, if you prefer to keep the template strictly dependent on the sender always providing `message`, I'm happy to remove the fallback. I mainly added it as a small defensive guard while experimenting with template-driven rendering.

8emk10:
Hi! Just following up on this in case it got buried.

I agree that in the current implementation `message` is always provided for the test notification flow, so this fallback is not fixing a current runtime issue.

My intention with this PR is to introduce a very small preparatory step toward moving email content out of the notification sender and into the templates themselves.

The larger goal is to make templates the source of truth for rendered email content, which would also make later localization / language-specific templates much easier to support incrementally.

So this change is meant less as a fix for today's flow and more as a safe first migration step toward template-driven email rendering.

If you’d rather not take preparatory steps like this in isolation, that’s totally fine — in that case I can close this PR and come back with a follow-up that makes the migration step more explicit.

0xSysR3ll:
@seerr-team/seerr-core wdyt ?

gauthier-th:
I think that this is fully AI-generated and non-sense. @8emk10 please respect us and our time and don't send copy/paste of ChatGPT response. Next PR like this will result in a ban.

PR #2746, opened 2026-03-22
fix(auth): resolve Plex OAuth client ID mismatch

### LINKED ISSUE #2736: Plex auth popup does not use the stored clientId
### Description

During the Plex auth popup process, Seerr is not including the stored clientId from the config files. This causes every browser to generate its own unique clientId. This clientId is not expected by Plex, which makes Plex return a blank auth response, which locks the users out.  This can be seen here:

```
root@Plex:/mnt/cache/appdata/swag/nginx/proxy-confs# curl -s http://localhost:5056/api/v1/settings/public | python3 -m json.tool | grep -i "client\|plex"
    "applicationUrl": "https://request.myplex.com/",
    "newPlexLogin": false,```
```

If the auth process passed in the stored clientId, this would allow Plex to see a previously authenticated client ID and complete the OAuth process.

### Version

2.7.3

### Steps to Reproduce

1. Go to Seerr
2. Click Authenticate with Plex
3. Approve the authentication in the Plex popup
4. Look at the empty response from the POST in Network dev tools
5. Not be logged into Seerr or get past the login page

I have tried nightly, your own container and hotio's container. All behave the same way.

### Screenshots

Image

Image

response - See "authToken:null"

{"id":1589111464,"code":"u4fsabbj9zomh3j6i2nimg2a3","product":"Jellyseerr","trusted":false,"qr":"https://plex.tv/api/v2/pins/qr/u4fsabbj9zomh3j6i2nimg2a3","clientIdentifier":"8cfe391f-b8d8-40ab-b453-734967d5b36d","location":{"code":"US","european_union_member":false,"continent_code":"NA","country":"United States","city":"none","time_zone":"Etc/UTC","postal_code":"none","in_privacy_restricted_country":false,"in_privacy_restricted_region":false,"subdivisions":"New York","coordinates":"redacted3, redacted"},"expiresIn":1800,"createdAt":"2026-03-20T19:19:34Z","expiresAt":"2026-03-20T19:49:34Z","authToken":null,"newRegistration":null}

### Logs

```shell
Seerr is not logging the response because it's not getting a successful response.
```

### Platform

desktop

### Database

SQLite (default)

### Device

Unraid 7.2.4

### Operating System

Unraid 7.2.4

### Browser

Firefox

### Additional Context

I did try to work through this issue 2 weeks ago with the Seerr team in discord, but we couldn't figure it out. I have since discovered more information to help troubleshoot, and narrowed it down to this null response issue.

### Search Existing Issues

### Code of Conduct

0xSysR3ll:
## Description

This PR fixes an issue where Plex Oauth login could fail because the browser and server were using different client identifiers.
The browser generated its own ID (`plex-client-id`, stored in localStorage), while the server used `clientId` (stored in `settings.json`. Now both use the same value: `plexClientIdentifier`

As part of this PR, I also discovered a security issue:
`clientId` was being used to sign sessions even though it is a public value.

To address this, this PR introduces a new `sessionSecret` dedicated to session signing.

> [!WARNING]
> Introducing this new session secret will obviously invalidate all existing sessions, meaning all users will be logged out.

- Fixes #2736

## How Has This Been Tested?

(Not tested yet but in theory should work).

## Checklist:

## Summary by CodeRabbit

* **Improvements**
  - Plex authentication now uses server-provided client identifiers for more consistent, reliable logins.
  - Sessions use dedicated, server-generated secrets for stronger session protection.
  - Plex account linking includes explicit client identification and improved error handling for clearer outcomes.
  - Popup handling and environment checks are more robust, reducing authentication failures and improving reliability.


server/lib/settings/index.ts
```diff
@@ -752,6 +756,10 @@ class Settings {
     return this.data.clientId;
   }
 
+  get sessionSecret(): string {
+    return this.data.sessionSecret!;
```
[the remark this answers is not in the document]
0xSysR3ll: Nice catch but actually I think this is kind of "out of scope" here. IMO this could be fixed in a dedicated follow-up PR so we can fix it once for all settings instead of piecemeal.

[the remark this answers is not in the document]
0xSysR3ll: No

src/utils/plex.ts
fallenbagel: Wouldn't `!this.popup?.closed` evaluate to true when `this.popup` is `undefined`? so if the popup was never opened or gets blocked, this might poll forever. Also I think `!response.data?.authToken` check is redundant since the first if already handles the truthy case. Although this is out of diff of current PR, i think it makes sense to fix it on this.

```suggestion
        } else if (this.popup && !this.popup.closed) {
```

fallenbagel: I think the CodeRabbit's suggestion here is also worth applying so we fail fast instead of entering the poll at all.

> Fail fast when the popup never opens. If `window.open()` is blocked or `preparePopup()` was skipped, Lines 111-117 still fall through to `pinPoll(). pinPoll()` treats `this.popup === undefined` as "not closed", so the login flow polls forever instead of returning an actionable error.

```suggestion
    if (!this.popup || this.popup.closed) {
      this.closePopup();
      throw new Error(
        'Unable to open Plex login popup. Please allow popups and try again.'
      );
    }

    this.popup.location.href = `https://app.plex.tv/auth/#!?${this.encodeData(
      params
    )}`;

    return this.pinPoll();
  }
```

fallenbagel (changes requested):
Overall looks good. A few comments

src/utils/plex.ts
0xSysR3ll: Yep you're right. I planned to do a follow-up PR but I guess since it's a small fix it makes sense including it in this one.

0xSysR3ll: Same as below.

server/lib/settings/migrations/0009_migrate_session_secret.ts
fallenbagel, line 1: Isn't this migration unnecessary? The `load()` method already has an established pattern for generating missing secrets (`clientId`, `apiKey`, and VAPID keys all use the same approach)
Here is the example:
```ts
if (!this.data.clientId) {
    this.data.clientId = randomUUID();
    change = true;
}
```
And you also have added same approach for `sessionSecret` below that already:
```ts
    if (!this.data.sessionSecret) { 
       this.data.sessionSecret = randomBytes(32).toString('hex'); 
       change = true; 
     }
```
 in `load()`. This is all that's needed and will be generated on first run and persisted via `save()`. The `mergeSettings` (lodash `mergeWith`) already handles new default values from the constructor, and the explicit check covers generating a unique value.

Settings migrations are only needed for transforming existing data (like renaming keys, restructuring objects, removing obsolete fields). It is not needed for adding new fields with generated values.

fallenbagel (changes requested):

server/lib/settings/migrations/0009_migrate_session_secret.ts
0xSysR3ll: Yep, that's on me my bad

src/utils/plex.ts
fallenbagel: So is this going to be a follow up pr or in this? Along with the above one. Just making sure because this was marked as resolved (so I can add a note to track incase this gets buried)

server/lib/settings/index.ts
```diff
@@ -387,6 +389,7 @@ class Settings {
   constructor(initialSettings?: AllSettings) {
     this.data = {
       clientId: randomUUID(),
+      sessionSecret: randomBytes(32).toString('hex'),
```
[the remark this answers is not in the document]
fallenbagel: @0xSysR3ll did you test this?

CodeRabbit's finding about `sessionSecret` not persisting might be correct. Since it's set in the constructor, `mergeSettings` preserves it when the file doesn't have it, the backfill check sees a truthy value, skips `save()`, and the value might be lost on restart, thus, new secret every restart, all sessions invalidated. Test if this is true or not.

I think `clientId` doesn't have this problem because every existing `settings.json` already has it persisted from day one. The file's value overwrites the constructor's during merge, so the backfill check is irrelevant. It's already the real persisted value.

I think we would need to set `sessionSecret` to `''` in the constructor (like `apiKey` is handled) so the backfill in `load()` detects it as missing, generates it, and triggers `save()`. A whole migration for this is still unnecessary though I would say.

Wdyt? @gauthier-th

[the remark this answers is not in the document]
0xSysR3ll: Yeah, confirmed on my side after a test - the assumption holds. Setting `sessionSecret` to `''` in the constructor should be sufficient, no need for a migration.

fallenbagel (approved):
Success!

gauthier-th (approved):

PR #2767, opened 2026-03-26
feat(timezone): fall back to /etc/timezone and /etc/localtime when TZ is not set

### LINKED ISSUE #2523: Pull from /etc/timezone?
### Description

Timezone always appears as `Etc/UTC` if `TZ` isn't specified even if `/etc/timezone` exists.

### Desired Behavior

I find it much easier to maintain and not bother with `TZ` variables when I can simply mount `/etc/timezone:readonly` for containers.

### Additional Context

_No response_

### Search Existing Issues

### Code of Conduct

Logvin:
## Description

When `TZ` is not set, Seerr defaults to UTC even if the host timezone is available through standard Linux files. This adds a startup check that resolves the timezone from `/etc/timezone` or `/etc/localtime` before any date operations run.

This matters for features like "upcoming" content discovery, which uses timezone offsets to calculate release dates. If Seerr thinks it's UTC but you're in America/Phoenix (UTC-7), those dates can be off by hours.

### What changed

- **New file: `server/utils/resolveTimezone.ts`** - utility that checks `TZ` env var first, then `/etc/timezone` (Debian/Ubuntu, plain text), then `/etc/localtime` symlink (Alpine/RHEL, resolves from zoneinfo path). Sets `process.env.TZ` if a source is found.
- **`server/index.ts`** - calls `resolveTimezone()` before the "Starting Seerr" log, so timezone is resolved before any date math.
- **`docs/getting-started/docker.mdx`** - added mount options (`/etc/timezone:ro` and `/etc/localtime:ro`) to the existing timezone warning block so Docker users know they have alternatives to the `TZ` env var.

- Fixes #2523

## How Has This Been Tested?

Three scenarios tested in Docker on a live server:

1. **TZ env var set** (`-e TZ=America/Phoenix`) - env var takes priority, no timezone log. Existing behavior preserved.
2. **`/etc/timezone` mounted** (`-v /etc/timezone:/etc/timezone:ro`) - reads the file, logs `Timezone set to America/Phoenix (from /etc/timezone)`.
3. **`/etc/localtime` symlink** (symlink to `/usr/share/zoneinfo/America/Phoenix`) - reads the symlink target, logs `Timezone set to America/Phoenix (from /etc/localtime symlink)`.

- `pnpm build` succeeds
- `pnpm i18n:extract` produces no changes (no user-facing text modified)
- TypeScript compiles with zero errors
- ESLint + Prettier pass

Test 2 output (`/etc/timezone` mounted):
```
[info][Timezone]: Timezone set to America/Phoenix (from /etc/timezone)
[info]: Starting Seerr version develop-
```

## Checklist:

**AI Disclosure:** This PR was developed with Claude Code as a pair programming tool. I directed the approach, reviewed all code, and tested all three scenarios on a live server. I'm the author, Claude is my editor.

## Release Notes

* **Documentation**
  * Updated Docker setup guidance to include an alternative timezone configuration method using mounted host timezone files.

* **New Features**
  * Application now automatically detects and configures timezone from mounted host timezone files when the TZ environment variable is not set.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,4 +55,5 @@ RUN touch config/DOCKER && \
 
 EXPOSE 5055
 
+ENTRYPOINT [ "/app/docker-entrypoint.sh" ]
 CMD [ "npm", "start" ]
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -1,0 +1,16 @@
+#!/bin/sh
+
+# Resolve timezone from mounted files when TZ is not explicitly set.
+# This lets Docker users mount /etc/timezone or /etc/localtime instead
+# of passing -e TZ=Region/City.
+if [ -z "$TZ" ]; then
+  if [ -f /etc/timezone ]; then
+    TZ=$(cat /etc/timezone | tr -d '[:space:]')
+    export TZ
+  elif [ -L /etc/localtime ]; then
+    TZ=$(readlink /etc/localtime | sed 's|.*/zoneinfo/||')
+    export TZ
+  fi
+fi
+
+exec "$@"
--- a/docs/getting-started/docker.mdx
+++ b/docs/getting-started/docker.mdx
@@ -33,6 +33,12 @@ Be sure to replace `/path/to/appdata/config` in the below examples with a valid 
 
     The `TZ` environment variable value should also be set to the [TZ database name](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones) of your time zone!
 
+    Alternatively, you can mount your host's timezone file instead of setting the `TZ` variable:
+    - **Debian/Ubuntu:** `-v /etc/timezone:/etc/timezone:ro`
+    - **Alpine/RHEL:** `-v /etc/localtime:/etc/localtime:ro`
+
+    Seerr will automatically detect the timezone from these files if `TZ` is not set.
+
 :::
 
 import Tabs from '@theme/Tabs';


fallenbagel (changes requested):
I find this implementation a bit messy.

I think this would be better handled at the Docker entrypoint level rather than in the application codebase. Resolving `/etc/timezone` and `/etc/localtime` is an OS/container-level concern, and introducing a new TypeScript module with symlink resolution, and error handling to work around Docker's environment feels like the wrong layer for this.

We should keep the app itself agnostic to where `TZ` comes from and let it just trust the environment which is the standard expectation for Node apps.

For non-Docker deployments (bare metal, systemd), `TZ` is typically already inherited from the system environment, so this is almost exclusively a container problem, which makes deriving it as an entrypoint a more natural place to solve it.

fallenbagel:
> This matters for features like "upcoming" content discovery, which uses timezone offsets to calculate release dates. If Seerr thinks it's UTC but you're in America/Phoenix (UTC-7), those dates can be off by hours.

This isn't accurate either. IIRC, TMDB release dates are plain dates (midnight UTC), and we don't localise them and display them as is as localising would create even more confusion around release dates as they would be even more inaccurate for most timezones. So the `TZ` value doesn't actually affect release date display at all.

Logvin:
I did some research into how other projects in this space handle this. Looked at LinuxServer (Sonarr, Radarr, Plex), Jellyfin, Home Assistant, Gitea, Grafana, Nextcloud, and OpenHAB.

The short version: almost nobody does entrypoint-level timezone resolution. The standard pattern across the ecosystem is:

1. Install `tzdata` in the image (Seerr already does this)
2. Document that users should set `TZ=Region/City` as an env var
3. Let the runtime handle it

The only project that touches timezone in its entrypoint is OpenHAB, and that's because Java's JVM specifically needs `/etc/localtime` as a symlink. Node.js just reads `process.env.TZ` directly.

Given that, I think the better fix here might just be improving the docs rather than adding entrypoint logic. The existing `TZ` env var approach is already the ecosystem standard. We could make the Docker getting started page more explicit about it and call it done.

Happy to go either direction. Want me to strip this back to a docs-only PR, or would you rather just close the issue?

fallenbagel:
> The existing TZ env var approach is already the ecosystem standard

Yes it is

PR #2793, opened 2026-04-01
fix(auth): await session.save() before responding to prevent missing Set-Cookie header

kenleejl:
## Description

All three login endpoints (/plex, /jellyfin, /local) set req.session.userId but immediately call res.json() without awaiting session.save(). With saveUninitialized: false and an async TypeormStore backend, the session write races against the HTTP response — the Set-Cookie: connect.sid header is never sent to the browser. Every subsequent GET /api/v1/auth/me returns 401 even after a successful credential check, leaving the user permanently stuck on the login page.

Root cause: express-session with an async store requires session.save() to complete before headers are flushed. Without an explicit await, res.json() exits before the store callback fires.

Fix: added await new Promise<void>((resolve) => req.session.save(resolve)) in all three auth handlers before returning the 200 response.

Fixes login returning HTTP 200 with user data but no Set-Cookie header
Fixes all subsequent /api/v1/auth/me calls returning 401 after successful login

This PR was debugged and authored with AI assistance (Cursor). I have reviewed the code, understand the fix, and verified it manually on a running v3.1.0 instance.

- Fixes #XXXX

## How Has This Been Tested?

Reproduced and verified on v3.1.0 with Jellyfin auth + TypeormStore (SQLite):

Sent POST /api/v1/auth/jellyfin with valid credentials — confirmed response had no Set-Cookie header before fix, Set-Cookie: connect.sid=... present after fix
Verified session row count in SQLite session table increments on login after fix (previously unchanged)
Confirmed GET /api/v1/auth/me returns 200 with valid session cookie after fix
Tested all three handlers: /plex, /jellyfin, /local

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session persistence during authentication to ensure user login sessions are reliably saved across Plex, Jellyfin, and local authentication methods.

### DIFF
--- a/server/routes/auth.ts
+++ b/server/routes/auth.ts
@@ -203,6 +203,7 @@ authRoutes.post('/plex', async (req, res, next) => {
     // Set logged in session
     if (req.session) {
       req.session.userId = user.id;
+      await new Promise<void>((resolve) => req.session.save(resolve));
     }
 
     return res.status(200).json(user?.filter() ?? {});
@@ -505,6 +506,7 @@ authRoutes.post('/jellyfin', async (req, res, next) => {
     // Set logged in session
     if (req.session) {
       req.session.userId = user?.id;
+      await new Promise<void>((resolve) => req.session.save(resolve));
     }
 
     return res.status(200).json(user?.filter() ?? {});
@@ -628,6 +630,7 @@ authRoutes.post('/local', async (req, res, next) => {
     // Set logged in session
     if (user && req.session) {
       req.session.userId = user.id;
+      await new Promise<void>((resolve) => req.session.save(resolve));
     }
 
     return res.status(200).json(user?.filter() ?? {});


fallenbagel:
Thanks for the PR, but what this PR claims shouldn't be the case, therefore, I am closing this one.

`express-session` already handles this internally. The middleware wraps `res.end()` and ensures the session is persisted to the store before headers are flushed. This is by design and works the same way regardless of whether the store backend is sync or async. An explicit `session.save()` before `res.json()` is redundant.

I haven't been able to reproduce the issue you described, and the underlying premise of the race condition doesn't hold up given how the middleware works.

If you're genuinely experiencing missing `Set-Cookie` headers, the root cause is likely elsewhere. Some common culprits include:

- Reverse proxy stripping or not forwarding `Set-Cookie`
- Missing `trust proxy` configuration when behind a proxy
- Having "Enable CSRF Protection" turned on without HTTPS

Happy to help debug if you can share your setup details and logs in a new support ticket in our support channel.

CC: @seerr-team/seerr-core 

For example:
```
curl -v -X POST http://localhost:5055/api/v1/auth/jellyfin \
  -H "Content-Type: application/json" \
  -d '{"username":"youruser","password":"yourpass"}' \
  2>&1 | grep -i set-cookie
```

creaux:
@fallenbagel I can independently confirm this bug is real, and reproduce it cleanly outside of any of the three causes you suspected. Wanted to share concrete evidence in case it changes the picture.

**Setup:** Seerr v3.2.0, `DB_TYPE=postgres`, Postgres running in a separate Docker container on the same host (reached over the Docker bridge network, not localhost/unix-socket).

**Ruling out your three suspected causes, one at a time:**

1. *CSRF protection without HTTPS* — `csrfProtection` is `false` in my settings. Not the cause.
2. *Missing trust proxy* — `network.trustProxy` is `true`, and confirmed working (real client IPs show up correctly in logs via `X-Forwarded-For`, not the Docker gateway IP).
3. *Reverse proxy stripping `Set-Cookie`* — ruled out by bypassing the reverse proxy entirely and hitting the container directly:

```
curl -s -D - -o /dev/null -X POST -H "Content-Type: application/json" \
  -d '{"username":"<user>","password":"<pass>"}' \
  http://127.0.0.1:5055/api/v1/auth/jellyfin
```

Result on unpatched v3.2.0 (correct credentials, 200 response, user matched and updated in the logs):

```
HTTP/1.1 200 OK
X-Powered-By: Express
Content-Type: application/json; charset=utf-8
Content-Length: 516
ETag: W/"204-AKzlknFU6qQfPC++4/xTKjH2Isc"
Date: Wed, 26 Aug 2026 18:16:47 GMT
```

No `Set-Cookie`, no reverse proxy, no TLS involved at all in this request. The `session` table in Postgres was also unchanged (row count and max `expiredAt` identical before and after several repeated login attempts) — the store write was never even reaching Postgres, not just losing a header race after writing.

**Applying the exact fix from this PR resolved it.** I patched the compiled `dist/routes/auth.js` to add `await new Promise((resolve) => req.session.save(resolve));` after each of the three `req.session.userId = ...` assignments (mirroring this PR's diff), restarted, and reran the same curl:

```
HTTP/1.1 200 OK
X-Powered-By: Express
Content-Type: application/json; charset=utf-8
Set-Cookie: connect.sid=...; Path=/; Expires=Fri, 25 Sep 2026 18:27:06 GMT; HttpOnly; SameSite=Lax
```

A new row appeared in the `session` table immediately, and subsequent `GET /api/v1/auth/me` calls returned 200 instead of 401. Confirmed again through the actual reverse-proxy/HTTPS path too, with `Secure` correctly attached this time (`trustProxy` working as expected once the session save is actually awaited).

**Why this is probably invisible on the default SQLite backend:** SQLite writes here happen in-process, synchronously enough that the unawaited `store.set()` call all but always "wins" the race against the response being flushed, even without an explicit await. Postgres in a separate container introduces a real network round-trip for that same write, which is enough latency to reliably lose the race instead of occasionally losing it. That would explain why this wasn't reproducible on your end but reproduces 100% of the time here.

Happy to share more logs/details if useful. Given this reproduces deterministically and the suggested fix directly resolves it without any observed downside, would you be open to reopening/reconsidering this?

fallenbagel:
> **Why this is probably invisible on the default SQLite backend:** SQLite writes here happen in-process, synchronously enough that the unawaited `store.set()` call all but always "wins" the race against the response being flushed, even without an explicit await. Postgres in a separate container introduces a real network round-trip for that same write, which is enough latency to reliably lose the race instead of occasionally losing it. That would explain why this wasn't reproducible on your end but reproduces 100% of the time here.
> 

**I run Postgres. Not SQLite.** I already said that above, and it removes the entire basis of your explanation.

Beyond that, the mechanism you are describing does not exist. I read `express-session`'s source before closing this, and again now.

`Set-Cookie` is not written by the save path. It is written by an `onHeaders` hook registered at the top of the middleware. The only gate on it is `shouldSetCookie()`, which reads `cookieId`, `saveUninitialized`, `rollingSessions` and `isModified()` (`originalId` / `originalHash`). An explicit `req.session.save()` sets `savedHash`, which feeds `shouldSave`, `shouldTouch` and `isSaved` and nothing else. It has no bearing on whether the cookie is emitted. The patch you applied cannot produce the before/after you posted.

The race is backwards too. When `shouldSave` is true, the wrapped `res.end` calls `writetop()`, which runs `res._implicitHeader()` immediately. Headers flush first, the cookie is written there, and only the final byte of the body is held until `store.set` calls back. A slow store delays the body finishing. It cannot retract a header that was already written before the store was touched. Postgres over a bridge network widens that window, it does not delete the header.

Your own evidence contradicts your theory. If no row reached Postgres across repeated logins, `store.set` was never called at all. That means `shouldSave` was false or `req.session` was undefined at `res.end`, which is not a lost race, and neither is fixed by adding a save inside `if (req.session)`. `Secure` also has nothing to do with awaiting a save. `secure: 'auto'` is resolved in `store.generate` from `issecure(req, trustProxy)`, before any save runs.

And the PR author says he reproduced this on SQLite. You say SQLite hides it because the write always wins. **Both cannot be true.** Also the versions you're using are not recent either. Please reproduce on develop.

Also, this reads like it was generated, and so did the PR. Please do read contributing guideline's AI policy first.

If you have a login with no `Set-Cookie` on latest develop, run it with `DEBUG=express-session` and post the log for that single request. It prints the branch it took: `no session`, `not secured`, or `saving`. That settles it. Without that, this stays closed. And I would recommend discussing this with us on Discord with your evidence if you want this looked at further as **I cannot reproduce it on SQLite or Postgres, and nothing you have claimed is supported by the source.**

0xSysR3ll:
@creaux  I went and re-checked this on my own, because the store-race story still didn't sit right with us.

Tested on latest develop commit.

I deliberately threw a slow store at it (300–500ms on every `set`). Cookie was there. Session was written.
With the delay, headers/`Set-Cookie` showed up around ~80ms and the body only finished after the store write (~576ms).
So point is, the slow write holds the body, it does not eat the cookie.

Adding `await session.save()` didn't change whether the cookie showed up; it just made headers wait longer.

Then full stack (with postgres) with `DEBUG=express-session`.
Here's what I observed:

```http
HTTP/1.1 200 OK
Set-Cookie: connect.sid=s%3ASGM_G_17SHfC-zGYAnnvetQQps8F6p_l....; Path=/; Expires=Fri, 25 Sep 2026 19:49:36 GMT; HttpOnly; SameSite=Lax
```

Session table went from `0` to `1`.
A request to `/auth/me` with that cookie returned `200`.

Debug for that request was:

```text
express-session no SID sent, generating session
express-session saving SGM_G_17SHfC-zGYAnnvetQQps8F6p_l
express-session set-cookie connect.sid=...
express-session split response
express-session fetching SGM_G_17SHfC-zGYAnnvetQQps8F6p_l
express-session session found
```

So... I'm not seeing the failure mode you described. Not on a slow store, not on Postgres + `connect-typeorm`, not on a real login.

A bridge-network hop doesn't explain a missing `Set-Cookie`, and awaiting `save()` isn't what puts that header on the response.

If you're still hitting this on **latest develop**, please grab one clean request against the container directly (no reverse proxy) and share it:

```bash
apk update && apk add curl -y
DEBUG=express-session
curl -v -X POST http://127.0.0.1:5055/api/v1/auth/jellyfin \
  -H 'Content-Type: application/json' \
  -d '{"username":"...","password":"..."}'
```

For that single request: please provide the response headers, the `express-session` DEBUG lines, and whether a session row appeared.

PR #2760, opened 2026-03-25
fix: await missing repository saves

0xSysR3ll:
## Description

While working on something, I noticed the movie branch was calling `save()` without a proper `await`.

After fixing that, I did a quick grep for other `Repository.save` usages to make sure they were properly awaited (or intentionally handled with something like `Promise.all`).

I might've missed a few spots, but this should cover the obvious ones.

- Fixes #XXXX

## Checklist:

- [x] I have read and followed the contribution [guidelines](https://github.com/seerr-team/seerr/blob/develop/CONTRIBUTING.md).
- [ ] Disclosed any use of AI (see our [policy](https://github.com/seerr-team/seerr/blob/develop/CONTRIBUTING.md#ai-assistance-notice))
- [ ] I have updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [x] Successful build `pnpm build`
- [ ] Translation keys `pnpm i18n:extract`
- [ ] Database migration (if required)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed data persistence issues affecting media request submissions, user notification preferences, and status management. All database operations now complete fully before confirming changes, improving reliability and preventing potential data loss.

PR #2804, opened 2026-04-02
test(usersettings): add integration tests for notification settings route

dougrathbone:
## Description

Adds integration tests for the `POST /user/:id/settings/notifications` route to provide regression coverage for the unawaited `userRepository.save()` fixed in #2760.

Prior to #2760, the handler returned `200` before the database write completed. There were no tests asserting the settings were actually persisted to the database, so the race condition went undetected. This suite locks in the expected behavior: changes must be readable from the database after the response returns.

**AI Disclosure:** I used Claude Code to help write the tests. I reviewed and understood all generated code before submitting.

## How Has This Been Tested?

New test suite at `server/routes/user/usersettings.test.ts` using the real SQLite test database via `setupTestDb()`, following the same pattern as `server/routes/auth.test.ts`:

- `persists notification settings to the database` -- the core regression guard: re-fetches the user from the DB with `relations: { settings: true }` after the POST and asserts the saved values match what was sent. This test would have caught the unawaited save.
- `returns the updated values in the response` -- asserts the response body reflects the values that were sent.
- `returns 404 for a non-existent user` -- asserts the correct error response for a missing user ID.
- `returns 403 for a non-owner non-admin trying to update another user settings` -- asserts the authorization guard is enforced.

All 4 tests pass.

N/A -- test-only change with no runtime behaviour change.

## Checklist:

- [x] I have updated the documentation accordingly. (no documentation changes required for a test-only PR)
- [x] Database migration (if required) (no schema changes)

## Summary by CodeRabbit

* **Tests**
  * Added an integration test suite for the user notification settings endpoint. Verifies updates persist, response payload reflects changes, returns 404 for missing users, and enforces authorization (403) for non-owners. Includes test setup for authenticated agents and coverage for error-handling behavior.

### DIFF
--- a/server/routes/user/usersettings.test.ts
+++ b/server/routes/user/usersettings.test.ts
@@ -23,6 +23,8 @@ function createApp() {
       secret: 'test-secret',
       resave: false,
       saveUninitialized: false,
+      // secure: false is intentional -- supertest uses HTTP, not HTTPS
+      cookie: { secure: false },
     })
   );
   app.use(checkUser);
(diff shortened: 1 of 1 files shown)


dougrathbone:
Fixed in the latest commit:

- **Settings leak**: `authenticatedAgent` now saves `settings.main.localLogin` before setting it to `true` and restores it in a `finally` block.
- **Clear-text cookie (GitHub Advanced Security)**: added `cookie: { secure: false }` explicitly to the test session config to make the intent clear -- supertest uses HTTP, not HTTPS, so this is intentional. The flag was always implicitly false; making it explicit also suppresses the finding.

gauthier-th:
Closing as discussed [here](https://github.com/seerr-team/seerr/pull/2806#issuecomment-4276993747).

PR #2805, opened 2026-04-02
perf(count-routes): parallelize sequential count queries with Promise.all

dougrathbone:
## Description

Both `GET /issue/count` and `GET /request/count` shared the same performance problem: they created a single mutable TypeORM `QueryBuilder` instance and called `.where(...).getCount()` on it in sequence. Each call awaited the previous one before starting, meaning **7 and 9 sequential database round trips** respectively for what are entirely independent queries.

The issue count handler ran these one after another:

```ts
// Before - 7 sequential awaits
const query = issueRepository.createQueryBuilder('issue');
const totalCount = await query.getCount();
const videoCount = await query.where('issue.issueType = :issueType', { issueType: IssueType.VIDEO }).getCount();
// ... 4 more
```

The request count handler had the same pattern across 9 counts, with an unnecessary `innerJoinAndSelect` on the base query (loading the full `media` entity) even for counts that only filter on `request.*` columns.

**Fix:** Replace both handlers with `Promise.all()` over independent `repository.count({ where: ... })` calls. The two request counts that genuinely need a join (`processing` and `available`, which filter on `media.status`/`media.status4k`) each use a shared `approvedMediaStatusCount(isAvailable)` helper to eliminate the duplicated QueryBuilder setup.

```ts
// After - all run concurrently
const [total, video, audio, subtitles, others, open, closed] = await Promise.all([
  issueRepository.count(),
  issueRepository.count({ where: { issueType: IssueType.VIDEO } }),
  issueRepository.count({ where: { issueType: IssueType.AUDIO } }),
  issueRepository.count({ where: { issueType: IssueType.SUBTITLES } }),
  issueRepository.count({ where: { issueType: IssueType.OTHER } }),
  issueRepository.count({ where: { status: IssueStatus.OPEN } }),
  issueRepository.count({ where: { status: IssueStatus.RESOLVED } }),
]);
```

**AI Disclosure:** I used Claude Code to help write the tests and validate the approach. I reviewed and understood all generated code before submitting.

## Performance improvement

| Endpoint | Before | After | Improvement |
|---|---|---|---|
| `GET /issue/count` | 7 sequential queries | 7 concurrent queries | Up to 7x faster (PostgreSQL); proportional to DB latency |
| `GET /request/count` | 9 sequential queries + unnecessary JOIN on 7 of them | 9 concurrent queries, JOIN only where needed | Up to 9x faster (PostgreSQL) |

The gain is most visible on PostgreSQL where each query has non-trivial round-trip latency. On SQLite (single connection, serialised writes) the queries still execute in order but the code is clearer and the unnecessary JOIN is eliminated regardless.

Both endpoints are called on page load for the admin dashboard and issue management pages, so the improvement is user-visible on every navigation to those pages.

## How Has This Been Tested?

Integration tests using the real SQLite test database via `setupTestDb()`.

**`server/routes/issue.test.ts`** (3 tests):
- Returns zero counts when no issues exist
- Counts issues by type (VIDEO, AUDIO, SUBTITLES, OTHER)
- Counts issues by status (OPEN, RESOLVED)

**`server/routes/request.count.test.ts`** (4 tests):
- Returns zero counts when no requests exist
- Counts requests by type (movie, tv)
- Counts requests by status (pending, approved, declined)
- Counts processing and available correctly for HD and 4K requests -- seeds APPROVED requests against media with PROCESSING and AVAILABLE status on both standard and 4K status columns, asserting the join path is correct for both `is4k = false` and `is4k = true`

To run:

```sh
node server/test/index.mts server/routes/issue.test.ts
node server/test/index.mts server/routes/request.count.test.ts
```

N/A -- backend-only change with no UI impact.

## Checklist:

- [x] I have updated the documentation accordingly. (no documentation changes required for this fix)
- [x] Database migration (if required) (no schema changes)

## Summary by CodeRabbit

* **Tests**
  * Added integration tests for issue and request count endpoints, validating totals by type and status, processing vs available counts (including 4K scenarios), and zero-state responses.

* **Refactor**
  * Count endpoints rewritten to run independent count operations concurrently for the same response fields and behavior, improving query isolation and efficiency.


dougrathbone:
Fixed in the latest commit:

- **4K/processing/available test coverage**: added a fourth test that seeds APPROVED requests for HD media at PROCESSING status, HD media at AVAILABLE status, 4K media at PROCESSING status4k, and 4K media at AVAILABLE status4k -- then asserts `processing === 2` and `available === 2`. This directly exercises the `is4k` branch in the join query.
- **Duplicated QueryBuilder**: extracted `approvedMediaStatusCount(isAvailable: boolean)` inline in the route handler -- both `processing` and `available` now call it with `false`/`true` respectively.
- **Settings leak + secure cookie**: applied the same `authenticatedAgent` save/restore fix and explicit `cookie: { secure: false }` as raised in the other PRs.

server/routes/issue.test.ts
```diff
@@ -0,0 +1,163 @@
...
+
+async function authenticatedAgent(email: string, password: string) {
+  const agent = request.agent(app);
+  const settings = getSettings();
+  settings.main.localLogin = true;
+
+  const res = await agent.post('/auth/local').send({ email, password });
+  assert.strictEqual(res.status, 200);
+  return agent;
+}
```
fallenbagel, line 65: same issue here as i flagged in your other PR.

server/routes/request.ts
```diff
@@ -338,83 +338,62 @@ requestRoutes.post<never, MediaRequest, MediaRequestBody>(
 requestRoutes.get('/count', async (_req, res, next) => {
   const requestRepository = getRepository(MediaRequest);
 
-  try {
-    const query = requestRepository
+  const approvedMediaStatusCount = (isAvailable: boolean) =>
```
fallenbagel: This helper builds a `QueryBuilder` but lives outside the try block. I would recommend moving it inside so any unexpected throw is caught by the shared error handler.

server/routes/request.test.ts
fallenbagel, line 1: This should be `request.test.ts` the /count route is part of request route, and splitting test files per endpoint will get unwieldy.

fallenbagel (changes requested):

dougrathbone:
Addressed the feedback:

- Moved the QueryBuilder helper inside the try block so unexpected throws are caught by the error handler
- Renamed `request.count.test.ts` to `request.test.ts` to keep tests grouped by route

gauthier-th:
Closing as discussed [here](https://github.com/seerr-team/seerr/pull/2806#issuecomment-4276993747).

PR #2832, opened 2026-04-06
feat: AdvancedRequestModal preselects the calculated routing rule on create

Mczubi:
### Reasoning
I discovered that there was an PR open which would solve most of the issues i have with assigning the correct service / profile automatically.
Testing this feature i saw that for non admin users, restricted to making only requests, this worked perfectly.
The rules were calculated correctly and the request could get approved by an admin.

But when using an administrative user, the advanced request modal would always fallback to the initial settings.
In other words, it did not use the calculation functionality introduced by the rules routing feature.
As this was a major letdown, i decided to implement a feature that would allow an admin user to also have the calculated default preselected, but still granting him the option to change the reasoning of the automation.

### Detail
Implemented a new route GET /api/v1/request/resolved-route which returns the calculated output of
the routing rule using the same logic as the post request route.
This route then is used by the request modal to determine the correct
initial setting for a newly created request.
This grants admin users the same quality of life feature as a non admin one has.

This PR is a follow-up to PR #2452. 
It adds preselection for advanced requests which depends on the changes introduced there.

## How Has This Been Tested?
Tests were performed manually by performing the request creation action with both admin and non admin user.
Test instance was connected to my local running Sonarr / Radarr instances and ran against both of them.

Each new request created through my test instance landed on the correct Instance either chosen by the auto calculated rules or by the manual override performed by an admin user.

No affect was noticed on any other component throughout testing

## Checklist:

        Ran cypress and all seemed fine

### DIFF
--- a/seerr-api.yml
+++ b/seerr-api.yml
@@ -6356,6 +6356,65 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/MediaRequest'
+  /request/resolved-route:
+    get:
+      summary: Get the default resolved route for a request
+      description: |
+        Returns the first route used when a new request, for this media, would be inserted.
+      tags:
+        - request
+      parameters:
+        - in: query
+          name: userId
+          schema:
+            type: integer # Changed from number to integer
+            nullable: true
+            example: 20
+        - in: query
+          name: is4k
+          schema:
+            type: boolean
+            nullable: true
+            example: true
+        - in: query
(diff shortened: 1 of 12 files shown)


fallenbagel:
The routing rules pr will be rebased for https://github.com/seerr-team/seerr/pull/2164 which literally adds the same support

This PR is the same thing as #2164

Issue #2842, opened 2026-04-08
Request Available - Expand to single episode releases


pinappleeyes:
### Description

Hey hows it going? 

I think it would be a really helpful addition to have per episode release notifications in seerr. Right now, the only way I'll get a notification for episodes is when the entire searson is done airing and all added. If you could maybe add a setting to allow for single episodes to have a notification sent out, that would be awesome.

### Desired Behavior

- Single episode is added to jellyfin
- seerr notices this and sends out a discord (or other platform) notification about said episode 

### Search Existing Issues

- [x] Yes, I have searched existing issues.

### Code of Conduct

- [x] I agree to follow Seerr's Code of Conduct

Issue #2242, opened 2025-12-20
Notifications for pilot/first episode of a season


HamletDuFromage:
### Description

For an upcoming show, the user that requested a season only gets notified after the season ended i.e. weeks after the first episode aired.

### Desired Behavior

For a given season, if the finale doesn't air the same day as the pilot, a notification is sent after the pilot has been imported to jellyfin.

### Code of Conduct

- [x] I agree to follow Seerr's Code of Conduct

HamletDuFromage:
nvm, essentially a duplicate of https://github.com/seerr-team/seerr/issues/480

Issue #2027, opened 2025-10-10
[Feature Request] Allow episode-level availability notifications for shows currently airing


zachyale:
### Description

When users request a series of a show that is currently being released, there isn’t an option for notifying when the request has been partially made available- a notification is only sent when the season is fully available. It would be nice if for shows that are currently airing, there were an option to receive notifications as each episode is made available.

### Desired Behavior

A new notification option, to enable episode availability notifications for seasons that are still airing. Ideally there would be a way for users to sign on to also “watch” notifications for a given show’s episode updates, even if they weren’t the user that made the request. End result, a subset of users can be pinged on discord as newly released episodes are made available.

### Code of Conduct

- [x] I agree to follow Seerr's Code of Conduct

gauthier-th:
Closing this as duplicate of #480

PR #2869, opened 2026-04-13
feat(notifications): add per-episode availability notifications

### LINKED ISSUE #480: Ability to send notifications for new episodes of ongoing seasons

eunux:
## Description

Adds per-episode availability notifications for requested TV series. Right now
Seerr only notifies you when a request becomes available at the media or season
level, which means if you're following a weekly show you don't hear anything
until the entire season finishes downloading. This adds a more granular
notification so users can find out when newly aired episodes become available.

The use case is basically:

- I requested a show thats airing weekly
- I want to know when each new episode lands

- Fixes #480
- Related: #891, #1361, #2027, #2242, #2842

This depends partially on the Episode entity work from #1671. The entity and
migration pieces are included here becuase the notification flow needs
episode-level availability records in the data model. The underlying Episode
tracking approach comes from #1671 and this branch is meant to be reviewed as
a dependent follow-up built on that work.

I could have implemented this as a standalone thing without the Episode entity
but #1671 already has the groundwork for episode-level state, so building on
that seemed like the lowest-variance option — keeps the review focussed on the
user-facing behaviour here rather than introducing a competing model for
episode tracking.

Happy to do either of these if maintainers prefer:

- rebase once #1671 lands
- split into a smaller follow-up that assumes the Episode entity has already merged

**How the notifications work:**

- **Weekly/staggered shows:** Each newly aired episode triggers its own
  notification as it becomes available. This is the main use case.
- **Batch drops (Netflix-style):** If every episode in a season shares the same
  air date on TMDB, per-episode notifications are suppressed and the existing
  "Request Available" notification handles it instead.
- **Premiere batches + weekly:** Shows that drop 2-3 episodes on premiere day
  then go weekly will still send per-episode notifications (the season has
  multiple distinct air dates). You get a small burst on premiere day then
  one per week after that. This is intentional.
- **Old episodes:** Only episodes that aired after the users request was
  created will trigger notifications. Bulk imports, library re-scans, and
  backfills are silently filtered out.
- **Overlap with "Request Available":** You can enable both. "Request Available"
  fires once when the entire request completes. "Newly Aired Episode Available"
  fires per-episode as they arrive. The UI descriptions clarify the difference.

**Season entity note:** The Season entity's `updatedAt` column was changed from
TypeORM's `UpdateDateColumn` to `DbAwareColumn` to match the Episode entity's
timestamp pattern and stay consistent with #1671. Not a behavioural change for
the notification feature itself.

**Unique constraint:** The final commit adds `@Unique(['season', 'episodeNumber'])`
to the Episode entity to prevent duplicate rows from concurrent scans. Same
pattern as `Blocklist`, `UserPushSubscription`, and `Watchlist`. This commit
can be dropped if maintainers would rather land it with #1671.

**AI Disclosure:** Used AI tooling to help inspect the codebase and organise
the work. Changes were reviewed manually, PR description is my own words.

## How Has This Been Tested?

- `pnpm lint`
- `pnpm test`
- `pnpm typecheck`
- `pnpm build`
- `pnpm i18n:extract`
- `pnpm format:check`
- Manually verified the notification type shows up in settings
- Checked the episode notification trigger logic against current code paths

Seerr User Settings Screenshot

## Checklist:


fallenbagel:
This should wait till #1671

edit: sorry I just saw that you mentioned it in description too 😅

PR #2478, opened 2026-02-17
fix: preserve blocklist on media deletion & optimise watchlist-sync

fallenbagel:
## Description
When a user blocklists media and then removes it from Sonarr/Radarr via the Seerr interface, the Media entity and its associated Blocklist record were both destroyed due to the cascade delete on the relationship. On the next watchlist sync cycle, no record existed to check against, so a fresh request was created and auto-approved, completely bypassing the blocklist.

This PR fixes the media delete endpoint to detect blocklisted status and preserve the entity by nulling out service-related fields rather than removing the row entirely. It also improves the watchlist sync filtering to exclude blocklisted media as a safety net and to skip items where the user already has an existing auto-request, eliminating the repeated unnecessary api calls and database lookups that were happening every sync cycle just to throw a duplicate request error.

- Fixes #2475 

## Checklist:

- [X] I have read and followed the contribution [guidelines](https://github.com/seerr-team/seerr/blob/develop/CONTRIBUTING.md).
- [ ] Disclosed any use of AI (see our [policy](https://github.com/seerr-team/seerr/blob/develop/CONTRIBUTING.md#ai-assistance-notice))
- [ ] I have updated the documentation accordingly.
- [X] All new and existing tests passed.
- [X] Successful build `pnpm build`
- [ ] Translation keys `pnpm i18n:extract`
- [ ] Database migration (if required)

## Release Notes

* **Bug Fixes**
  * Watchlist sync now excludes items the user has already auto-requested, reducing duplicate requests.
  * Deleting a blocklisted media item now preserves the record while clearing external/service identifiers instead of removing it.
  * Watchlist sync still respects user permissions and settings; existing error handling for request creation remains unchanged.

PR #2880, opened 2026-04-14
fix: remove orphaned media entries when all requests are deleted

Kenshin9977:
## Description

When a user deletes the last request associated with a media entry, the media record is left orphaned in the database with status `UNKNOWN`. This prevents users from submitting a new request for the same title since Seerr still treats it as an existing item.

`handleRemoveParentUpdate()` now removes the media entry entirely when no requests remain and the media is not in an active state (`AVAILABLE`, `PARTIALLY_AVAILABLE`, or `PROCESSING`). Media that is already available on disk is never removed.

Also aligns the status-reset branch with the same `ACTIVE_STATUSES` set to avoid resetting `PROCESSING` or `PARTIALLY_AVAILABLE` media to `UNKNOWN`.

## How Has This Been Tested?

Manual testing on a self-hosted instance with orphaned media entries. Verified:
- Deleting the last request on non-available media removes the media entry
- Re-requesting the same title works after deletion
- Media with status AVAILABLE is preserved
- Media with remaining requests (standard or 4k) is not affected

## Checklist:

> This pull request was authored with AI assistance.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of orphaned media by implementing proper cleanup when no active requests remain.

* **Refactor**
  * Optimized request status checking to reduce redundant scans.
  * Enhanced status tracking logic for media items.

### DIFF
--- a/server/subscriber/MediaRequestSubscriber.ts
+++ b/server/subscriber/MediaRequestSubscriber.ts
@@ -42,6 +42,11 @@ const sanitizeDisplayName = (displayName: string): string => {
 
 @EventSubscriber()
 export class MediaRequestSubscriber implements EntitySubscriberInterface<MediaRequest> {
+  private static readonly ACTIVE_STATUSES = new Set([
+    MediaStatus.AVAILABLE,
+    MediaStatus.PARTIALLY_AVAILABLE,
+    MediaStatus.PROCESSING,
+  ]);
   private async notifyAvailableMovie(
     entity: MediaRequest,
     event?: UpdateEvent<MediaRequest>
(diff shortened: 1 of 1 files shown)


fallenbagel:
Isnt this #2757?

fallenbagel:
Hey, thanks for the PR but this is already handled on `develop`.

When a request is deleted via the [request delete endpoint](https://github.com/seerr-team/seerr/blob/develop/server/routes/request.ts#L601-L621), `requestRepository.remove()` fires the TypeORM [`afterRemove` subscriber](https://github.com/seerr-team/seerr/blob/develop/server/subscriber/MediaRequestSubscriber.ts#L1049-L1058), which calls [`handleRemoveParentUpdate()`](https://github.com/seerr-team/seerr/blob/develop/server/subscriber/MediaRequestSubscriber.ts#L946-L978). That method loads the media with all the remaining requests, and checks if there are no standard/4K requests left, and resets the status to `UNKNOWN` if the media isn't `AVAILABLE`. This is exactly what lets users re-request the title.

Additionally, #2478 already hardened which also fixes the exact issue you mention as now when the [media delete endpoint](https://github.com/seerr-team/seerr/blob/develop/server/routes/media.ts#L182) is hit, it preserves the blocklisted media by calling [`resetServiceData()`](https://github.com/seerr-team/seerr/blob/develop/server/entity/Media.ts#L203-L214) instead of hard-deleting the row and specifically because the [`cascade: ['insert', 'remove']`](https://github.com/seerr-team/seerr/blob/develop/server/entity/Media.ts#L110-L113) on the requests relation would cascade-delete associated blocklist entries.

There are also a couple of concerns with the changes here as well:

1. **Hard-deleting the media row** (`manager.remove(fullMedia)`) is destructive. This is a **big problem**. It throws away seasons, external service IDs, and other metadata that resetting to `UNKNOWN` preserves. It would also cascade-delete any associated blocklist entries, which is the exact bug #2478 fixed.

2. **The `ACTIVE_STATUSES` guard** would actually make things worse too by skipping `PROCESSING` and `PARTIALLY_AVAILABLE`, it would *prevent* the status reset to `UNKNOWN` when the last request is deleted. If there are no requests left, there's nothing to process, so leaving media stuck in `PROCESSING` is the exact orphaned state users will complain about.

Please test on `develop` first as this should already be working correctly already. I've tested it myself and have also written a supertest test case that verifies this behaviour. As it stands, these changes would actually break the existing logic rather than fix it. If you are hitting a specific case where media gets stuck after request deletion on `develop`, please open a GitHub issue with reproduction steps so we can investigate properly then we can decide on how to go about it. 

But as it stands, this PR introduces regressions in the existing logic for a problem that should already be fixed on develop, so I'll be closing it.

PR #2903, opened 2026-04-18
fix(override-rules): use find() instead of index lookup for service resolution

fallenbagel:
## Description
This PR is fixes a bug in the Override Rules modal where dropdowns (Root Folder, Quality Profile, Tags) would remain greyed out when editing an existing rule or selecting a service. The root cause was that `radarrServices` and `sonarrServices` were being looked up by array index using the service ID, which only worked by coincidence when the ID matched the index and most commonly broken after a user deletes and re-adds a service, causing the ID to increment while the array index stays at 0. Replaced the index lookups with `.find()` to correctly resolve services by their ID. 

## How Has This Been Tested?
- This has been tested by the reported user and they have confirmed that it works:
image

## Checklist:

## Summary by CodeRabbit

* **Refactor**
  * Improved service lookup mechanism in settings to enhance reliability and maintainability.

### DIFF
--- a/src/components/Settings/OverrideRule/OverrideRuleModal.tsx
+++ b/src/components/Settings/OverrideRule/OverrideRuleModal.tsx
@@ -118,20 +118,15 @@ const OverrideRuleModal = ({
   );
 
   useEffect(() => {
-    if (
-      rule?.radarrServiceId !== null &&
-      rule?.radarrServiceId !== undefined &&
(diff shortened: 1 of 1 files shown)


gauthier-th (approved):
Success

fallenbagel:
Accidentally closed :sweat_smile:

0xSysR3ll (approved):

PR #2962, opened 2026-04-27
fix(discover): preserve keyword filter input focus after selection

### LINKED ISSUE #2830: WEBUI QoL - Filter text fields lose focus after entering a single filter
### Description

One thing that always bugs me when I'm using filters in Seerr is how the text fields lose focus as soon as you add a tag/keyword

### Desired Behavior

Selecting the keyword/exclude keyword/tag field then adding an entry shouldn't cause the field to lose focus, so you can start typing the next entry you want to add straight away

### Additional Context

_No response_

### Search Existing Issues

### Code of Conduct

0xSysR3ll:
## Description

The PR fixes an annoying issue in Discover where the keyword and exclude-keyword filters would lose focus every time you selected a value. This happened because the selector component was being remounted on each update.

- Fixes #2830

## How Has This Been Tested?

Typed a keyword, hit enter, repeated without loosing input focus.

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Keyword selector now reliably syncs selection state between single- and multi-select modes and with external handlers.
  * Default-value handling improved: missing defaults clear the current selection; when defaults are present they load correctly without forcing component remounts, and user choices update state immediately before external change handlers are called.

[Review Change Stack](https://app.coderabbit.ai/change-stack/seerr-team/seerr/pull/2962?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/components/Selector/index.tsx
+++ b/src/components/Selector/index.tsx
@@ -295,7 +295,7 @@ export const KeywordSelector = ({
 }: BaseSelectorMultiProps | BaseSelectorSingleProps) => {
   const intl = useIntl();
   const [selectedValue, setSelectedValue] = useState<
-    MultiValue<SingleVal> | SingleValue<SingleVal> | null
+    MultiValue<SingleVal> | SingleValue<SingleVal>
   >(null);
 
   useEffect(() => {
--- a/src/components/Selector/index.tsx
+++ b/src/components/Selector/index.tsx
@@ -295,7 +295,7 @@ export const KeywordSelector = ({
 }: BaseSelectorMultiProps | BaseSelectorSingleProps) => {
   const intl = useIntl();
   const [selectedValue, setSelectedValue] = useState<
-    MultiValue<SingleVal> | SingleValue<SingleVal>
+    MultiValue<SingleVal> | SingleValue<SingleVal> | null
   >(null);
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


src/components/Selector/index.tsx
```diff
@@ -294,13 +294,14 @@ export const KeywordSelector = ({
   onChange,
 }: BaseSelectorMultiProps | BaseSelectorSingleProps) => {
   const intl = useIntl();
-  const [defaultDataValue, setDefaultDataValue] = useState<
-    { label: string; value: number }[] | null
+  const [selectedValue, setSelectedValue] = useState<
+    MultiValue<SingleVal> | SingleValue<SingleVal>
```
M0NsTeRRR:
```suggestion
    MultiValue<SingleVal> | SingleValue<SingleVal> | null
```

null should be kept ?

M0NsTeRRR (changes requested):

src/components/Selector/index.tsx
0xSysR3ll: It may be misleading, but it's technically redundant since `SingleValue<Option>` is exported as `Option | null`

M0NsTeRRR: For me even if SingleValue already includes null, writing `MultiValue<SingleVal> | SingleValue<SingleVal> | null` makes the contract obvious, prevents silent regressions if SingleValue changes 😄

M0NsTeRRR (dismissed):

src/components/Selector/index.tsx
gauthier-th: +1, keeping ` | null` avoid confusion

M0NsTeRRR (approved):

gauthier-th (approved):

PR #2983, opened 2026-05-01
feat(user-settings): add a per-user media language setting

### LINKED ISSUE #2308: Separate UI language from media data language
### Description

Currently we only have 1 settings for languange, this is control both the app language and media metadata language. 

Would be nice if this setting would be split, so user can choose in which language he wants to view the UI and which language use to display media info.

### Desired Behavior

Split "Display Language" parameter to 2 comboboxes, one for "UI Language" second for "Media Language"

### Additional Context

This is required since UI localization in bad for some languages, so I want to view ui in english, but movie description in my native language.

### Search Existing Issues

### Code of Conduct

0xSysR3ll:
## Description

This PR introduces a clearer separation between user language preferences by distinguishing the **Display Language** used for the UI and the **Media Language** used for media metadata.

To preserve backward compatibility, the content language automatically falls back to the effective locale when it hasn't been explicitly set.

- Fixes #2308

## How Has This Been Tested?

- Set Display Language to **English**
- Set Media Language to **French**

UI is correctly using **English** as display language whereas media related metadata is displayed in **French**.   

### User settings
image

### Result
image

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Per-user "Media Language" setting to control the language used for media titles, overviews, and descriptions; falls back to display language if unset.
  * UI: "Media Language" selector added to user settings.

* **Backend / API**
  * Setting persisted and exposed via the API; used as the default for media-related endpoints when no language query is provided.

* **Documentation**
  * User docs updated to describe the Media Language override.

* **Chores**
  * Database migrations added to store the new setting.


seerr-api.yml
gauthier-th, line 1: Looks like there is a linter/formatter issue here?

gauthier-th (changes requested):
As always, DB migrations will have to be re-generated before we merge this.

seerr-api.yml
0xSysR3ll: Wow didn't see this, that's strange.

gauthier-th (commented):
I think the mediaLocale should be initially populated with the current locale. Otherwise this will "reset" the locale previously used to display media info.

Anarh2404:
Tested this PR on a live instance (built from `d70c191`, Jellyfin media server, existing user base copied from my production install):

- Works as expected on movie/TV detail pages, search and genre listings — with Display Language = English and Media Language = Русский I get English UI with Russian titles/overviews. 👍
- The read-time fallback (`mediaLocale || locale`) also works: users who never set Media Language keep their current behavior, even though the migration doesn't backfill the column.

One gap I noticed while testing: **the Discover pages still follow the display language**. The PR converts `req.locale` → `mediaLocale` in `movie.ts`, `tv.ts`, `person.ts`, `search.ts` and the genre routes, but `discover.ts` (16 `language:` call sites — sliders, trending, upcoming, genre/studio/network pages), `collection.ts` (1) and `blocklist.ts` (2) still pass `req.locale`. So with EN UI + RU media language, a movie shows a Russian title/overview on its detail page but an English one in every Discover row — which is the surface most users browse. If it helps, I can list the exact line numbers.

Minor note: the conversion also changed `?? req.locale` to `|| mediaLocale`, so an explicitly empty `?language=` query param now falls through to the user's media locale instead of being passed as-is — probably fine (arguably an improvement), just flagging the semantic change for third-party API clients.

*AI disclosure: this comment was written by AI (Claude Code) — my English is not good enough to write it myself. The live testing and code reading behind it are my own, and replies will likely be AI-assisted in the same way.*

0xSysR3ll:
> One gap I noticed while testing: **the Discover pages still follow the display language**. The PR converts `req.locale` → `mediaLocale` in `movie.ts`, `tv.ts`, `person.ts`, `search.ts` and the genre routes, but `discover.ts` (16 `language:` call sites — sliders, trending, upcoming, genre/studio/network pages), `collection.ts` (1) and `blocklist.ts` (2) still pass `req.locale`. So with EN UI + RU media language, a movie shows a Russian title/overview on its detail page but an English one in every Discover row — which is the surface most users browse. If it helps, I can list the exact line numbers.

Nice catch! I wasn't focused on this when I made the PR, I'll add this.

> Minor note: the conversion also changed ?? req.locale to || mediaLocale, so an explicitly empty ?language= query param now falls through to the user's media locale instead of being passed as-is — probably fine (arguably an improvement), just flagging the semantic change for third-party API clients.

As you said, minor but worth the fix.

Thanks for testing :)

Anarh2404:
Rebuilt and retested from `8d024fc` (same live setup as before): the gap is fixed — Discover sliders/trending/upcoming, genre/studio/network pages, collections and the blocklist all follow Media Language now. With Display Language = English and Media Language = Русский the detail pages and every Discover row agree, and clearing Media Language falls back to the display language as before. Nothing else from my previous report is outstanding. Thanks for the quick turnaround!

One more observation from live testing: for a title with no Russian translation on TMDB at all (e.g. TV 204269), the detail page mixes languages — the title falls back to the original Japanese name (TMDB's own fallback for `name`), while the overview falls back to English via the existing `if (!data.overview)` refetch in the routes. It might be nicer to fall back consistently — both title and overview to the display language (or at least to the same language) when the media language has no translation. Not a regression of this PR (the asymmetry pre-dates it), but per-user Media Language makes it much more visible.

*AI disclosure: this comment was written by AI (Claude Code) — my English is not good enough to write it myself. The live testing behind it is my own.*

0xSysR3ll:
> Rebuilt and retested from `8d024fc` (same live setup as before): the gap is fixed — Discover sliders/trending/upcoming, genre/studio/network pages, collections and the blocklist all follow Media Language now. With Display Language = English and Media Language = Русский the detail pages and every Discover row agree, and clearing Media Language falls back to the display language as before. Nothing else from my previous report is outstanding. Thanks for the quick turnaround!
> 
> One more observation from live testing: for a title with no Russian translation on TMDB at all (e.g. TV 204269), the detail page mixes languages — the title falls back to the original Japanese name (TMDB's own fallback for `name`), while the overview falls back to English via the existing `if (!data.overview)` refetch in the routes. It might be nicer to fall back consistently — both title and overview to the display language (or at least to the same language) when the media language has no translation. Not a regression of this PR (the asymmetry pre-dates it), but per-user Media Language makes it much more visible.
> 
> _AI disclosure: this comment was written by AI (Claude Code) — my English is not good enough to write it myself. The live testing behind it is my own._

Thanks again for the thorough testing (and the example, that will help)! I'll see what I can do about this.

Anarh2404:
Rebuilt and retested from `75d5dc7` (same live setup): the consistent fallback works — the title from my previous comment (TV 204269, no Russian translation on TMDB at all) now shows the display-language title and overview together instead of the Japanese/English mix. 👍

One regression though: the `!data.overview` trigger now also overwrites titles that **are** translated. Partially translated entries are common on TMDB — name localized, overview missing. Example: TV 34860 — the Russian translation has a name («Мальчик-ниндзя Рантаро») but no overview. On `8d024fc` the detail page showed the Russian name + English overview; on `75d5dc7` the fallback replaces the name too, so the existing Russian title is lost ("Ninja Boy Rantaro"). Same for TV 90388 («Страсть на кончиках пальцев» → "Fire in His Fingertips").

An implementation idea that fixes this exactly and also removes the extra request: append `translations` to the details call (`append_to_response=...,translations`). TMDB returns all translations in the same response, so the route can build the fallback per field locally — keep `name` from the media language when its translation has one, take `overview` from the display language (then `en`) — and the current second full details fetch (which re-downloads credits etc. for every untranslated title) goes away entirely. Cost: I measured the worst case on Breaking Bad (68 translations) with this PR's full `append_to_response` set — +29 KB raw / +13 KB gzipped on top of a ~200 KB response, and it's cached for 12 h like the rest. If that feels too invasive, a minimal guard would also work: only take `name`/`title` from the fallback response when the localized value equals `original_name`/`original_title`.

Also worth noting (pre-existing, not a regression of this PR): the fallback only covers the detail routes — Discover/search rows and the title-card hover still render the raw list payload, so an untranslated title shows its original-language name and no overview in the row while its detail page falls back. List endpoints have no `append_to_response`, so fixing that would mean N+1 extra requests per row — probably fine to leave as a known limitation, just flagging the inconsistency.

*AI disclosure: this comment was written by AI (Claude Code) — my English is not good enough to write it myself. The live testing and measurements behind it are my own.*

0xSysR3ll:
> One regression though: the !data.overview trigger now also overwrites titles that are translated. Partially translated entries are common on TMDB — name localized, overview missing. Example: TV 34860 — the Russian translation has a name («Мальчик-ниндзя Рантаро») but no overview. On 8d024fc the detail page showed the Russian name + English overview; on 75d5dc7 the fallback replaces the name too, so the existing Russian title is lost ("Ninja Boy Rantaro"). Same for TV 90388 («Страсть на кончиках пальцев» → "Fire in His Fingertips").

Agreed. I'll bring back partial translations.

> An implementation idea that fixes this exactly and also removes the extra request: append translations to the details call (append_to_response=...,translations). TMDB returns all translations in the same response, so the route can build the fallback per field locally — keep name from the media language when its translation has one, take overview from the display language (then en) — and the current second full details fetch (which re-downloads credits etc. for every untranslated title) goes away entirely. Cost: I measured the worst case on Breaking Bad (68 translations) with this PR's full append_to_response set — +29 KB raw / +13 KB gzipped on top of a ~200 KB response, and it's cached for 12 h like the rest. If that feels too invasive, a minimal guard would also work: only take name/title from the fallback response when the localized value equals original_name/original_title.

IMHO implementing this `append_to_response=translations` thing would be a bigger change, so I'd rather keep that for a follow-up instead of making this PR too large.

I also agree about the lists/search stuff, but that's an inconsistency that already exists.

> Also worth noting (pre-existing, not a regression of this PR): the fallback only covers the detail routes — Discover/search rows and the title-card hover still render the raw list payload, so an untranslated title shows its original-language name and no overview in the row while its detail page falls back. List endpoints have no append_to_response, so fixing that would mean N+1 extra requests per row — probably fine to leave as a known limitation, just flagging the inconsistency.

List endpoints can't use `append_to_response`, so fixing that would mean extra requests for every item, which feels out of scope for this PR.

Detail pages are where the missing overview hurts the most anyway, so I'd rather focus on that for now.

Anarh2404:
Rebuilt and retested from `78deee5`: all three cases from my previous comment now behave correctly — fully untranslated titles fall back consistently, partially translated ones keep their localized names with the fallback overview, and fully translated titles are untouched. The regression is fixed and nothing is outstanding from my side. Thanks for the quick fixes!

*AI disclosure: this comment was written by AI (Claude Code) — my English is not good enough to write it myself. The live testing behind it is my own.*

PR #3017, opened 2026-05-11
feat: add a demo feature

gauthier-th:
## Description

This PR introduces a new demo feature when the environment variable `UNSAFE_DO_NOT_USE_DEMO` is set to `true`:
- The username / password of a non-admin local user is pre-filled when logging in.
- Demo users are unable to change their password or user info.
- The data is reset every hour.
- It renames the `friend@seerr.dev` test user into `demo@seerr.dev`
- It exports the `seedTestUsers` function of `server/utils/seedTestDb.ts`

This can only be run if the database of the project is empty.

## How Has This Been Tested?

N/A

N/A

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Demo mode available with login fields prefilled and a seeded demo user; demo flag exposed to the frontend.

* **Improvements**
  * Hourly automatic resets refresh demo data.
  * Demo accounts have restricted/short-circuited sign-in, password, settings, and external media unlinking behaviors for safety.

* **Tests**
  * Test fixtures and Cypress configs updated to use demo account credentials.

[Review Change Stack](https://app.coderabbit.ai/change-stack/seerr-team/seerr/pull/3017?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -9,7 +9,7 @@ export default defineConfig({
   env: {
     ADMIN_EMAIL: 'admin@seerr.dev',
     ADMIN_PASSWORD: 'test1234',
-    USER_EMAIL: 'friend@seerr.dev',
+    USER_EMAIL: 'demo@seerr.dev',
     USER_PASSWORD: 'test1234',
   },
   retries: {
--- a/cypress/e2e/discover.cy.ts
+++ b/cypress/e2e/discover.cy.ts
@@ -145,7 +145,7 @@ describe('Discover', () => {
           requestedBy: {
             permissions: 4194336,
             id: 18,
-            email: 'friend@seerr.dev',
+            email: 'demo@seerr.dev',
             plexUsername: null,
             username: '',
             recoveryLinkExpirationDate: null,
@@ -159,7 +159,7 @@ describe('Discover', () => {
             createdAt: '2022-08-17T04:55:28.000Z',
             updatedAt: '2022-08-17T04:55:28.000Z',
             requestCount: 1,
-            displayName: 'friend@seerr.dev',
+            displayName: 'demo@seerr.dev',
           },
           seasonCount: 0,
         },
--- a/server/routes/auth.test.ts
+++ b/server/routes/auth.test.ts
@@ -189,7 +189,7 @@ describe('POST /auth/local', () => {
   it('allows the non-admin user to log in', async () => {
     const res = await request(app)
       .post('/auth/local')
-      .send({ email: 'friend@seerr.dev', password: 'test1234' });
+      .send({ email: 'demo@seerr.dev', password: 'test1234' });
 
     assert.strictEqual(res.status, 200);
     assert.ok('id' in res.body);
--- a/server/routes/issue.test.ts
+++ b/server/routes/issue.test.ts
@@ -100,7 +100,7 @@ describe('POST /issue', () => {
     const userRepo = getRepository(User);
     const media = await seedMedia();
     const friend = await userRepo.findOneOrFail({
-      where: { email: 'friend@seerr.dev' },
+      where: { email: 'demo@seerr.dev' },
     });
 
     const agent = await loginAs('admin@seerr.dev', 'test1234');
@@ -114,8 +114,8 @@ describe('POST /issue', () => {
     });
 
     assert.strictEqual(res.status, 201);
-    assert.strictEqual(res.body.createdBy.email, 'friend@seerr.dev');
-    assert.strictEqual(res.body.comments[0].user.email, 'friend@seerr.dev');
+    assert.strictEqual(res.body.createdBy.email, 'demo@seerr.dev');
+    assert.strictEqual(res.body.comments[0].user.email, 'demo@seerr.dev');
 
     const persisted = await issueRepo.findOneOrFail({
       where: { id: res.body.id },
@@ -144,13 +144,13 @@ describe('POST /issue', () => {
     const userRepo = getRepository(User);
     const media = await seedMedia();
     const friend = await userRepo.findOneOrFail({
-      where: { email: 'friend@seerr.dev' },
+      where: { email: 'demo@seerr.dev' },
     });
 
     friend.permissions = Permission.CREATE_ISSUES;
     await userRepo.save(friend);
 
-    const agent = await loginAs('friend@seerr.dev', 'test1234');
+    const agent = await loginAs('demo@seerr.dev', 'test1234');
     const res = await agent.post('/issue').send({
       issueType: IssueType.SUBTITLES,
       message: 'Subtitles are missing.',
@@ -159,15 +159,15 @@ describe('POST /issue', () => {
     });
 
     assert.strictEqual(res.status, 201);
-    assert.strictEqual(res.body.createdBy.email, 'friend@seerr.dev');
-    assert.strictEqual(res.body.comments[0].user.email, 'friend@seerr.dev');
+    assert.strictEqual(res.body.createdBy.email, 'demo@seerr.dev');
+    assert.strictEqual(res.body.comments[0].user.email, 'demo@seerr.dev');
   });
 
   it('prevents non-managers from supplying another userId', async () => {
     const userRepo = getRepository(User);
     const media = await seedMedia();
     const friend = await userRepo.findOneOrFail({
-      where: { email: 'friend@seerr.dev' },
+      where: { email: 'demo@seerr.dev' },
     });
     const admin = await userRepo.findOneOrFail({
       where: { email: 'admin@seerr.dev' },
@@ -176,7 +176,7 @@ describe('POST /issue', () => {
     friend.permissions = Permission.CREATE_ISSUES;
     await userRepo.save(friend);
 
-    const agent = await loginAs('friend@seerr.dev', 'test1234');
+    const agent = await loginAs('demo@seerr.dev', 'test1234');
     const res = await agent.post('/issue').send({
       issueType: IssueType.OTHER,
       message: 'Something else is wrong.',
--- a/server/routes/request.test.ts
+++ b/server/routes/request.test.ts
@@ -88,7 +88,7 @@ async function seedRequest(status = MediaRequestStatus.PENDING) {
   const requestRepo = getRepository(MediaRequest);
 
   const requestedBy = await userRepo.findOneOrFail({
-    where: { email: 'friend@seerr.dev' },
+    where: { email: 'demo@seerr.dev' },
   });
 
   const media = await mediaRepo.save(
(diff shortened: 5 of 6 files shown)


M0NsTeRRR (commented):
I'm not a big fan of adding demo-specific code into the application. I prefer handling it on the system side and running `seeTestDb.ts` as a script via a cron job or similar even if we can't configure default email and password in the login form.

gauthier-th:
> I prefer handling it on the system side and running `seeTestDb.ts` as a script via a cron job

That would mean a down time every hour to run this script.
It would also allow some API calls I blocked to avoid modification of the email or the password that would make the demo inaccessible until it's reset (and avoid silly modifications of the display name).

> even if we can't configure default email and password in the login form.

If we can't configure default email/password it greatly reduces the accessibility and the interest of such a feature.

This PR only adds a few lines to add a default email/password when demo is enabled, and keeps everything else separated in another file. It's no overhead and doesn't add any technical debt.

I'm not sure to see what's the issue here?

M0NsTeRRR:
> That would mean a down time every hour to run this script.

Not exactly. We could initialize a SQLite database, stop the service, replace the file, and restart it resulting in about 2 minutes of downtime.
Many applications use this approach. For example, I work with NetBox (a network tool), and they provide a demo environment. The data is reset nightly at 04:00 UTC (to reduce downtime issue).
https://demo.netbox.dev/


> It would also allow some API calls I blocked to avoid modification of the email or the password that would make the demo inaccessible until it's reset (and avoid silly modifications of the display name).

We could handle this on the database side by marking certain fields as read-only.

> If we can't configure default email/password it greatly reduces the accessibility and the interest of such a feature.

Yes and no, because users can still log in and see the interface. Instead, we could add a feature to display a customizable welcome message on the login page. For example, an admin could include a note like: "If you have any issues logging in, contact me here...". In our case, we could include the demo login credentials in that message.

> I'm not sure to see what's the issue here ?

You are adding demo features that could end up in production code, potentially leading to security issues. For example, creating default user credentials in an application is a bad practice, which is why many applications include a setup page to avoid this.

Security misconfiguration is ranked 2 in the [OWASP Top 10 2025](https://owasp.org/Top10/2025/A02_2025-Security_Misconfiguration/). This is why I avoid including such features in production code as much as possible.

gauthier-th:
> We could handle this on the database side by marking certain fields as read-only.

That seems very hard-coded and will very probably fail next time we update the database structure. My approach permits zero downtime, easily block admin users and unwanted changes, and will still works if the database structure changes.

> Yes and no, because users can still log in and see the interface.

Of course, but IMO this still reduces the accessibility a lot.

> You are adding demo features that could end up in production code, potentially leading to security issues.

This feature is strictly guarded by a if statement, there is no complicated logic that could be exploited. Nothing else is run if this simple condition is not fulfilled.

> Security misconfiguration is ranked 2 in the [OWASP Top 10 2025](https://owasp.org/Top10/2025/A02_2025-Security_Misconfiguration/). This is why I avoid including such features in production code as much as possible.

Yes. There are also a so many other ways to misconfigure the app, leading to security issue. That doesn't mean we remove all those features.
This feature will also not be added to any documentation anywhere, and it will only be available when setting a custom environment variable. There is no way for it to be enabled by someone by mistake.

@fallenbagel @0xSysR3ll any thoughts?

0xSysR3ll:
> @fallenbagel @0xSysR3ll any thoughts?

TBH I'm kind of between both opinions here, but I lean a bit more toward @M0NsTeRRR's concerns on the security side.

I totally agree that demos should stay simple and easy to use, and I don't think DB read-only workarounds are the right solution either. I also get that having downtime just to reset the demo would be annoying.

My main concern is that relying mostly on a single `DEMO_MODE=true` flag feels risky for something that can just "nuke" the instance".

I like the idea of having demo support, and the UX makes sense to me. I just think the destructive parts should be a bit more isolated and safer. Things like a dedicated demo DB/config for example, or refusing to start if real user data already exists, and safer reset logic would make me much more comfortable with it.

So overall I support the feature, but I'm not fully convinced by the current implementation yet.

gauthier-th:
> My main concern is that relying mostly on a single `DEMO_MODE=true` flag feels risky for something that can just "nuke" the instance".

This can't be done in any Docker environment. The apps need to be re-build entirely with the env `DEMO_MODE=true`. Just adding this env variable on a already built instances (i.e. everything we publish) will not work. It could only happen if users building the app from source have this environment variable enabled.

I could add an additional check making it impossible to start the demo mode if the app is already setup.

Happy to hear other suggestions.

0xSysR3ll:
> This can't be done in any Docker environment. The apps need to be re-build entirely with the env `DEMO_MODE=true`. Just adding this env variable on a already built instances (i.e. everything we publish) will not work. It could only happen if users building the app from source have this environment variable enabled.

I might have misunderstood this part then.

> I could add an additional check making it impossible to start the demo mode if the app is already setup.

Apart from this I don't have any other thoughts.

server/utils/demoMode.ts
```diff
@@ -0,0 +1,77 @@
...
+  setInterval(resetDemoData, 60 * 60 * 1000);
+
+  // Disable password sign-in for non-demo users
+  server.post('/api/v1/auth/local', async (req, res, next) => {
+    if (req.body?.email !== 'demo@seerr.dev') {
+      return res.status(403).json({ error: 'Password sign-in is disabled.' });
+    }
+    next();
+  });
+  // Disable password change for non-demo users
```
M0NsTeRRR: the comment is not accurate, it blocks every users

M0NsTeRRR (changes requested):
Except for the nitpick, still not a fan of the idea but LGTM

fallenbagel (changes requested):
Several unit tests are broken with this PR because they match by `friend@seerr.dev` instead of `demo`. You'll need to update those files

server/utils/demoMode.ts
```diff
@@ -0,0 +1,77 @@
...
+  server.post('/api/v1/user/:id/settings/main', async (_, res) => {
+    return res.status(200).json({});
+  });
+  // Disable unlinking media servers
+  server.post(
+    '/api/v1/user/:id/settings/linked-accounts/plex',
+    async (_, res) => {
+      return res.status(204).send();
+    }
+  );
```
fallenbagel, line 85: `/settings/password`, `/settings/main`, and `/settings/linked-accounts/plex` match `:id` unconditionally, so that would mean they would short-circuit for every user and not just the demo account. Wouldnt it be better to scope these to the demo user's id?

server/utils/demoMode.ts
```diff
@@ -0,0 +1,77 @@
...
+      await dataSource.getRepository(entity.name).clear();
+    }
+    // Add the default users
+    await seedTestUsers();
+    // Re-seed discover sliders
+    await DiscoverSlider.bootstrapSliders();
+  } catch (error) {
+    logger.error('Failed to reset demo data', { error });
+  }
+}
```
fallenbagel, line 29: `.clear()` iterates `entityMetadatas` with no dependency ordering. If any table has FK constraints without cascade, a truncate can throw mid-loop and it's caught and logged, but that leaves the reset partially applied and the hourly interval will keep failing the same way silently afterward. I would recomment truncating with cascade or ordering by dependency. Wdyt?

server/utils/demoMode.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export async function initDemoData(server: Express): Promise<void> {
+  // Check that no data exists in the database before seeding demo data
+  if ((await dataSource.getRepository('user').count()) > 0) {
+    logger.error('Database is not empty, cannot seed demo data. Aborting.');
+    process.exit(1);
+  }
```
fallenbagel, line 43: this empty-user-count guard only distinguishes "never initialized" from "already has data." Since `resetDemoData` excludes `session` and `user` from the hourly clear, user count stays > 0 permanently after the first successful boot. Any restart of the process after that point (like a crash, redeploy, host reboot) will hit this same check and `exit(1)`, since it can't tell an already-initialized demo apart from a real production DB with existing users. This should key off something set during init (a flag persisted in settings, or checking specifically for the seeded demo/admin accounts and nothing else) rather than a raw user count imo

fallenbagel (changes requested):

server/utils/demoMode.ts
gauthier-th: No? During demo mode, we don't allow any other users anyway.

fallenbagel: > No? During demo mode, we don't allow any other users anyway.

Hmm. True. I was thinking just incase for security measure but yeah both the media server auth endpoints would already return 500 once `mediaServerLogin` is false anyways on top of the local route only allowing `demo@seerr.dev`

server/utils/demoMode.ts
gauthier-th: True. I added a catch so that it throws an error only if `userCount !== 1` or if `user.email !== 'demo@seerr.dev'`.

server/utils/demoMode.ts
gauthier-th: I didn't see any issue during testing, but yes it could maybe be an issue.
Fixed it with `.delete({})` that will use the `onDelete: 'CASCADE'`.

server/utils/demoMode.ts
fallenbagel: > True. I added a catch so that it throws an error only if `userCount !== 1` or if `user.email !== 'demo@seerr.dev'`.

That doesn't quite work though. `userCount` never actually lands on 1 in either real state, since `seedTestUsers()` always writes `admin@seerr.dev` and `demo@seerr.dev` together. A genuinely fresh boot has `userCount` at 0, and a restart after a successful init will have it at 2. Both of those will trip `userCount !== 1` on their own, since it's an `||` and that clause eould short circuit before the email check ever runs. So this isn't scoped to some narrower failure case as it would throw on a fresh empty database and on every restart after the first one.

I think this needs to be something like checking that whatever users exist are exactly the two seeded accounts, rather than `!== 1`.

fallenbagel (changes requested):
https://github.com/seerr-team/seerr/pull/3017#discussion_r4043953626

server/utils/demoMode.ts
gauthier-th: Oh yes, didn't thought about that.
Would it works to remove the admin@seerr.dev just after seeding the test data?

gauthier-th: Or I could just check that the 2 only users are demo and admin

fallenbagel: > Or I could just check that the 2 only users are demo and admin

Do we need the admin in demo though. No right? So better we remove (just incase?)

fallenbagel (approved):

M0NsTeRRR (approved):

PR #3082, opened 2026-05-27
fix(email): respect hosts file for SMTP connections

### LINKED ISSUE #3020: /etc/hosts entry ignored when sending mail
### Description

Hello,
I have two FQDN added to my container via the "extra_hosts:" directive in my compose file.

plex.example.com
mail.example.com

When I use plex.example.com in the Plex section of the settings, it works like a charm.

When I use mail.example.com from the Notifications section of the settings, it doesn't work (read ECONNRESET).
But if I use the IP and tick the box "Allow Self-Signed Certificates" (IP is not in the certificate, so I have to check the box), the email is sent.

### Version

3.2.0

### Steps to Reproduce

1. Add "extra_hosts:" to the compose and add a FQDN that points to a mail server.
2. Use this FQDN in the Notifications section of the settings.

### Screenshots

_No response_

### Logs

```shell
Seerr logs (fake username) :

2026-05-11T18:57:13.120Z [debug][Notifications]: Sending email notification {"recipient":"random5302","type":"TEST_NOTIFICATION","subject":"Test Notification"}
2026-05-11T18:57:32.905Z [error][Notifications]: Error sending email notification {"recipient":"random5302","type":"TEST_NOTIFICATION","subject":"Test Notification","errorMessage":"read ECONNRESET"}

/etc/hosts content inside the container (IPs hidden and fake FQDNs)

/app $ cat /etc/hosts
127.0.0.1      localhost
::1            localhost ip6-localhost ip6-loopback
fe00::         ip6-localnet
ff00::         ip6-mcastprefix
ff02::1        ip6-allnodes
ff02::2        ip6-allrouters
192.168.X.X    mail.example.com
192.168.X.X    plex.example.com
172.X.X.X      seerr

Ping from inside the container (IP hidden and fake FQDN)

/app $ ping mail.example.com
PING mail.example.com (192.168.X.X): 56 data bytes
64 bytes from 192.168.X.X: seq=0 ttl=42 time=0.081 ms
64 bytes from 192.168.X.X: seq=1 ttl=42 time=0.057 ms
64 bytes from 192.168.X.X: seq=2 ttl=42 time=0.074 ms
64 bytes from 192.168.X.X: seq=3 ttl=42 time=0.089 ms
```

### Platform

desktop

### Database

SQLite (default)

### Device

Computer

### Operating System

Windows 11

### Browser

Firefox

### Additional Context

_No response_

### Search Existing Issues

### Code of Conduct

0xSysR3ll:
## Description

This PR fixes an issue where SMTP hostnames defined through Docker `extra_hosts` or `/etc/hosts` were not respected when sending email notifications.

Previously, nodemailer could resolve the SMTP hostname through DNS before opening the connection, which could bypass `/etc/hosts`.

- Fixes #3020 

## How Has This Been Tested?

Tested locally with Docker:

1. Built a seerr docker image from latest develop commit
2. Started a maildev server
3. Ran seerr with:
   ```bash
   --add-host example.com:host-gateway
   ```
4. Verified that` getent hosts example.com` inside the container resolved to the host gateway.
5. Confirmed the develop image did not reach the maildev server. In fact the container's
   `/etc/hosts` entry mapped `example.com` to the host gateway, but the SMTP
   client still followed nodemailer's DNS resolution path instead of the hosts
   entry.
6. Built a new image with this change applied.
7. Re-ran the same test and confirmed that:
   * the email send completed successfully;
   * the maildev server received and logged the message.

## Checklist:

## Summary by CodeRabbit

* **Chores**
  * Aligned generated type references with build output to improve type resolution and build reliability.
  * Hardened SMTP email connection handling: more robust socket handling, clearer timeout/error behavior, and safer handling when connecting to IP-based hosts to reduce delivery failures and timeouts.

### DIFF
--- a/server/lib/email/index.ts
+++ b/server/lib/email/index.ts
@@ -1,9 +1,43 @@
 import type { NotificationAgentEmail } from '@server/lib/settings';
 import { getSettings } from '@server/lib/settings';
 import Email from 'email-templates';
+import net from 'node:net';
 import nodemailer from 'nodemailer';
+import type SMTPTransport from 'nodemailer/lib/smtp-transport';
 import { URL } from 'url';
 import { openpgpEncrypt } from './openpgpEncrypt';
+
+const getSocket: SMTPTransport.Options['getSocket'] = (options, callback) => {
+  const socket = net.connect({
+    host: options.host ?? 'localhost',
+    port: options.port ?? 587,
+    localAddress: options.localAddress,
+  });
+  const cleanup = () => {
+    socket.setTimeout(0);
+    socket.removeListener('error', onError);
+    socket.removeListener('connect', onConnect);
+    socket.removeListener('timeout', onTimeout);
+  };
+  const onError = (error: Error) => {
+    cleanup();
+    callback(error, undefined);
+  };
+  const onConnect = () => {
+    cleanup();
+    callback(null, { connection: socket });
+  };
+  const onTimeout = () => {
+    cleanup();
+    socket.destroy();
+    callback(new Error('SMTP connection timed out'), undefined);
+  };
+
+  socket.once('error', onError);
+  socket.once('connect', onConnect);
+  socket.once('timeout', onTimeout);
+  socket.setTimeout(10000);
+};
 
 class PreparedEmail extends Email {
   public constructor(settings: NotificationAgentEmail, pgpKey?: string) {
@@ -28,6 +62,7 @@ class PreparedEmail extends Email {
               pass: settings.options.authPass,
             }
           : undefined,
+      getSocket: net.isIP(settings.options.smtpHost) ? undefined : getSocket,
     });
 
     if (pgpKey) {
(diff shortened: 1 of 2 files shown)


fallenbagel (dismissed):

server/lib/email/index.ts
```diff
@@ -1,10 +1,44 @@
...
 import nodemailer from 'nodemailer';
+import type SMTPTransport from 'nodemailer/lib/smtp-transport';
 import { URL } from 'url';
 import { openpgpEncrypt } from './openpgpEncrypt';
 
+const getSocket: SMTPTransport.Options['getSocket'] = (options, callback) => {
+  const socket = net.connect({
+    host: options.host ?? 'localhost',
+    port: options.port ?? 587,
+    localAddress: options.localAddress,
```
M0NsTeRRR: Shouldn’t host and port never be undefined here? Also, localAddress isn’t passed to createTransport, so it will always be undefined, right?

0xSysR3ll: 1. Yep, right. `host` and `port` should always be set. However Nodemailer types them as optional in the `getSocket` callback contract, hence the fallback.
2. And yes, `localAddress` can indeed be removed. Oversight on my end.

0xSysR3ll: As discussed in DMs with @M0NsTeRRR: we find it preferable to fail explicitly instead of falling back to a different SMTP server.

M0NsTeRRR (approved):

fallenbagel (approved):

PR #3102, opened 2026-06-02
fix(availability-sync): keep seasons when the media server is unreachable

### LINKED ISSUE #1729: Availability Sync marks seasons as "deleted" and shows as "partially available"
### Description

Hello,

I'm encountering an issue where TV shows are being incorrectly marked during the Availability Sync job. Specifically:

- TV shows are being marked as "partially available"

- Seasons are being marked as "deleted"

However, when I manually run a Sonarr scan afterward, the shows and seasons are correctly detected and marked as "available."

I'm unsure whether this is a bug or a potential misconfiguration on my end. 
Also, This issue does not occur with movies in Radarr, which are consistently detected correctly.

### Version

2.7.0

### Steps to Reproduce

1. Wait for the Availability Sync job to run
2. Observe TV shows being marked as "partially available" and seasons being marked as "deleted."
3. Run a manual Sonarr scan.
4. Observe correct "available" status afterward.

### Screenshots

_No response_

### Logs

```shell
2025-06-21T10:00:01.168Z [info][AvailabilitySync]: The non-4K show [TMDB ID 119100] still exists. Preventing removal.
2025-06-21T10:00:01.261Z [info][AvailabilitySync]: The non-4K season(s) [0,1] [TMDB ID 119100] was not found in any Sonarr and jellyfin instance. Status will be changed to deleted.
```

### Platform

desktop

### Database

SQLite (default)

### Device

PC

### Operating System

Windows 11

### Browser

Firefox

### Additional Context

_No response_

### Code of Conduct

OmarB97:
## Description

Fixes #1729.

This PR stops availability sync from marking TV seasons as deleted when Plex or Jellyfin is temporarily unreachable. If the media-server lookup fails with a non-404 error, seasons that were already available stay available instead of being treated as missing.

The net effect of this is that instead of Seerr showing the show that DOES exist on Plex as deleted (even though it's not deleted), NOW it will show as available (synced).

## How Has This Been Tested?

Added regression coverage for Plex and Jellyfin outage cases.

N/A

## Checklist:

### AI Disclosure

AI assistance was used while working on this PR. I reviewed the changes before submitting them.

## Release Notes

* **Bug Fixes**
  * Improved handling of TV season availability when media server connections fail with non-404 errors
  * Fixed season verification logic during availability sync to properly filter and process seasons

* **Tests**
  * Added regression tests for media server connection error scenarios

### DIFF
--- a/server/lib/availabilitySync.ts
+++ b/server/lib/availabilitySync.ts
@@ -937,30 +937,27 @@ class AvailabilitySync {
       }
     }
 
-    // Here we check each season in plex for availability
-    // If the API returns an error other than a 404,
-    // we will have to prevent the season check from happening
+    // Here we check each season in plex for availability.
+    // If the API returned an error other than a 404 we could not verify the
+    // seasons, so we assume the previously-available ones still exist instead
+    // of deleting media we simply failed to reach (mirrors the show-level guard).
     if (media.mediaType === 'tv') {
       const seasonsMap: Map<number, boolean> = new Map();
 
-      if (!preventSeasonSearch) {
-        const filteredSeasons = media.seasons.filter(
-          (season) =>
-            season[is4k ? 'status4k' : 'status'] === MediaStatus.AVAILABLE ||
-            season[is4k ? 'status4k' : 'status'] ===
-              MediaStatus.PARTIALLY_AVAILABLE
-        );
+      const filteredSeasons = media.seasons.filter(
+        (season) =>
+          season[is4k ? 'status4k' : 'status'] === MediaStatus.AVAILABLE ||
+          season[is4k ? 'status4k' : 'status'] ===
+            MediaStatus.PARTIALLY_AVAILABLE
+      );
 
-        for (const season of filteredSeasons) {
-          const seasonExists = await this.seasonExistsInPlex(
-            media,
-            season,
-            is4k
-          );
+      for (const season of filteredSeasons) {
+        const seasonExists = preventSeasonSearch
+          ? true
+          : await this.seasonExistsInPlex(media, season, is4k);
 
-          if (seasonExists) {
-            seasonsMap.set(season.seasonNumber, true);
-          }
+        if (seasonExists) {
+          seasonsMap.set(season.seasonNumber, true);
         }
       }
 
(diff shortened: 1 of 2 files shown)


server/lib/availabilitySync.ts
```diff
@@ -937,30 +937,27 @@ class AvailabilitySync {
...
-        for (const season of filteredSeasons) {
-          const seasonExists = await this.seasonExistsInPlex(
-            media,
-            season,
-            is4k
-          );
+      for (const season of filteredSeasons) {
+        const seasonExists = preventSeasonSearch
+          ? true
+          : await this.seasonExistsInPlex(media, season, is4k);
```
fallenbagel: Would prefer an early return here rather than threading `preventSeasonSearch` into each loop iteration. Imo, this should make it clearer and should avoid any latent bugs.
Something like:
```suggestion
if (preventSeasonSearch) {
  filteredSeasons.forEach((s) => seasonsMap.set(s.seasonNumber, true));
  return { existsInPlex, seasonsMap };
}
for (const season of filteredSeasons) { ... }
```

server/lib/availabilitySync.ts
```diff
@@ -1074,30 +1071,27 @@ class AvailabilitySync {
...
-        for (const season of filteredSeasons) {
-          const seasonExists = await this.seasonExistsInJellyfin(
-            media,
-            season,
-            is4k
-          );
+      for (const season of filteredSeasons) {
+        const seasonExists = preventSeasonSearch
+          ? true
+          : await this.seasonExistsInJellyfin(media, season, is4k);
```
fallenbagel: Same as above ^

```suggestion
if (preventSeasonSearch) {
  filteredSeasons.forEach((s) => seasonsMap.set(s.seasonNumber, true));
  return { existsInJellyfin, seasonsMap };
}
for (const season of filteredSeasons) { ... }
```

fallenbagel (changes requested):
Few changes requested.

Also a nitpick:
The inline comments explaining the test scenario are a bit much. The test name and the mock setup are self-explanatory enough. Would prefer to trim these comments down.

server/lib/availabilitySync.ts
OmarB97: Done in 8572f42e with the same shape on the Jellyfin/Emby path. The outage fallback now returns before the per-season loop, and the regular path only calls `seasonExistsInJellyfin`.

server/lib/availabilitySync.ts
OmarB97: Done in 8572f42e. I agree this reads better as an explicit bailout: after building `filteredSeasons`, the Plex path now fills `seasonsMap` and returns immediately when `preventSeasonSearch` is set. The normal loop is back to only calling `seasonExistsInPlex`.

M0NsTeRRR:
@OmarB97 Rule number 1 from our `CONTRIBUTING.md`.

> PR descriptions and all comments must be your own words. Do not paste LLM output as your PR description, review response, or issue comment. We want your understanding and explanation of the changes, not a machine-generated summary. An exception is made for LLM-assisted translation, however, note it explicitly if used.

Don’t copy/paste unnecessary sentences. The CI handles tests no need to share results, we won’t trust them.

OmarB97:
Removed the generated PR/comment text.

The bug is that a Plex/Jellyfin outage preserves the show but leaves the season map empty, so seasons get marked deleted. This PR keeps already-available seasons available when the media server lookup fails with a non-404 error.

Let me know if you have any other issues/concerns.

server/lib/availabilitySync.ts
```diff
@@ -937,30 +937,29 @@ class AvailabilitySync {
       }
     }
 
-    // Here we check each season in plex for availability
-    // If the API returns an error other than a 404,
-    // we will have to prevent the season check from happening
+    // Here we check each season in plex for availability.
```
fallenbagel, line 940: Why is this being removed? My previous comment was to trim the test file comments not these.

server/lib/availabilitySync.ts
```diff
@@ -1074,30 +1073,33 @@ class AvailabilitySync {
       }
     }
 
-    // Here we check each season in jellyfin for availability
-    // If the API returns an error other than a 404,
-    // we will have to prevent the season check from happening
+    // Here we check each season in jellyfin for availability.
```
fallenbagel, line 1078: Same as above

OmarB97:
addressed comments above. captured the full relevant context in each of those comments without being overly verbose.

fallenbagel (approved):

gauthier-th:
@0xSysR3ll could you please test if you can reproduce the root issue of this? And test if this PR fixes it if you can reproduce?

PR #3003, opened 2026-05-08
chore(deps): update dependency Next.js to v16

gauthier-th:
## Description

Update Next.js to v16. I ran the [codemod](https://nextjs.org/docs/app/guides/upgrading/version-16) and everything went fine.

## How Has This Been Tested?

N/A

## Checklist:

- [x] I have read and followed the contribution [guidelines](https://github.com/seerr-team/seerr/blob/develop/CONTRIBUTING.md).
- [x] Disclosed any use of AI (see our [policy](https://github.com/seerr-team/seerr/blob/develop/CONTRIBUTING.md#ai-assistance-notice))
- [ ] I have updated the documentation accordingly.
- [x] All new and existing tests passed.
- [x] Successful build `pnpm build`
- [ ] Translation keys `pnpm i18n:extract`
- [ ] Database migration (if required)

## Summary by CodeRabbit

* **Chores**
  * Upgraded and pinned Next/React and related packages; updated TS config, CI workflow triggers, and formatter ignore list; migrated SVG handling to the new bundler rules.

* **New Features**
  * Replaced notification system with a new global toaster and introduced a project toast hook for unified notifications.

* **Refactor**
  * Standardized TypeScript/JSX imports and explicit return typings across UI components; converted Next config to a typed export.

* **Bug Fixes**
  * Desktop sidebar logo now loads eagerly to improve perceived load performance.

PR #3147, opened 2026-06-10
fix: restore Webpack SVG loader config next to Turbopack

DevGitPit:
When building Next.js with Webpack (such as on architectures where Turbopack is unsupported, or using production builds), SVGs are imported as plain asset objects rather than React components. This causes a crash (React Error #130) when rendering SVGs. This PR restores the Webpack rules alongside the Turbopack configuration.

## Summary by CodeRabbit

* **New Features**
  * SVG files can now be imported and used as components within the application, improving asset handling and flexibility in UI development.

### DIFF
--- a/next.config.ts
+++ b/next.config.ts
@@ -21,6 +21,15 @@ const nextConfig: NextConfig = {
       },
     },
   },
+  webpack(config) {
+    config.module.rules.push({
+      test: /\.svg$/,
+      issuer: /\.(js|ts)x?$/,
+      use: ['@svgr/webpack'],
+    });
+
+    return config;
+  },
   experimental: {
     scrollRestoration: true,
     largePageDataBytes: 512 * 1000,


fallenbagel:
Closing as not applicable. This restores dead code to our build pipeline.

The `webpack()` hook this PR restores was deliberately removed in #3003 when we upgraded to Next.js 16, where Turbopack is the default bundler for both `next dev` and `next build`.

SVG handling was migrated to `turbopack.rules` at the same time, and that's the **only supported path**. Nothing in our scripts, CI, or Dockerfile passes `--webpack`, so the restored hook never executes. I verified this against a clean build on next 16.2.6 with this exact config: because a `turbopack` key is present, Next treats the project as migrated and silently ignores the `webpack()` function.

The justifications in the description also don't hold up:
- We only publish linux/amd64 and linux/arm64 images, both of which Turbopack fully supports. There is no supported architecture where Turbopack is unavailable.
- Production builds use Turbopack by default in Next 16, so the claim about production builds is backwards.
- "React Error #130" auto-linked to an unrelated 2022 PR. The actual React error I'm guessing you're mentioning is ("Element type is invalid"), can only occur here if you deliberately build with `next build --webpack`, which is not a workflow this project uses or supports.

If you're maintaining a fork that builds with `--webpack`, you're welcome to carry this patch locally, but we won't maintain a second bundler path upstream.

Separately, the PR template was left empty and the AI disclosure checkbox was not checked. Please review our contributing guidelines before submitting again. Undisclosed AI-generated PRs with unverified claims may be closed without review in the future.

**If you have a concrete scenario where Turbopack is unavailable on a supported deployment (linux/amd64 or arm64), open a GitHub Discussion with details, or reach out on our Discord (recommended).**

PR #3148, opened 2026-06-10
fix: availability sync demotion and orphan season rollup edge cases

### LINKED ISSUE #3129: Series incorrectly marked as "Partially Available" even though all seasons are available
### Description

Several series showed up as "Partially Available" after the 3.3.0 update. I think it may be due to logic changes to fix #3044. Because these series were correctly marked "Available" prior to 3.3.0, I don't believe it's a duplicate of #3085.

### Version

3.3.0

### Steps to Reproduce

I'm using the [official docker image](ghcr.io/seerr-team/seerr:latest) on an unRAID server.

Several series are affected, but I'll pick two where neither has placeholder seasons. Shogun, in fact, only has a single season and zero special episodes.

[Victoria (2016)](https://thetvdb.com/series/victoria). There are 3 seasons + specials. I have "Allow Special Episodes Requests" disabled.

[Shogun (1980)](https://thetvdb.com/series/shogun). There is a single season, no special episodes, and no placeholder seasons; it's about as simple as a series gets.

### Screenshots

Image

Image

### Logs

```shell

```

### Platform

desktop

### Database

SQLite (default)

### Device

unRAID NAS

### Operating System

Docker

### Browser

N/A

### Additional Context

_No response_

### Search Existing Issues

### Code of Conduct

fallenbagel:
## Description
This follows up on #3044, which made the scanner's availability rollup ignore season 0 (specials) when deciding if a show is fully available, since specials availability isn't a reliable signal and not every server/agent reports them consistently. However, that change only touched `baseScanner`, so `availabilitySync` was left checking season 0 like any other season.
A show could be scanned as AVAILABLE, then demoted to PARTIALLY_AVAILABLE by the next sync run because the specials season was missing from Sonarr/Plex/Jellyfin, then promoted back to AVAILABLE on the next scan, repeating indefinitely.

This PR makes availability sync consistent with the scanner: season 0 is excluded from the "is this show missing seasons" check, so it can no longer demote the show's overall status. A missing specials season can still be marked DELETED at the season level, it just won't drag the show down anymore.

Separately, this PR also fixes an unrelated bug in the scanner's rollup that #3044 introduced as a side effect. That PR changed the rollup to treat any season not present in the current scan as missing unless it had `totalEpisodes > 0`, with `?? Infinity` as the fallback for DB-only seasons. This meant that  a season that exists in the database but isn't returned by the metadata provider would always block the rollup. If such a row has status UNKNOWN (e.g. a placeholder created while TMDB briefly listed an upcoming season that was later removed), no future scan can ever revisit it, so the show gets pinned at PARTIALLY_AVAILABLE permanently. The rollup now only lets these DB-only seasons block availability if their status indicates they previously held content (AVAILABLE/PARTIALLY_AVAILABLE/DELETED) and UNKNOWN orphans are ignored, so the show can return to AVAILABLE.

**One thing worth flagging for later**: this still doesn't account for `enableSpecialEpisodes`. If that setting is on and someone specifically requests/monitors specials, a missing specials season still won't affect the show's overall status and this is the existing behavior in `baseScanner`, and this PR just makes availability sync consistent with it. Properly handling the setting would mean sharing one "does season 0 count" predicate between the scanner and availability sync, gated on `enableSpecialEpisodes`, rather than each having its own hardcoded skip. Left as a follow-up since it's a separate, more invasive change.

- Fixes #3129

## How Has This Been Tested?
- [X] Covered by the new unit tests added in this PR
- [x] Verified against my production instance with ~160 orphaned seasons affected by the rollup bug
- [x] Verified by the OP of #3129 against the specials-demotion fix (preview build available at `preview-sync-regression-fix`)

Previously:

DB:
```
  id  | season_row_id | seasonNumber | status | status4k
------+---------------+--------------+--------+----------
 1965 |          2065 |            1 |      5 |        1
 1965 |          4182 |            2 |      1 |        1
(2 rows)
```

With this fix:

## Checklist:

## Summary by CodeRabbit

* **Improvements**
  * TV shows with missing special seasons now maintain their "Available" status instead of being demoted to "Partially Available"
  * Refined season availability calculations for more accurate show status determination

* **Tests**
  * Added test coverage for special season handling in availability syncing

### DIFF
--- a/server/lib/availabilitySync.test.ts
+++ b/server/lib/availabilitySync.test.ts
@@ -1167,4 +1167,286 @@ describe('AvailabilitySync', () => {
       );
     });
   });
+
+  describe('specials season handling', () => {
+    const tmdbSeasonsWithSpecials = [
+      {
+        id: 100,
+        air_date: '2024-01-01',
+        episode_count: 3,
+        name: 'Specials',
+        overview: '',
+        season_number: 0,
+      },
+      {
+        id: 101,
+        air_date: '2024-01-01',
+        episode_count: 10,
+        name: 'Season 1',
+        overview: '',
+        season_number: 1,
+      },
+    ];
+
+    it('should not demote an available show when only the specials season is missing (Jellyfin)', async () => {
+      configureJellyfin();
+      configureSonarr([{ syncEnabled: true }]);
+
+      const mediaRepository = getRepository(Media);
+
+      const media = new Media();
+      media.tmdbId = 13862;
+      media.mediaType = MediaType.TV;
+      media.status = MediaStatus.AVAILABLE;
+      media.jellyfinMediaId = 'jellyfin-shogun-id';
+      media.externalServiceId = 300;
+      media.seasons = [
+        new Season({
+          seasonNumber: 0,
+          status: MediaStatus.UNKNOWN,
+          status4k: MediaStatus.UNKNOWN,
+        }),
+        new Season({
+          seasonNumber: 1,
+          status: MediaStatus.AVAILABLE,
+          status4k: MediaStatus.UNKNOWN,
+        }),
+      ];
+
+      await mediaRepository.save(media);
+
+      getTvShowImpl = async () => fakeTmdbShow(13862, tmdbSeasonsWithSpecials);
+
+      getItemDataImpl = async (id: string) => {
+        if (id === 'jellyfin-shogun-id') {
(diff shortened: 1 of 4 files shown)


gauthier-th (approved):
LGTM

fallenbagel:
> LGTM

Success

M0NsTeRRR (approved):

PR #3191, opened 2026-06-22
fix(scanners): decline orphaned requests when media is removed from servarr

fallenbagel:
## Description

This fixes a stuck state where a movie/show shows a Request button but blocks the request as a duplicate.

When an approved request is sent to Radarr/Sonarr and the entry later disappears before completion (manual removal, DB rebuild, list cleanup), the orphan cleanup from #2757 resets the media to UNKNOWN but doesn't touch the request. The request stays APPROVED, so the duplicate-request guard keeps blocking re-requests while the media page renders as un-requested and the user can neither see it as requested nor request it again.

Cleanup now declines the approved requests bound to the orphaned media. DECLINED is the correct terminal state here (the request's fulfilment was abandoned) and it's the one status that clears the duplicate guard, so the user can simply request again. The decline runs through the normal request path, so the standard **`MEDIA_DECLINED` notification fires**, which is still appropriate here, since removing an in-flight item from the servarr is an implicit cancellation.

This PR also adds an empty-scan guard. If a server returns an empty list, orphan cleanup for that profile type is skipped. A servarr mid-rebuild serving `200 []` would otherwise look like "every movie was deleted" and mass-orphan (and now mass-decline) the whole library. Genuinely-deleted items are still cleaned up on the next scan once the servarr is populated again.

Orphan cleanup no longer treats an empty servarr response as "everything deleted." Existing tests that encoded deletion as an empty list were updated to use a populated scan missing the target instead.

## How Has This Been Tested?

(Only via unit tests)

## Checklist:

- [X] Disclosed any use of AI (see our [policy](https://github.com/seerr-team/seerr/blob/develop/CONTRIBUTING.md#ai-assistance-notice)) (for unit tests)

## Summary by CodeRabbit

* **New Features**
  * Automatically decline approved media requests for orphaned movies and TV shows when they are no longer found on the scanner service, including both standard and 4K variants.

* **Tests**
  * Added comprehensive test coverage for orphaned request handling across scanner types.

### DIFF
--- a/server/lib/scanners/baseScanner.ts
+++ b/server/lib/scanners/baseScanner.ts
@@ -1,7 +1,12 @@
 import TheMovieDb from '@server/api/themoviedb';
-import { MediaStatus, MediaType } from '@server/constants/media';
+import {
+  MediaRequestStatus,
+  MediaStatus,
+  MediaType,
+} from '@server/constants/media';
 import { getRepository } from '@server/datasource';
 import Media from '@server/entity/Media';
+import MediaRequest from '@server/entity/MediaRequest';
 import Season from '@server/entity/Season';
 import { getSettings } from '@server/lib/settings';
 import logger from '@server/logger';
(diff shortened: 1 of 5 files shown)


server/lib/scanners/baseScanner.ts
```diff
@@ -637,6 +642,43 @@ class BaseScanner<T> {
...
+      // Ensure that the media relation is set so the AfterUpdate
+      // notification hook can resolve it
+      request.media = media;
+      await requestRepository.save(request);
+      this.log(
+        `Declined orphaned ${
+          media.mediaType === MediaType.MOVIE ? 'movie' : 'series'
+        } request ${request.id} for ${media.tmdbId} not found in any Sonarr/Radarr server.`,
+        'info'
+      );
```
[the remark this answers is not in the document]
fallenbagel: Tis fine for now

gauthier-th (approved):
LGTM

server/lib/scanners/radarr/index.ts
```diff
@@ -106,6 +122,13 @@ class RadarrScanner
         this.didScan4k = false;
       }
 
+      if (this.serverReturnedEmpty) {
+        this.didScanStandard = false;
```
M0NsTeRRR, line 126: Nitpick: I would have used `didScan` since we never use the term "standard" for non-4K movies. But it’s not an issue so far.

M0NsTeRRR (approved):

server/lib/scanners/radarr/index.ts
fallenbagel: Thats an already existing variable that was added a while back that isn't really touched by this PR. Would be out of scope to rename it in this PR. A future PR could work to clean it up

PR #3206, opened 2026-06-28
fix(api): prevent 500 when TMDB omits credits/keywords

Clancey:
## Description

TMDB intermittently omits appended sub-objects (`credits`, `keywords`, `aggregate_credits`) from movie/TV detail responses, even though they are requested via `append_to_response`. When that happens, `mapMovieDetails` / `mapTvDetails` access `movie.credits.cast` on an `undefined` value and throw:

```
Cannot read properties of undefined (reading 'cast')
```

The error is caught by the route handler and the **entire** detail page returns a 500 ("Unable to retrieve movie/series"), so the title becomes completely unviewable.

Observed in the wild, e.g.:

```
[debug][API]: Something went wrong retrieving movie {"errorMessage":"Cannot read properties of undefined (reading 'cast')","movieId":"1119449"}
```

## Change

Guard the credits/keywords accesses in the model mappers with optional chaining and fall back to empty arrays:

- `server/models/Movie.ts` — `credits.cast`, `credits.crew`, `keywords.keywords`
- `server/models/Tv.ts` — `aggregate_credits.cast`, `credits.crew`, `keywords.results`

The title now loads with empty cast/crew/keywords instead of failing outright. This mirrors how other optional appended data (e.g. `watch/providers`) is already handled with `?.` + a fallback.

## Notes

- Scope intentionally kept to the runtime mappers. The TMDB interface types still declare these fields as required; tightening them to optional would cascade required guards into many scanner/route files and is out of scope for this fix.
- No behavior change for the common case where TMDB returns the appended data.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of missing cast, crew, and keyword data so movie and TV details load more reliably.
  * Prevented errors when some optional credits or keyword fields are unavailable by using safer fallbacks.
  * Kept the displayed details structure unchanged while making data mapping more resilient.

### DIFF
--- a/server/models/Movie.ts
+++ b/server/models/Movie.ts
@@ -132,8 +132,8 @@ export const mapMovieDetails = (
   runtime: movie.runtime,
   tagline: movie.tagline,
   credits: {
-    cast: movie.credits.cast.map(mapCast),
-    crew: movie.credits.crew.map(mapCrew),
+    cast: (movie.credits?.cast ?? []).map(mapCast),
+    crew: (movie.credits?.crew ?? []).map(mapCrew),
   },
   collection: movie.belongs_to_collection
     ? {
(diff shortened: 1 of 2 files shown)


fallenbagel:
Closing this. The PR description is AI-generated output rather than your own words, and the Copilot co-author on the commit indicates this was AI-driven rather than AI-assisted. Both are against our [contributing guidelines](https://github.com/seerr-team/seerr/blob/develop/CONTRIBUTING.md). Please read them before submitting.

PR #3207, opened 2026-06-28
fix(deps): override handlebars to 4.7.9 (CVE-2026-33937)

clabeuhtegrite:
## Description

`handlebars` is pulled in transitively (`email-templates@13.0.1` → `@ladjs/consolidate@1.0.4` → `handlebars@4.7.8`). Versions `4.0.0`–`4.7.8` are affected by **CVE-2026-33937** — a JavaScript injection / RCE in Handlebars.js: `Handlebars.compile()` also accepts a pre-parsed AST, and the `value` field of a `NumberLiteral` node is emitted into the generated function body without sanitization, so code that passes an attacker-controlled value to `compile()` can be made to execute arbitrary code. It is fixed in **4.7.9**.

This adds a `handlebars` entry to `pnpm.overrides` forcing the patched `4.7.9`. `4.7.9` is a patch release and is semver-compatible with the `^4.7.6` peer requirement of `@ladjs/consolidate`, so there is no API or behaviour change.

Scope note: seerr only compiles its own (string) email templates through `email-templates`, so it does not pass attacker-controlled input to `Handlebars.compile()` — real exploitability here is low. This is defense-in-depth, and it clears the CRITICAL finding that image scanners (Trivy / Grype) currently report on the published image.

Disclosure: this change was prepared with AI assistance and reviewed by me before submission.

Ref: https://nvd.nist.gov/vuln/detail/CVE-2026-33937

## How Has This Been Tested?

- Added the `handlebars: "4.7.9"` override and regenerated `pnpm-lock.yaml` with the project's pinned `pnpm@10.24.0`.
- Verified the lockfile diff is limited to `handlebars` (`4.7.8` → `4.7.9`) and the peer-dependency suffixes that reference it — no other dependency is touched.
- Confirmed `4.7.9` satisfies the `^4.7.6` peer requirement of `@ladjs/consolidate`.
- I did not run a full `pnpm build` locally (my Node is 20, below the project's `^22.19.0` engine). The change is a transitive **runtime** dependency override with no source changes, so it has no build-time impact.

N/A — dependency-only change.

## Checklist:

- [x] Disclosed any use of AI (see our [policy](https://github.com/seerr-team/seerr/blob/develop/CONTRIBUTING.md#ai-assistance-notice)) — prepared with AI assistance, reviewed by me.
- [ ] I have updated the documentation accordingly. _(no documentation change required for a dependency override)_
- [ ] All new and existing tests passed. _(not run locally — Node engine mismatch; no source change)_
- [ ] Successful build `pnpm build` _(not run locally — Node engine mismatch; no build-time impact)_
- [ ] Translation keys `pnpm i18n:extract` _(N/A — no strings changed)_
- [ ] Database migration (if required) _(N/A)_

### DIFF
--- a/package.json
+++ b/package.json
@@ -213,7 +213,8 @@
       "sqlite3>node-gyp": "8.4.1",
       "@types/express-session": "1.18.2",
       "@types/react": "19.2.14",
-      "@types/react-dom": "19.2.3"
+      "@types/react-dom": "19.2.3",
+      "handlebars": "4.7.9"
     }
   },
   "scarfSettings": {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -9,6 +9,7 @@ overrides:
   '@types/express-session': 1.18.2
   '@types/react': 19.2.14
   '@types/react-dom': 19.2.3
+  handlebars: 4.7.9
 
 importers:
 
@@ -82,7 +83,7 @@ importers:
         version: 0.2.9
       email-templates:
         specifier: 13.0.1
-        version: 13.0.1(@babel/core@7.29.0)(handlebars@4.7.8)(mustache@4.2.0)(pug@3.0.4)(react-dom@19.2.6(react@19.2.6))(react@19.2.6)(underscore@1.13.7)
+        version: 13.0.1(@babel/core@7.29.0)(mustache@4.2.0)(pug@3.0.4)(react-dom@19.2.6(react@19.2.6))(react@19.2.6)(underscore@1.13.7)
       express:
         specifier: 5.2.1
         version: 5.2.1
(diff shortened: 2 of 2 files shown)


fallenbagel:
Thanks for the write-up, but this isn't exploitable in seerr and we don't need the override.

Handlebars is a peer dep of `@ladjs/consolidate` because consolidate supports every template engine, but `consolidate.handlebars()` is never called by and seerr explicitly uses pug templates exclusively. 

The `require('handlebars')` inside consolidate is lazy (it's inside the render function body, not at module load), so it's never executed at runtime. CVE-2026-33937 requires passing attacker-controlled input to `Handlebars.compile()`, which can't happen here.

The override just essentially adds noise by adding a lockfile-touching override to fix a code path that's never reachable.

Plus, this was already assessed during our last dep upgrade (by me personally) and left as-is for the same reason.

CC: @seerr-team/seerr-core wdyt?

clabeuhtegrite:
Thanks for the detailed explanation, fair call. You're right that the path is unreachable here (pug-only templates + the lazy `require('handlebars')` inside consolidate means `Handlebars.compile()` is never invoked), which is a stronger reason than the low-exploitability note I'd written. Closing this — quieting the scanner CRITICAL isn't worth lockfile noise for a dead code path. Appreciate the thorough review.

PR #3218, opened 2026-07-02
Fix "hide blocklist items" setting description

Asinin3:
## Description

Setting description currently incorrectly states it hides blocklisted items for users with the manage blocklist items permissions when it does the opposite. Also should specifically say the "view blocklisted media" permission, instead of "manage blocklist" as users can be assigned the view permission, without the manage permission.

## How Has This Been Tested?

Haven't tested, but its only a description change

## Checklist:

No AI

### DIFF
--- a/src/components/Settings/SettingsMain/index.tsx
+++ b/src/components/Settings/SettingsMain/index.tsx
@@ -54,7 +54,7 @@ const messages = defineMessages('components.Settings.SettingsMain', {
   streamingRegionTip: 'Show streaming sites by regional availability',
   hideBlocklisted: 'Hide Blocklisted Items',
   hideBlocklistedTip:
-    'Hide blocklisted items from discover pages for all users with the "Manage Blocklist" permission',
+    'Hide blocklisted items from discover pages for all users without the "View Blocklisted Media" permission',
   toastApiKeySuccess: 'New API key generated successfully!',
   toastApiKeyFailure: 'Something went wrong while generating a new API key.',
   toastSettingsSuccess: 'Settings saved successfully!',


fallenbagel:
This is wrong.

Closing. The current description is correct, and this change would make it wrong.

How blocklist visibility actually works:

- Users with neither permission (`MANAGE_BLOCKLIST` and `VIEW_BLOCKLIST`) never see blocklisted items anywhere. They are filtered out unconditionally in `MediaSlider` and `ListView`, regardless of this setting.
- Users with View Blocklist can always see blocklisted items. This setting does not target them. Even with this setting on, they will still see them.
- Users with Manage Blocklist see blocklisted items by default so they can manage them. This setting exists specifically to hide blocklisted items from discover pages for those users. The filter in `useDiscover.ts` checks the global setting together with `hasPermission(Permission.MANAGE_BLOCKLIST)`.

So the description saying it hides items for users with the "Manage Blocklist" permission is exactly what the code does, not the opposite. With the setting on, a user with only View Blocklist still sees blocklisted items, and they disappear the moment that user is granted Manage Blocklist.

`VIEW_BLOCKLIST` plays no part in this setting. It gates access to the blocklist page and being able to see blocklisted items at all. Rewording the description to reference the view permission would describe behavior that does not exist in the code.

With the setting off, for a user with the Manage Blocklist permission:


With the setting on, for a user with the Manage Blocklist permission:


**Another user with neither the Manage Blocklist nor the View Blocklisted Media permission, with the Settings > General > Hide Blocklisted Items setting OFF. They still cannot see the item:**


Now give that same user the View Blocklisted Media permission or don't give that permission but only Manage Blocklist perms:


Now that user sees it, even with the Hide Blocklisted Items setting turned on:


See the docs for what this setting does:
https://docs.seerr.dev/using-seerr/settings/general/#hide-blocklisted-items

To be clear on why the description names Manage Blocklist: it is the only permission whose visibility this setting changes. Users with Manage Blocklist see blocklisted items by default so they can manage them, and this setting is the opt-in to hide those items from their discover pages. That check is on Manage Blocklist alone, so it applies even if the user also has View Blocklisted Media. Users with only View Blocklisted Media can always see blocklisted items, that is the entire point of that permission, so this setting does not touch them. And users with neither permission never see blocklisted items regardless of this setting, which is likely what was mistaken for this setting "doing the opposite". The proposed wording "for all users without the View Blocklisted Media permission" is therefore wrong on both ends: normal users are hidden by the permission system rather than this setting, and users holding both permissions are still hidden by this setting despite having the view permission.

PR #3219, opened 2026-07-02
feat(filters): save active filters for movies and tv

### LINKED ISSUE #369: [Feature Request] Save active filter in Movies and Series
### Description

It would be nice to have the possibility to save your Active Filter in Moves and series.
When you change the filter the way you like it, it works, But it's gone when you came back. Then you have to redo the filter again.

Running the latest version 1.5.0

### Desired Behavior

I like to have a button that says SAVE filter, so next time to go to Movies or Series your filter is like it was when you save it.

### Additional Context

_No response_

### Code of Conduct

modarah:
## Description
This is the implementation of feature request to save active filters in movies and series.

Closes #369 

The PR adds:
- a hook to manage saving and retrieving filters in local storage which uses `QueryFilterOptions` for parsing.
- Save active filter button in`FilterSlideover` component.
- new translation key for the new button with DE translation

The hook gets initiliased in `FilterSlideover` and uses type movie and tv to create a storage key for each when clicking on the save button and use it for retrieving data on render.
Clear active filters button also removes the values from local storage. I added a delete button because users might want to clear the active filters and set other filters without losing the one that they explicitly saved but then went agains it (three buttons is too much).
The saved filters in local storage will only be updated if the user clicks on the save button, so adding other filters without clicking save won't affect the saved ones.

I went with localStorage to keep it simple. If syncing between devices is required then we need a new column in some table somewhere.

### AI:
Claude code for the first draft.

## How Has This Been Tested?
- Clicked filters, saved them, removed them and navigated between pages
- No errors when running `typecheck`
- `pnpm build`

grafik

## Checklist:

## Summary

* **New Features**
  * Added a full-width **“Save Active Filters”** action to the Discover filter panel.
  * Saved filters now **automatically reapply** when you reopen the filter panel.

* **Updates / Improvements**
  * **Clear Active Filters** now removes saved filter settings and closes the panel (no direct query-state clearing).
  * Added localized notifications when saving or removing filters fails.

* **Documentation**
  * Updated Discover filter panel translations for the new button and related error messages (including German and English).


src/components/Discover/FilterSlideover/index.tsx
```diff
@@ -64,6 +67,20 @@ const FilterSlideover = ({
...
+    updateLocalStorage,
+  } = useSavedFilters(type);
+
+  const savedFilters = getSavedFilters();
+
+  useEffect(() => {
+    if (savedFilters && updateLocalStorage.current) {
+      batchUpdateQueryParams(savedFilters);
+    }
+  }, [savedFilters, batchUpdateQueryParams, updateLocalStorage]);
```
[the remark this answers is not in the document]
modarah: Daaamn rabbit, I forgot to commit this one :D

src/i18n/locale/de.json
0xSysR3ll, line 1: Translations are handled by Weblate. Only `en.json` should be committed ;)

0xSysR3ll (changes requested):

src/i18n/locale/de.json
modarah: Adapted, thank you!

modarah:
Hey @0xSysR3ll :)

This PR (and my other one) always showing that there are changes requested although you resolved yur comment. Is there something else or just awaiting re-review?

Thank you!

0xSysR3ll:
> Hey @0xSysR3ll :)
> 
> This PR (and my other one) always showing that there are changes requested although you resolved yur comment. Is there something else or just awaiting re-review?
> 
> Thank you!

Yep. It will show until I properly review and approve the PR ;)

src/components/Discover/FilterSlideover/index.tsx
```diff
@@ -64,6 +67,30 @@ const FilterSlideover = ({
...
   const updateQueryParams = useUpdateQueryParams({});
   const batchUpdateQueryParams = useBatchUpdateQueryParams({});
+  const {
+    saveFilters,
+    getSavedFilters,
+    removeSavedFilters,
+    updateLocalStorage,
+  } = useSavedFilters(type);
+
+  const savedFilters = getSavedFilters();
```
fallenbagel: This runs on every render of `FilterSlideover`, whether it's needed or not. `getSavedFilters()` does a `localStorage.getItem` + `JSON.parse` + zod `.parse()` each time, including re-renders that have nothing to do with filters. Imo, this should be wrapped in `useMemo(() => getSavedFilters(), [])` (it only needs to be read once on mount) or moving the read into the effect itself so it's not recomputed on unrelated renders.

src/components/Discover/FilterSlideover/index.tsx
```diff
@@ -64,6 +67,30 @@ const FilterSlideover = ({
...
+    ) {
+      batchUpdateQueryParams(savedFilters);
+      updateLocalStorage.current = false;
+    }
+  }, [
+    savedFilters,
+    batchUpdateQueryParams,
+    updateLocalStorage,
+    currentFilters,
+  ]);
```
fallenbagel: `savedFilters` and `currentFilters` are new object references every render (`getSavedFilters()` isn't memoized, and `currentFilters` comes in as a fresh object from `prepareFilterValues` in the parent), so this effect body re-executes on essentially every render rather than only when something meaningful changes. It's currently only safe because of the `updateLocalStorage.current` guard. It works, but the deps array isn't doing its job. Once `getSavedFilters` gets memoized per the comment above, you should derive a stable key (e.g. `JSON.stringify(currentFilters)` or maybe just `Object.keys(currentFilters).length`) for the dep array instead of the object itself.

fallenbagel (changes requested):

src/components/Discover/FilterSlideover/index.tsx
modarah: Thank you for the feedback, it was very helpfull!

I believe I don't need any dependency for the effect or the `updateLocalStorage` ref. Like you said about `savedFilters`, the whole thing only needs to run once and that's it.

PR #3224, opened 2026-07-04
fix(availability-sync): prevent false removals for merged-version Plex items

fallenbagel:
## Description

The availability sync incorrectly handles merged Plex items where 1080p and 4K versions share a single library item. For movies this can mark 4K copies as deleted immediately after a Plex scan confirms they exist, because both versions receive the same rating key and the sync assumes the shared key means the 4K copy disappeared instead of inspecting the actual Plex versions. It hits movies whose 1080p and 4K files are versions of one Plex item, which is the normal layout for a single movie library with two Radarr instances. The scanner gives both versions the same rating key, and since #2298 the sync treats a shared rating key as proof that the 4K copy is gone, without ever looking at the actual versions. The 4K status then only survives while the Radarr lookup succeeds, so removing the movie from Radarr after import, or running the 4K instance with Enable Scan off, deletes a movie that is sitting in Plex. The removal also wipes the Radarr link on the media row, so a Plex scan restores the status and the next sync run deletes it again, forever.

This PR lets the actual Plex versions decide instead of the rating key comparison for both movies and TV. Availability is now determined from the media versions Plex reports rather than assuming a shared rating key means one copy disappeared. This also fixes several related issues: merged libraries no longer keep stale non-4K statuses after the 1080p version is removed, metadata Plex has not analyzed yet no longer counts as proof that media is missing, TV season availability is resolved from episode versions in merged libraries, and movies now preserve their service metadata while an approved request is still in flight, matching the existing TV behavior.

Existence lookups now also consult every configured Radarr/Sonarr server regardless of Enable Scan. That toggle gates the library import, while these are point lookups by stored id for state Seerr created itself, so a scan-disabled server still counts as evidence that media exists. Without this the sync and the scanners disagree about which servers exist, and the scanners are the ones writing the 4K statuses.

- Regressed by #2298
- Fixes https://discord.com/channels/783137440809746482/1522678935270195300

Depends on #3223, the new request test here trips that unhandled rejection without it.

## How Has This Been Tested?

- Tested via unit tests to ensure it did not regress the issue fixed by #2298

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved 4K movie availability handling so Plex and Radarr data is interpreted more accurately.
  * Prevented non-4K availability from being inferred when only 4K media is present.
  * Kept 4K availability marked correctly for scan-disabled servers.
  * Preserved 4K-related service details when a matching 4K request is still active, and cleared them when no request remains.

### DIFF
--- a/server/lib/availabilitySync.test.ts
+++ b/server/lib/availabilitySync.test.ts
@@ -1717,6 +1717,316 @@ describe('AvailabilitySync', () => {
     });
   });
 
+  describe('show availability - merged Plex versions', () => {
+    function fakeVersionedEpisode(
+      ratingKey: string,
+      widths: number[]
+    ): PlexMetadata {
+      return {
+        ratingKey,
+        guid: `plex://episode/${ratingKey}`,
+        type: 'movie' as const,
+        title: ratingKey,
+        Guid: [],
+        index: 1,
+        leafCount: 0,
+        viewedLeafCount: 0,
+        addedAt: 0,
+        updatedAt: 0,
+        Media: widths.map((width, i) => ({
+          id: i,
+          duration: 2400,
+          bitrate: width >= 2000 ? 20000 : 4000,
+          width,
+          height: width >= 2000 ? 2160 : 1080,
+          aspectRatio: 1.78,
+          audioChannels: 6,
+          audioCodec: 'eac3',
+          videoCodec: width >= 2000 ? 'hevc' : 'h264',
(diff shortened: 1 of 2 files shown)


M0NsTeRRR (dismissed):

M0NsTeRRR (approved):

0xSysR3ll (approved):

fallenbagel:
Don't merge till #3223 is merged

PR #3241, opened 2026-07-12
test: add scanner update rate override for testing

michaelhthomas:
## Description

The scanner tests are currently pretty slow to run, as they wait for 4 seconds before running the scanner due to the update rate. This can be overridden in tests, which considerably speeds them up.

## How Has This Been Tested?

Tests pass.

## Checklist:

## Summary by CodeRabbit

* **Tests**
  * Improved the reliability and consistency of automated scanner testing for Jellyfin, Radarr, and Sonarr.
  * Added consistent timer handling during scanner test runs to better support asynchronous workflows.
  * Maintained coverage for status handling, cleanup workflows, request processing, orphaned items, and multi-server scenarios.
  * Existing scanner behavior and test expectations remain unchanged.

### DIFF
--- a/server/lib/scanners/baseScanner.ts
+++ b/server/lib/scanners/baseScanner.ts
@@ -59,8 +59,8 @@ export interface ProcessableSeason {
 }
 
 class BaseScanner<T> {
-  private bundleSize;
-  private updateRate;
+  protected bundleSize;
+  protected updateRate;
   protected progress = 0;
   protected items: T[] = [];
   protected totalSize?: number = 0;
@@ -791,6 +791,18 @@ class BaseScanner<T> {
 
   get protectedBundleSize(): number {
     return this.bundleSize;
+  }
+
+  /**
+   * Test-only override to reduce scan delay in unit tests.
+   * Sets updateRate and/or bundleSize for the next run.
+   */
+  public setTestOverrides(opts: {
+    updateRate?: number;
+    bundleSize?: number;
+  }): void {
+    if (opts.updateRate !== undefined) this.updateRate = opts.updateRate;
+    if (opts.bundleSize !== undefined) this.bundleSize = opts.bundleSize;
   }
 }
 
--- a/server/lib/scanners/baseScanner.ts
+++ b/server/lib/scanners/baseScanner.ts
@@ -59,8 +59,8 @@ export interface ProcessableSeason {
 }
 
 class BaseScanner<T> {
-  protected bundleSize;
-  protected updateRate;
+  private bundleSize;
+  private updateRate;
   protected progress = 0;
   protected items: T[] = [];
   protected totalSize?: number = 0;
@@ -798,18 +798,6 @@ class BaseScanner<T> {
 
   get protectedBundleSize(): number {
     return this.bundleSize;
-  }
-
-  /**
-   * Test-only override to reduce scan delay in unit tests.
-   * Sets updateRate and/or bundleSize for the next run.
-   */
-  public setTestOverrides(opts: {
-    updateRate?: number;
-    bundleSize?: number;
-  }): void {
-    if (opts.updateRate !== undefined) this.updateRate = opts.updateRate;
-    if (opts.bundleSize !== undefined) this.bundleSize = opts.bundleSize;
   }
 }
 
--- a/server/lib/scanners/baseScanner.ts
+++ b/server/lib/scanners/baseScanner.ts
@@ -59,8 +59,8 @@ export interface ProcessableSeason {
 }
 
 class BaseScanner<T> {
-  private bundleSize;
-  private updateRate;
+  protected bundleSize;
+  protected updateRate;
   protected progress = 0;
   protected items: T[] = [];
   protected totalSize?: number = 0;
@@ -798,6 +798,18 @@ class BaseScanner<T> {
 
   get protectedBundleSize(): number {
     return this.bundleSize;
+  }
+
+  /**
+   * Test-only override to reduce scan delay in unit tests.
+   * Sets updateRate and/or bundleSize for the next run.
+   */
+  public setTestOverrides(opts: {
+    updateRate?: number;
+    bundleSize?: number;
+  }): void {
+    if (opts.updateRate !== undefined) this.updateRate = opts.updateRate;
+    if (opts.bundleSize !== undefined) this.bundleSize = opts.bundleSize;
   }
 }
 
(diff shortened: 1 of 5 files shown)


server/lib/scanners/baseScanner.ts
```diff
@@ -59,8 +59,8 @@ export interface ProcessableSeason {
 }
 
 class BaseScanner<T> {
-  private bundleSize;
-  private updateRate;
+  protected bundleSize;
+  protected updateRate;
```
fallenbagel: Imo, no need to widen these. `setTestOverrides` is a method on this class, it already has access to the private fields. Widening also makes `protectedBundleSize`/`protectedUpdateRate` below redundant for subclasses, now there's two ways to read the same state.

server/lib/scanners/baseScanner.ts
```diff
@@ -792,6 +792,18 @@ class BaseScanner<T> {
...
+   * Test-only override to reduce scan delay in unit tests.
+   * Sets updateRate and/or bundleSize for the next run.
+   */
+  public setTestOverrides(opts: {
+    updateRate?: number;
+    bundleSize?: number;
+  }): void {
+    if (opts.updateRate !== undefined) this.updateRate = opts.updateRate;
+    if (opts.bundleSize !== undefined) this.bundleSize = opts.bundleSize;
+  }
```
fallenbagel: This is a public, unguarded mutator on a class with live singleton instances (`radarrScanner`, `sonarrScanner`, `jellyfinFullScanner`). Nothing stops it from being called outside tests, unlike `protectedBundleSize`/`protectedUpdateRate` above which are read-only. I'd rather drop the source change and use `mock.timers` in the three test files instead since mock is already imported there for `mock.method`, so `mock.timers.enable()` + `mock.timers.tick(this.updateRate)` gets the same speedup and with the added bonus of zero production code changes.

CC: @seerr-team/seerr-core

server/lib/scanners/radarr/radarr.test.ts
```diff
@@ -75,6 +75,7 @@ function fakeRadarrMovie(overrides: Partial<RadarrMovie> = {}): RadarrMovie {
 
 describe('Radarr Scanner', () => {
   beforeEach(() => {
+    radarrScanner.setTestOverrides({ updateRate: 0 });
```
fallenbagel: Same concern as stated below. Would rather see `mock.timers.enable({ apis: ['setTimeout'] })` in `beforeEach` and a `mock.timers.tick(4000)` after triggering the scan, gets the same speedup without touching `baseScanner.ts`. One tick covers everything currently mocked in this file and bundleSize is 50 (20 for Jellyfin) and nothing here mocks more than a handful of items. So this only needs repeating if a test starts mocking more than a bundle's worth.

server/lib/scanners/jellyfin/jellyfin.test.ts
```diff
@@ -223,6 +223,8 @@ function configureJellyfinWithLibrary(
 
 describe('Jellyfin Scanner', () => {
   beforeEach(async () => {
+    jellyfinFullScanner.setTestOverrides({ updateRate: 0 });
```
fallenbagel: Same as the Radarr test

server/lib/scanners/sonarr/sonarr.test.ts
```diff
@@ -163,6 +163,7 @@ function configureSonarr(overrides: Partial<SonarrSettings>[] = [{}]): void {
 
 describe('Sonarr Scanner', () => {
   beforeEach(() => {
+    sonarrScanner.setTestOverrides({ updateRate: 0 });
```
fallenbagel: Same as the Radarr test

fallenbagel (changes requested):

server/lib/scanners/baseScanner.ts
michaelhthomas: Lowkey didn't realize the Node test runner supported mock timers. I will just go with that, should be a lot simpler.

server/lib/scanners/sonarr/sonarr.test.ts
```diff
@@ -103,6 +103,8 @@ Object.defineProperty(TheMovieDb.prototype, 'getTvShow', {
   configurable: true,
 });
 
+import { sonarrScanner } from '@server/lib/scanners/sonarr';
```
michaelhthomas: Sonarr was being imported before the `getTvShow` mock which caused it to not work. Moved it here so the mock actually applies.

server/test/runWithMockTimers.ts
```diff
@@ -0,0 +1,29 @@
...
+    const maxTicks = 100000;
+    for (let i = 0; i < maxTicks && !settled; i++) {
+      await new Promise((resolve) => setImmediate(resolve));
+      mock.timers.tick(tickMs);
+    }
+    if (!settled) {
+      throw new Error(
+        `runWithMockTimers: promise did not settle after ${maxTicks} ticks`
+      );
+    }
```
fallenbagel, line 29:
```suggestion
    for (let i = 0; i < maxTicks && !settled; i++) {
      await new Promise((resolve) => setImmediate(resolve));
      mock.timers.tick(tickMs);
    }
    await new Promise((resolve) => setImmediate(resolve));
    if (!settled) {
      throw new Error(
        `runWithMockTimers: promise did not settle after ${maxTicks} ticks`
      );
    }
```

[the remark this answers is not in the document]
fallenbagel: @michaelhthomas can you add one more flush before the check? `settled` lags a microtask behind the tick that resolves it `.finally()` is a microtask. If the resolving tick lands on the last iteration, the loop exits before the next flush and this throws a false negative. 
Like so:
https://github.com/seerr-team/seerr/pull/3241#discussion_r3820239286

server/test/runWithMockTimers.ts
```diff
@@ -0,0 +1,29 @@
+import { mock } from 'node:test';
+
+export async function runWithMockTimers<T>(
+  runPromise: Promise<T>,
+  tickMs = 4000
+): Promise<T> {
```
fallenbagel, line 6: Takes an already-invoked Promise instead of a factory. Its fine as it is for now since `setTimeout` in `loop()` only fires after several awaits, but every call site falls back to real timers silently if that ever changes or accidentally forgets to add. `runPromise: () => Promise<T>`, and call it after `enable()`

fallenbagel (changes requested):

fallenbagel (commented):
Just some very few changes and this can be merged

gauthier-th (approved):
LGTM, not tested

fallenbagel:
> not tested

Already tested by CI 😛

https://github.com/seerr-team/seerr/actions/runs/34162930122/job/103298168440#step:8:86

fallenbagel (approved):

PR #3265, opened 2026-07-19
fix: quota bypass when editing a pending request's seasons

### LINKED ISSUE #633: series request limit can be bypassed by users
### Description

If you have a "Series Request Limit" set for a user, for example

4 Seasons per 7 days 

The user is able to bypass the season limit when you have the "Auto Approve" setting disabled. 

While it's in the "Pending State" awaiting admin approval, the user can go back in and edit the request and select more seasons than they are allowed. 

### Version

1.7.0

### Steps to Reproduce

1. Create a test user 
2. Assign a limit to this user for seasons. For example, allow 4 season requests per 7 days. Note: Assign a limit to the user using the Override Global Limit option, don't assign a limit to the user just using the global limit. (General > Users > Select User > General Tab > Series Request Limit > Override Global Limit)
3. Disable "Auto Approve" for this user, so that it requires admin approval. 
4. On the test user, request a show that has more than 4 seasons, but select all that you're allotted to (4 seasons), Then submit
5. Once submitted, head to the requests tab, edit your request, then select the remaining seasons and click save. 

### Screenshots

image
image
image
image
image

### Logs

_No response_

### Platform

desktop

### Device

Any

### Operating System

Unraid

### Browser

Any

### Additional Context

_No response_

### Code of Conduct

kushiemoon-dev:
## Description

Editing a pending TV request's seasons didn't re-check the requester's quota against the new season list. `quota.tv.remaining` is computed from the currently persisted seasons, so any seasons being removed in the same edit were still counted as "used". A user (or an admin editing on a user's behalf) could add seasons beyond the quota limit as long as the request already existed, bypassing the quota enforcement that applies to new requests.

This adds a quota check on edit: removed seasons are credited back before comparing the new season count against the remaining quota, and an edit that would exceed quota is rejected (403) unless a manager explicitly passes `ignoreQuota` (only usable by accounts with `MANAGE_REQUESTS` permission; non-managers passing it get rejected).

Also fixes the 3 request modals (`MovieRequestModal`, `TvRequestModal`, `CollectionRequestModal`) fetching quota only for the modal owner or a manager: a non-manager editing their own existing request via `requestOverrides` wasn't shown their own quota.

Fixes #633

**AI Disclosure:** This PR was written primarily by Claude Code, reviewed and directed by the account holder.

## How Has This Been Tested?

Added a test suite in `server/routes/request.test.ts` covering: rejecting an edit beyond quota, rejecting a manager edit beyond quota without `ignoreQuota`, rejecting a non-manager attempting to set `ignoreQuota` themselves, allowing a manager to bypass quota via `ignoreQuota`, and allowing a same-count season swap at full quota. Full suite passes (`pnpm test`, 126/126), along with `pnpm lint` and `pnpm build`.

N/A: backend/logic fix, no visual change.

## Checklist:

- [x] Disclosed any use of AI (see our policy).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TV season quota enforcement by correctly accounting for seasons removed during edits.
  * Season swaps and replacements are now allowed when the final season count stays within quota.
  * Updated admin quota-bypass behavior and ensured unauthorized bypass attempts are rejected.
  * Quota/permission failures now consistently return clear access-denied responses.
* **Improvements**
  * Request modals now fetch and display quota information consistently for the currently selected requester (including overrides) across movies, TV, and collections.
* **Tests**
  * Added coverage for TV quota enforcement and ignore-quota bypass behavior, including edge cases with concurrent request counts.

### DIFF
--- a/server/routes/request.test.ts
+++ b/server/routes/request.test.ts
@@ -349,6 +349,28 @@ describe('PUT /request/:requestId (tv, season quota enforcement)', () => {
     );
     assert.strictEqual(saved.ignoreQuota, true);
   });
+
+  it('allows swapping seasons at full quota when the count stays the same', async () => {
+    const requestRepo = getRepository(MediaRequest);
+    const mediaRequest = await seedTvRequestAtQuota(90005);
+
+    const agent = await loginAs('friend@seerr.dev', 'test1234');
+    const res = await agent.put(`/request/${mediaRequest.id}`).send({
+      mediaType: MediaType.TV,
+      seasons: [3, 4],
+    });
+
+    assert.strictEqual(res.status, 200);
+
+    const saved = await requestRepo.findOneOrFail({
+      where: { id: mediaRequest.id },
+      relations: { seasons: true },
+    });
+    assert.deepStrictEqual(
+      saved.seasons.map((s) => s.seasonNumber).sort(),
+      [3, 4]
+    );
+  });
 });
 
 describe('POST /request/:requestId/:status', () => {
--- a/server/routes/request.test.ts
+++ b/server/routes/request.test.ts
@@ -371,6 +371,90 @@ describe('PUT /request/:requestId (tv, season quota enforcement)', () => {
       [3, 4]
     );
   });
+
+  it('rejects a non-bypass edit that swaps seasons on an ignoreQuota-exempt request past quota used elsewhere', async () => {
+    const userRepo = getRepository(User);
+    const mediaRepo = getRepository(Media);
+    const requestRepo = getRepository(MediaRequest);
+
+    const friend = await userRepo.findOneOrFail({
+      where: { email: 'friend@seerr.dev' },
+    });
+    friend.tvQuotaLimit = 2;
+    friend.tvQuotaDays = undefined;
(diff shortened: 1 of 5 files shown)


fallenbagel:
Closing this in favour of #3378 (which is a stack of several bug fixes involving quota), which covers the same ground plus the cases this misses: a pure reassignment skips the check here since it's gated on `newSeasons.length > 0`, and it builds on the existing newSeasons expression, which counts a season already held by another user's request as new. #3378 also serializes the handler so two concurrent edits can't both pass, and covers movie reassignment.

Separately, and not the reason for closing (this is more for future contributions): thanks for disclosing. However, that is not what we meant by AI-assisted vs AI-driver. It doesn't change the outcome though, and it isn't meant to. What we ask for is AI-assisted development, not AI-driven: disclosure tells us how much scrutiny to apply, it doesn't turn a contribution that was primarily generated into one that was written. 
https://github.com/seerr-team/seerr/blob/develop/CONTRIBUTING.md#disclosure-requirements

Thanks for having a go at it either way, #633 had been open a long time.

PR #3313, opened 2026-07-30
chore: exclude .pnpm-store from Docker build context

samtheruby:
## Description

On local builds the .pnpm-store directory is not excluded from the build context. Since it's never referenced by the Dockerfile RUN --mount=type=cache uses its own in-container store path, and it gets shipped to the daemon on every build for no benefit. Adding .pnpm-store to .dockerignore cuts build context size by over 1gb with no change to build output.

- Fixes N/A

## How Has This Been Tested?

Local Docker build was run successfully with image output size reduced by 1.2gb

N/A
## Checklist:

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -8,6 +8,7 @@
 .gitconfig
 .github
 .gitignore
+.pnpm-store
 .husky
 .next
 .prettierignore


M0NsTeRRR:
Hello,

What does the `pnpm store path` command return? By default from what I read, it should not be in the seerr folder unless you explicitly configured it to be there.

https://github.com/pnpm/pnpm/blob/main/pnpm/crates/config/src/defaults.rs#L77

samtheruby:
Generally, you are right and on any CI build or even a lot of local builds .pnpm-store should not be carried over, however since my home folder and project folder live on different disks/volumes, pnpm can't hardlink the files properly, so it falls back to putting the store inside the project folder instead. Whether it's worth fixing a niche case is a different matter, but either way I don't think there is any downside to explicitly excluding .pnpm-store so it's never copied over.

```pnpm store path``` returns ```/mnt/user/<share name>/docker-build/seerr/.pnpm-store/v11```

M0NsTeRRR (approved):

gauthier-th (approved):

PR #3324, opened 2026-07-30
fix: stop masking connection failures across media server sync and login

fallenbagel:
## Description

Previously, stopping the Jellyfin server and hitting Sync Libraries returned "Something went wrong while syncing libraries" with no real error in Setup, because the connection failure got treated the same as "no libraries found," and the fallback check for that case failed too and threw the wrong error entirely. Now sync distinguishes an unreachable server from an empty library list, and the toast says the server is unreachable instead of giving a generic error.

Connection failures in every Jellyfin API method threw either `InvalidAuthToken` or a bare `Unknown` error on any failure, so a downed server looked like a bad token everywhere. Login now catches this too: signing in while Jellyfin is unreachable gives the same "unable to connect" message instead of a generic login failure. Plex sync no longer silently
clears the library list on failure either and it throws, so both the route and the startup migration can leave existing config in place and let the user (or the next boot) retry.

In addition, Setup's Continue gate used to run its own validation request that toasted on failure. Switching that to reuse the settings data already being fetched removes a redundant request on every library toggle, but that path is silent on failure, so this PR adds an inline alert to keep users from being stuck on a disabled Continue button with no explanation.

Moreover, scan status polling ran at a constant 1 second regardless of whether a sync was actually in progress. It now backs off to 10 seconds when idle and only ramps up while scanning. This should cut idle traffic significantly without missing scans triggered by the job scheduler.

Also cleans up now-unused setup-flow state and makes Plex's setup-mode handling consistent with Jellyfin's.

## How Has This Been Tested?
- On a fresh setup tested the /Setup library sync and toggling.

image

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Plex and Jellyfin connection error handling with clearer, dedicated messages.
  * Prevented library settings from being cleared when synchronization fails.
  * Plex library migration failures no longer prevent startup and will retry later.

* **Improvements**
  * Library synchronization now checks more frequently while active and less frequently when idle.
  * Setup library completion is determined automatically from loaded libraries.
  * Setup displays an inline error when libraries cannot be loaded.
  * Updated Jellyfin login and setup messages for unreachable servers.


server/api/jellyfin.ts
```diff
@@ -233,7 +233,11 @@ class JellyfinAPI extends ExternalAPI {
         }
       );
 
-      throw new ApiError(e.response?.status, ApiErrorCode.Unknown);
+      if (!e.response) {
+        throw new ApiError(502, ApiErrorCode.ConnectionError);
+      }
+
+      throw new ApiError(e.response.status, ApiErrorCode.Unknown);
```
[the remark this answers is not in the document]
fallenbagel: Irrelevant. jellyfin.ts has the only InvalidUrl thrower reaching auth.ts's switch, so converting it would make that case and the matching one in JellyfinLogin.tsx dead code, with no change to what the user sees.

server/api/plexapi.ts
```diff
@@ -167,7 +169,11 @@ class PlexAPI extends ExternalAPI {
         message: e.message,
       });
 
-      settings.plex.libraries = [];
+      if (!e.response) {
+        throw new ApiError(502, ApiErrorCode.ConnectionError);
+      }
+
+      throw new ApiError(e.response.status, ApiErrorCode.Unknown);
```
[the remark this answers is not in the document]
fallenbagel: fixed in https://github.com/seerr-team/seerr/pull/3324/commits/a095b9f3302aaab3e97d1207240c6d4c8afdb36e

server/routes/auth.ts
```diff
@@ -543,6 +543,30 @@ authRoutes.post('/jellyfin', async (req, res, next) => {
           message: e.errorCode,
         });
 
+      case ApiErrorCode.ConnectionError:
+        logger.error(
+          `Unable to reach the ${
+            settings.main.mediaServerType === MediaServerType.JELLYFIN
+              ? ServerType.JELLYFIN
```
M0NsTeRRR, line 550: As discussed, stick to the 'Jellyfin' string, and let's create another PR to refactor these references.

M0NsTeRRR (changes requested):

M0NsTeRRR (dismissed):

src/components/Login/JellyfinLogin.tsx
```diff
@@ -98,6 +98,7 @@ const JellyfinLogin = ({ revalidate, serverType }: JellyfinLoginProps) => {
             let errorMessage = messages.loginerror;
             switch (e?.response?.data?.message) {
               case ApiErrorCode.InvalidUrl:
+              case ApiErrorCode.ConnectionError:
                 errorMessage = messages.invalidurlerror;
```
0xSysR3ll, line 102: Might be a bit out-of-scope but since you added this here, shouldn't we also add this to [ JellyfinSetup.tsx](https://github.com/seerr-team/seerr/blob/7baed83723839d64199fd8e4257f00cb375811fe/src/components/Setup/JellyfinSetup.tsx#L135) ?

fallenbagel: Nah not out of scope, setup is arguably where it matters more. A first-time user typing an unreachable hostname getting a "Something went wrong while trying to sign in" instead of "Unable to connect," with no prior working config to compare against is a way worse error masking. Good point out!

0xSysR3ll (approved):
LGTM!

M0NsTeRRR (approved):

PR #3355, opened 2026-08-08
feat: track episode availability from sonarr

0xSysR3ll:
## Description

Part 2 of #1671 to make the review easier.
Implements episode availability in Sonarr's scanner.

## How Has This Been Tested?

Same tests as part 1 + server checks.

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Added optional episode-level availability tracking for Sonarr TV shows.
  - Episode file status is synchronized, including support for 4K availability.
  - Anime-specific metadata provider selection is supported during synchronization.

- **Bug Fixes**
  - Removed episodes no longer available are excluded from season availability results.
  - Synchronization continues when episode lookups or metadata requests fail.

### DIFF
--- a/server/lib/availabilitySync.ts
+++ b/server/lib/availabilitySync.ts
@@ -3,21 +3,28 @@ import JellyfinAPI from '@server/api/jellyfin';
 import type { PlexMetadata } from '@server/api/plexapi';
 import PlexAPI from '@server/api/plexapi';
 import RadarrAPI, { type RadarrMovie } from '@server/api/servarr/radarr';
-import type { SonarrSeason, SonarrSeries } from '@server/api/servarr/sonarr';
+import type {
+  EpisodeResult,
+  SonarrSeason,
+  SonarrSeries,
+} from '@server/api/servarr/sonarr';
 import SonarrAPI from '@server/api/servarr/sonarr';
 import TheMovieDb from '@server/api/themoviedb';
+import { ANIME_KEYWORD_ID } from '@server/api/themoviedb/constants';
 import type { TmdbTvDetails } from '@server/api/themoviedb/interfaces';
 import { MediaRequestStatus, MediaStatus } from '@server/constants/media';
 import { MediaServerType } from '@server/constants/server';
 import { getRepository } from '@server/datasource';
+import Episode from '@server/entity/Episode';
 import Media from '@server/entity/Media';
 import MediaRequest from '@server/entity/MediaRequest';
 import type Season from '@server/entity/Season';
 import { User } from '@server/entity/User';
 import type { RadarrSettings, SonarrSettings } from '@server/lib/settings';
-import { getSettings } from '@server/lib/settings';
+import { MetadataProviderType, getSettings } from '@server/lib/settings';
 import logger from '@server/logger';
 import { getHostname } from '@server/utils/getHostname';
+import { In } from 'typeorm';
 
 class AvailabilitySync {
   public running = false;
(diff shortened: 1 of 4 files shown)


fallenbagel:
Is the media server scanner gonna be stacked on this PR?

0xSysR3ll:
> Is the media server scanner gonna be stacked on this PR?

It's supposed to but I guess I missed a step somewhere.

server/lib/availabilitySync.ts
```diff
@@ -30,6 +37,7 @@ class AvailabilitySync {
   private jellyfinEpisodeExistsCache: Record<string, boolean>;
 
   private sonarrSeasonsCache: Record<string, SonarrSeason[]>;
+  private sonarrEpisodesCache: Record<string, EpisodeResult[]>;
```
fallenbagel: This holds the raw /episode payload. title, overview, air dates, **all of it**. For every series in the library for the whole run. The keys are per-series and never read for a different media, so it only exists to carry data from `mediaExistsInSonarr` across to `seasonExistsInSonarr` for the same show. Returning it, or clearing it per media, avoids sitting on the entire library at once.

The scanner already does the right thing and reduces to what it needs on the way in:

https://github.com/seerr-team/seerr/blob/105baae944eaa5d0d34235ff6750af942fad48ba/server/lib/scanners/sonarr/index.ts#L198-L206

Same two fields here would cut it down a lot. After #3367  I'd rather not put a library-sized cache back.

server/lib/availabilitySync.ts
```diff
@@ -773,6 +793,22 @@ class AvailabilitySync {
...
+            } catch (e) {
+              logger.error(
+                `Failed to fetch episodes for show [TMDB ID ${media.tmdbId}] from Sonarr.`,
+                {
+                  errorMessage: e.message,
+                  label: 'AvailabilitySync',
+                }
+              );
+            }
+          }
```
fallenbagel: One extra Sonarr call per series per sync, on top of the same per series per scan in the scanner. A 1000-series library picks up 2000 extra requests across the two jobs. Worth skipping series with no files at all, `sonarr.statistics.episodeFileCount` is checked a few lines down anyway and there's nothing to mark available in that case.

server/lib/availabilitySync.ts
```diff
@@ -861,6 +972,56 @@ class AvailabilitySync {
...
+
+        if (
+          hasFile !== true &&
+          currentStatus !== MediaStatus.DELETED &&
+          currentStatus !== MediaStatus.UNKNOWN
+        ) {
+          existingEpisode[is4k ? 'status4k' : 'status'] = MediaStatus.DELETED;
+          toSave.push(existingEpisode);
+        }
+      }
```
fallenbagel, line 1030: Rows for numbers Sonarr no longer returns get marked DELETED but stay in the table, and the season route's guard counts every tracked row regardless of status:
https://github.com/seerr-team/seerr/blob/906297602bff14050a6c4cca5e60cb992510cd61/server/routes/tv.ts#L115-L130

So one renumbered episode kills badges for that whole season permanently, with only a debug line to explain it. Delete the rows whose numbers are gone instead of soft-removing them, or maybe filter DELETED out of the guard.

server/lib/scanners/baseScanner.ts
```diff
@@ -644,11 +653,95 @@ class BaseScanner<T> {
...
+            status4k: is4k
+              ? episodeDetail.hasFile
+                ? MediaStatus.AVAILABLE
+                : MediaStatus.UNKNOWN
+              : MediaStatus.UNKNOWN,
+            season: Promise.resolve(dbSeason),
+          });
+          toSave.push(newEpisode);
+          existingBySeasonAndNumber.set(key, newEpisode);
+        }
```
fallenbagel: Rows get created for episodes with no file. Nothing reads UNKNOWN as available and the demotion path skips UNKNOWN, so these are inert, and they widen the surface for the number mismatch guard. Only persist episodes that have a file.

fallenbagel (changes requested):

PR #3361, opened 2026-08-10
feat(episodes): add episode availability tracking and sync

### LINKED ISSUE #270: See which episode of a series is avaliable and which not

0xSysR3ll:
## Description

> [!NOTE]
> Reopened as a new PR so it can stack with #3355.
> Prior discussion and reviews: **#1671** (closed as duplicate of this PR).

This allows seerr to track the availability status of individual episodes, enabling better status reporting for partially available seasons.

- Fixes #270

## How Has This Been Tested?

- Complete season except one episode available.
- Correctly show the "available" badge for each episode except for the missing one.

### Settings
image

### Result
image

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Added an optional episode availability tracking setting.
  - Season pages now show an availability badge for episodes confirmed available through connected media services.
  - Episode availability is matched against metadata providers for more accurate results.

- **Improvements**
  - Added support for storing and displaying episode-level availability across seasons.
  - Added configuration and API documentation for the new setting and availability information.
  - Added Ntfy notification tag configuration support.


server/routes/tv.ts
```diff
@@ -84,7 +87,72 @@ tvRoutes.get('/:id/season/:seasonNumber', async (req, res, next) => {
...
 
-    return res.status(200).json(mapSeasonWithEpisodes(season));
+    let availableMap: Record<number, boolean> | undefined;
+
+    const settings = getSettings();
+    const shouldTrackEpisodes =
+      settings.main.enableEpisodeAvailability &&
+      (isAnime
+        ? settings.metadataSettings.anime === MetadataProviderType.TVDB
+        : settings.metadataSettings.tv === MetadataProviderType.TVDB);
```
fallenbagel: Doesn't this mean that this trusts the setting rather than the provider we actually got back. `getMetadataProvider` swallows a failed `Tvdb.getInstance()` and quietly hands back TheMovieDb:

https://github.com/seerr-team/seerr/blob/818a09dcdf045e622bde1566218efd60f4a8c8cc/server/api/metadata.ts#L31-L37

So with TVDB selected but broken, `season.episodes` comes back in TMDB numbering while we still think we're tracking. `metadataProvider` is right there and `!(metadataProvider instanceof TheMovieDb)` would've been better?

server/routes/tv.ts
```diff
@@ -84,7 +87,72 @@ tvRoutes.get('/:id/season/:seasonNumber', async (req, res, next) => {
...
+
+    if (shouldTrackEpisodes) {
+      availableMap = {};
```
fallenbagel: When there's no matching row this stays `{}` and every episode serializes as `available: false` rather than being omitted. The field is optional in the schema, so it'd be better to only assign the map once we have something to say, otherwise clients can't tell "not available" from "we don't track this", right?

server/routes/tv.ts
```diff
@@ -84,7 +87,72 @@ tvRoutes.get('/:id/season/:seasonNumber', async (req, res, next) => {
...
+        relations: {
+          episodes: true,
+        },
+      });
+
+      if (dbSeason) {
+        if (dbSeason.status === MediaStatus.AVAILABLE) {
+          for (const episode of season.episodes) {
+            availableMap[episode.episode_number] = true;
+          }
```
fallenbagel: This ignores the rows we just loaded and skips the guard the other branch applies. It asserts every episode is available off the season flag alone.

They can disagree without any provider weirdness involved. `availabilitySync` in #3355 demotes an individual episode to DELETED when Sonarr says the file is gone, and doesn't touch the season status. So: file deleted, row says DELETED, season still AVAILABLE, and we report it available anyway.

Reading the rows whenever there are any, and only falling back to the season status when there are none, makes the two branches consistent.

server/entity/Episode.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  @Column({ type: 'int', default: MediaStatus.UNKNOWN })
+  public status4k: MediaStatus;
+
+  @Index()
+  @ManyToOne(() => Season, (season: Season) => season.episodes, {
+    onDelete: 'CASCADE',
+    nullable: true,
+  })
+  public season?: Promise<Season>;
```
fallenbagel: Why nullable? `Season.media` isn't, and an episode with no season is unreachable. Also a unique on (season, episodeNumber) would be worth adding too, imo. The mismatch check in the season route compares a Set size against `episodes.length`, which quietly assumes it.

server/routes/tv.ts
```diff
@@ -84,7 +87,72 @@ tvRoutes.get('/:id/season/:seasonNumber', async (req, res, next) => {
...
+              'Skipping episode availability due to episode number mismatch',
+              {
+                label: 'API',
+                tvId: req.params.id,
+                seasonNumber: req.params.seasonNumber,
+                metadataEpisodeCount: metadataEpisodeNumbers.size,
+                trackedEpisodeCount: dbSeason.episodes.length,
+              }
+            );
+            availableMap = undefined;
```
fallenbagel: This logs at warn but fires in perfectly normal situations. An empty `dbSeason.episodes` against N metadata episodes counts as a mismatch, so that's every partially available season until the scanner has run, plus any show where TVDB hands back an episode-less season. It's once per request too, so it repeats every time someone opens a season tab.

Better at debug, and skip the log entirely when there are no rows since that's "nothing tracked yet", not a mismatch.

server/entity/Episode.ts
```diff
@@ -0,0 +1,50 @@
...
+  public id: number;
+
+  @Column()
+  public episodeNumber: number;
+
+  @Column({ type: 'int', default: MediaStatus.UNKNOWN })
+  public status: MediaStatus;
+
+  @Column({ type: 'int', default: MediaStatus.UNKNOWN })
+  public status4k: MediaStatus;
```
fallenbagel, line 27: I dont see anything reading `status4k` and the season route only looks at status
- https://github.com/seerr-team/seerr/pull/3361/changes#diff-a9ea41230489d4e371d6c319f659f700719a92779a2c2c49b32ff419e1d0f0f4R148
- https://github.com/seerr-team/seerr/pull/3361/changes#diff-a9ea41230489d4e371d6c319f659f700719a92779a2c2c49b32ff419e1d0f0f4R121
- https://github.com/seerr-team/seerr/pull/3361/changes#diff-a9ea41230489d4e371d6c319f659f700719a92779a2c2c49b32ff419e1d0f0f4R116
so a 4k-only Sonarr setup gets no badges even though #3355 writes this column.

server/interfaces/api/settingsInterfaces.ts
```diff
@@ -49,6 +50,10 @@ export interface PublicSettingsResponse {
   newPlexLogin: boolean;
   youtubeUrl: string;
   versionCheck: boolean;
+  metadataSettings: {
+    tv: string;
+    anime: string;
+  };
```
fallenbagel: This puts `metadataSettings` on `/settings/public`, which is served without auth:

https://github.com/seerr-team/seerr/blob/818a09dcdf045e622bde1566218efd60f4a8c8cc/server/routes/index.ts#L113

for the sake of a warning only admins ever see. `/api/v1/settings/metadatas` already exists and `SettingsMetadata` pulls from it:

https://github.com/seerr-team/seerr/blob/818a09dcdf045e622bde1566218efd60f4a8c8cc/src/components/Settings/SettingsMetadata.tsx#L72-L73

`{ tv: string; anime: string }` re-declares a shape that already exists in `MetadataSettings` in `server/lib/settings` is exported and is what `FullPublicSettings` uses for this same field:

https://github.com/seerr-team/seerr/blob/818a09dcdf045e622bde1566218efd60f4a8c8cc/server/lib/settings/index.ts#L116-L119

`import type` from there keeps the two descriptions of the same response in sync, and `string` here loses the 'tmdb' | 'tvdb' narrowing at the call site.

src/components/TvDetails/Season/index.tsx
```diff
@@ -52,6 +59,14 @@ const Season = ({ seasonNumber, tvId }: SeasonProps) => {
...
                       <AirDateBadge airDate={episode.airDate} />
                     )}
+                    {showEpisodeAvailability &&
+                      episode.airDate &&
+                      new Date(episode.airDate) <= new Date() &&
+                      episode.available === true && (
+                        <Badge badgeType="success">
+                          {intl.formatMessage(globalMessages.available)}
+                        </Badge>
+                      )}
```
fallenbagel: The `new Date(episode.airDate) <= new Date()` check stops an unaired episode from showing a badge when the season-AVAILABLE bug I flagged on tv.ts fires. It's also a second "has this aired" calculation next to `AirDateBadge` a few lines up, which has its own.

If tv.ts starts preferring the DB rows, `episode.available` becomes trustworthy on its own (unaired episodes come back UNKNOWN), and this can drop to just `episode.available === true`. Better to have one place answering "has it aired," not two.

src/components/TvDetails/Season/index.tsx
```diff
@@ -30,6 +34,9 @@ const Season = ({ seasonNumber, tvId }: SeasonProps) => {
     return <div>{intl.formatMessage(messages.somethingwentwrong)}</div>;
   }
 
+  const showEpisodeAvailability =
+    settings.currentSettings.enableEpisodeAvailability;
```
fallenbagel: This can go. When tracking is off the server never populates the map, so `episode.available` is already undefined for everyone and the `=== true` check below already fails on its own.

src/components/Settings/SettingsMain/index.tsx
```diff
@@ -588,6 +599,45 @@ const SettingsMain = () => {
...
+                      onChange={() => {
+                        setFieldValue(
+                          'enableEpisodeAvailability',
+                          !values.enableEpisodeAvailability
+                        );
+                      }}
+                    />
+                  </div>
+                  {values.enableEpisodeAvailability &&
+                    currentSettings.metadataSettings?.tv !== 'tvdb' &&
```
fallenbagel: instead of comparing against string literals why dont we add a constant enum like mediaServerType, ServerType?

fallenbagel (changes requested):
Main TLDRs;
- the read path only knows about `status`, so 4k is effectively dead weight right now
- the AVAILABLE short-circuit overrides the stored rows and can report a deleted episode as available
- `shouldTrackEpisodes` trusts the setting rather than the provider we actually got back
- `metadataSettings` ends up on the unauthenticated public settings payload just to render an admin-only warning

Also once #3391 is in, can you regenerate the migrations so we get clean sqlite migrations instead of the rebuild churn for `user_push_subscriptions`.

Also, should this be Sonarr-only? Media-server-only setups should get this too right? jellyfin/emby/plex with library set to tmdb, seerr set to tmdb, no sonarr, should also get this feature no? Same for tvdb/tvdb. Plex/Jellyfin already pull per-episode number + file data for their season counts, and `episodeDetails`/`syncEpisodeDetails` in baseScanner isn't Sonarr-specific, so nothing stops wiring them in the same way.

server/routes/tv.ts
0xSysR3ll: Yep right, `{}` is truthy here.

server/entity/Episode.ts
0xSysR3ll: Yeah you're right, that should match `Season.media`; can't remember why I made this `nullable`.

src/components/Settings/SettingsMain/index.tsx
0xSysR3ll: We could yes.

server/entity/Episode.ts
0xSysR3ll: Right. Oversight on my part.

fallenbagel:
@0xSysR3ll is this stack now ready for review? As in all the changes I requested added? If so could you resolve the merge conflicts so I can re-review again. I want to get this merged for next version

0xSysR3ll:
> @0xSysR3ll is this stack now ready for review? As in all the changes I requested added? If so could you resolve the merge conflicts so I can re-review again. I want to get this merged for next version

IIRC it should yes. Will do the rebase tonight. It will still be missing the updated migrations though.

server/interfaces/api/settingsInterfaces.ts
```diff
@@ -41,6 +41,7 @@ export interface PublicSettingsResponse {
   mediaServerType: number;
   partialRequestsEnabled: boolean;
   enableSpecialEpisodes: boolean;
+  enableEpisodeAvailability: boolean;
```
fallenbagel: I dont see anything on the frontend using this off of public settings? 🤔
The badge runs off `episode.available` from the season route, and the settings form reads `/api/v1/settings/main`. Since its not used then we should drop it here and in `server/lib/settings/index.ts` (line :211, line :736), `src/context/SettingsContext.tsx` and `src/pages/_app.tsx`.

server/routes/tv.ts
```diff
@@ -84,7 +87,59 @@ tvRoutes.get('/:id/season/:seasonNumber', async (req, res, next) => {
...
+                metadataEpisodeCount: metadataEpisodeNumbers.size,
+                trackedEpisodeCount: trackedEpisodes.length,
+              }
+            );
+          } else {
+            availableMap = {};
+            for (const episode of trackedEpisodes) {
+              availableMap[episode.episodeNumber] =
+                episode.status === MediaStatus.AVAILABLE;
+            }
```
fallenbagel, line 140: only status is read here. Do we need a 4k status to be read here?

fallenbagel (changes requested):

server/entity/Episode.ts
fallenbagel, line 1: This needs to be imported in server/datasource.ts  added to the entities array

fallenbagel (changes requested):

server/interfaces/api/settingsInterfaces.ts
0xSysR3ll: Yeah IIRC it was used before. I'll remove this.

server/routes/tv.ts
0xSysR3ll: Depends if we want this now and in this PR or as a follow-up.

fallenbagel: > Depends if we want this now and in this PR or as a follow-up.

I might be blind but I didn't see it in the upstack PRs (I'll check again later)

0xSysR3ll: I didn't do anything about it that's why I ask :p

fallenbagel: > I didn't do anything about it that's why I ask :p

I think fine to add here tbh since when this lands all three prs will come together anyways. The only reason for the stack is for reviewability

server/lib/availabilitySync.ts
gauthier-th, line 1: Not sure to get why this has been moved?

server/routes/tv.ts
```diff
@@ -84,7 +87,65 @@ tvRoutes.get('/:id/season/:seasonNumber', async (req, res, next) => {
...
+          season.episodes.map((episode) => episode.episode_number)
+        );
+
+        if (trackedEpisodes.length > 0 && metadataEpisodeNumbers.size > 0) {
+          const hasEpisodeNumberMismatch = trackedEpisodes.some(
+            (episode) => !metadataEpisodeNumbers.has(episode.episodeNumber)
+          );
+
+          if (hasEpisodeNumberMismatch) {
+            logger.debug(
```
gauthier-th, line 122: Do we want debug or warning?

server/routes/tv.ts
```diff
@@ -72,9 +74,10 @@ tvRoutes.get('/:id/season/:seasonNumber', async (req, res, next) => {
     const tmdbTv = await tmdb.getTvShow({
       tvId: Number(req.params.id),
     });
-    const metadataProvider = tmdbTv.keywords.results.some(
+    const isAnime = tmdbTv.keywords.results.some(
       (keyword: TmdbKeyword) => keyword.id === ANIME_KEYWORD_ID
-    )
+    );
+    const metadataProvider = isAnime
```
gauthier-th, line 80: Same, why this change?

0xSysR3ll: Yep, that's leftover from the old tracking gate. It's useless now.

server/lib/availabilitySync.ts
0xSysR3ll: Good catch. It got pulled up while moving episode tracking out of availability sync. I'll revert.

server/routes/tv.ts
0xSysR3ll: It used to be a warn, but it was triggering on normal cases and getting logged on every season-tab request. So while fixing that, I figured debug made more sense.

0xSysR3ll: wdyt @fallenbagel ?

fallenbagel: > It used to be a warn, but it was triggering on normal cases and getting logged on every season-tab request. So while fixing that, I figured debug made more sense.

Cant we just gate it to if enableEpisode and episodeMismatch then warn?

0xSysR3ll: That's already the case, just had to revert to warn.

gauthier-th (changes requested):
Can you please regenerate the migrations?
Your migrations are behind #3510, it may cause conflicts.

PR #3367, opened 2026-08-11
perf: bound tmdb cache & split scan lookups into their own tier

### LINKED ISSUE #3307: Library scans blow out the V8 heap and kill the process (still happens at mem_limit 1g)
## Description

Library scans keep blowing out the heap and killing the process.

I started out at `mem_limit: 512m`. Bumping it to `1g` helped but didn't actually fix anything, it still dies roughly once a day.

Reason the limit matters so much: Node 22 reads the cgroup limit and sets `max-old-space-size` to about half of it. So `512m` gets you a ~256MB heap ceiling and `1g` gets you ~524MB. Both are easy to hit.

What made this take me so long is that it doesn't look like a crash from the outside. `docker inspect` said `ExitCode=0`, `OOMKilled=false`,
`Health=healthy`, `RestartCount=0` the entire time. I had 60 restarts in 16 days and nothing in the docker state said why.

When I finally counted the two mechanisms separately over that window I got 59 V8 heap aborts and 4 cgroup OOM kills. The cgroup ones only show up in `dmesg` on the host, not inside the container. So it's almost entirely V8 doing the killing, the kernel is barely involved.

Couple of debugging notes since both of these wasted my time:

* `FATAL ERROR` goes to stderr with no timestamp on it, so it doesn't interleave  with seerr's normal timestamped output. `grep -B3 FATAL` gets you nothing useful. `docker logs -t` stamps everything including stderr, use that.
* Don't grep for just `FATAL`. A lot of movie titles have "Fatal" in them. I got 16 hits and only 3 of them were real.

### What I found

Radarr, Sonarr and Jellyfin scanners each pull a whole library in one call:

* `server/lib/scanners/radarr/index.ts:80` - `this.items = await this.radarrApi.getMovies()`
* `server/lib/scanners/sonarr/index.ts:88` - `this.items = await this.sonarrApi.getSeries()`
* `server/lib/scanners/jellyfin/index.ts:524` - `this.items = await this.jfClient.getLibraryContents(library.id)`

I measured the responses from inside the container. Radarr gives me 16,451
items, 79.7MiB on the wire, 75.8MiB once parsed. Here's a Radarr scan kicked off from a fresh restart:

```
BASELINE 165.4MiB
12:10:11 166.0MiB
12:10:22 499.6MiB   <- one 10s sample later
12:10:33 243.6MiB
...      sits at 250-275MiB for the rest of the scan
```

That's +335MiB inside a single sampling interval, right at the start, before it processes a single item. Makes sense, the parse is holding the raw response string and the object graph at the same time. On a 256MB ceiling that alone kills you. The Plex scanner doesn't do this, it pages through with`paginateLibrary`.

Separate thing, and this one I understand less. RSS doesn't drop back down when a Plex Full Library Scan finishes:

```
01:27:14 178.5MiB   scanning, movie libraries, flat
01:31:12 198.7MiB
01:54:01 Full Scan Complete
02:45:27 608.6MiB   <- 50 minutes after it finished, nothing running
02:51:15 605.5MiB
(still around 621MiB at 09:00, 7+ hours later, idle)
```

So about 450MiB just stays there, and it's still there hours later.

That's what actually kills it, I think. The fetch spike on its own survived fine on a fresh process. But once seerr is already sitting at ~620MiB of leftovers, the next scheduled scan hits the same spike and there's no room for it:

```
10:00:00  radarr-scan fires (cron is 04:00 local)
10:00:09  FATAL ERROR: Reached heap limit          <- 9 seconds in
10:00:13  Starting Seerr version 3.4.0
10:30:00  sonarr-scan fires (cron is 04:30 local)
10:34:25  FATAL ERROR: Ineffective mark-compacts near heap limit
10:34:29  Starting Seerr version 3.4.0
```

The numbers line up with that. From a clean 165MiB start, 165 + 335 is about 500, which fits under 524 and my manual Radarr scan survived fine. From 620MiB of leftovers, 620 + 335 is about 955, and the scheduled one died 9 seconds in, right at the fetch.

It's pretty repeatable too. Manual Sonarr scan died 4m50s in, the scheduled one died 4m25s in.

Fixing either one would probably stop my box crashing. But the leftover memory is what'll bite me again once the library gets bigger, so that's the one I care about more.

### Stuff that might help

1. Page or stream the Radarr/Sonarr/Jellyfin fetches like `paginateLibrary` already does for Plex. Gets rid of the spike.
2. Figure out what a full scan is holding onto.
3. `BUNDLE_SIZE` in `server/lib/scanners/baseScanner.ts` is hardcoded to `20`, and each bundle runs concurrently through `Promise.all`. There's no setting for it, so if you hit this there's no way to trade scan speed for memory. Only lever you've got is raising the container limit, and that just scales with library size. I noticed #3241 adds an update rate override for testing, might be somewhere obvious to hang this off.
4. Whatever happens with the rest, a minimum `mem_limit` in the docs would help. The way Node sizes its heap makes `512m` kind of a trap.

Happy to test patches, I've got a box that reproduces this every day.

## Version

3.4.0 (commit `2dbe8860179e83ed2044d5fc8ed17679ae05e5ad`)

## Steps to Reproduce

1. Run seerr in docker on Node 22, `mem_limit: 1g`, no `NODE_OPTIONS` set.
2. Hook up a Radarr with a big library. Mine's 16,451 movies.
3. Check what heap ceiling you actually got:
   `docker exec seerr node -e "console.log(require('v8').getHeapStatistics().heap_size_limit/1048576)"`
   512m gives ~256, 1g gives ~524.
4. Restart the container so you've got a clean baseline, then run **Radarr Scan** from Settings > Jobs with `docker stats seerr` open. RSS jumps by roughly the parsed size of your Radarr response within one sample.
5. For the actual crash: let it run a **Plex Full Library Scan**, check RSS is still high after `Full Scan Complete`, then fire off a Radarr or Sonarr scan.
   `docker logs -t seerr` gives you `FATAL ERROR: Reached heap limit` and then`Starting Seerr version 3.4.0` a few seconds later, while `docker inspect` is still telling you `ExitCode=0` and `OOMKilled=false`.

## Screenshots

n/a, server side

## Logs

```
2026-07-29T10:00:09Z FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
2026-07-29T10:00:13Z [info]: Starting Seerr version 3.4.0
2026-07-29T10:34:25Z FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
2026-07-29T10:34:29Z [info]: Starting Seerr version 3.4.0

$ docker inspect seerr --format '{{.State.ExitCode}} {{.State.OOMKilled}} {{.RestartCount}}'
0 false 0

response sizes, measured from inside the container:
sonarr /api/v3/series: items=2352   wire=9.1MiB   heapDeltaOnParse=8.6MiB
radarr /api/v3/movie:  items=16451  wire=79.7MiB  heapDeltaOnParse=75.8MiB
```

## Platform

desktop

## Database

SQLite (default)

## Device

Beelink mini PC, running Proxmox, seerr is in docker inside an LXC

## Operating System

Debian 12 (LXC), Docker 29.6.1, Node 22.22.2

## Browser

Firefox

## Additional Context

Image is `ghcr.io/seerr-team/seerr:latest`. Library sizes are 16,451 movies in Radarr, 2,352 series in Sonarr, and two Plex libraries, one movies one TV.

The bulk fetch thing I'm sure about. It lands in one sample, right when the scan kicks off, and the response size accounts for how big the jump is.

The leftover memory I'm much less sure about. I know it's there and I know it persists, but I didn't sample continuously through the whole scan so I can't tell you which part of it accumulates, and I haven't worked out what's actually being held. I did chase down two of my own guesses and both were wrong:
`AsyncLock` cleans its keys up properly on release, and the recursive pagination doesn't seem to hold onto pages, since the Plex movie path runs that same code
and stays flat.

One last thing. Setting `NODE_OPTIONS=--max-old-space-size=...` on its own made things worse for me. All it did was turn a clean V8 abort I could grep for into
a kernel OOM kill, which was harder to work out. If you set it you have to move the container limit with it.

AI disclosure: I used an AI assistant to help collect and format the measurements here and to read through the scanner code. The numbers are all from my own instance, I ran them myself, and I've gone over everything above.

## Search Existing Issues

  Closest I found was #513 from 2023, but that's a request approval loop, different thing. Couldn't find anything open about scanner memory.

## Code of Conduct

fallenbagel:
## Description

Library scans and availability sync fetched the full `/tv/:id` payload including `aggregate_credits` and cached it for 12 hours in the same unbounded cache the interactive TV detail page uses, so a full scan left most of a library resident until the TTL expired. Scanners and availability sync now request only `keywords` and `external_ids`, which is everything they read, and write to a separate LRU-bounded cache with a 15 minute TTL, so scan lookups age out on their own instead of evicting manual browsing traffic. The interactive cache gains a key-count backstop against unbounded growth, and `aggregate_credits.crew` and `credits.cast` are stripped before caching since nothing in the codebase reads either.

Both tiers are capped by key count rather than bytes. There is no per-entry ceiling to derive for the interactive tier because TMDB doesn't cap the credit rosters, and for the scan tier the TTL is what actually bounds occupancy: a larger library takes proportionally longer to walk rather than holding proportionally more at once.

Separately, the recently added scan asked Plex for items without `includeGuids=1`, unlike the full scan, so every newly added movie fell back to a per-item `/library/metadata` request on each pass while it sat inside the scan's ten minute overlap window. That flag is now set on the movie query only as the show query is left alone because it returns episodes, whose guids the scanner never reads. The `plexguid` cache is removed along with it as it was keyed by rating key, its cache-hit branch never actually skipped anything since the metadata fetch below it ran unconditionally and overwrote the cached ids, and with guids arriving inline the only thing it still did, reusing an external-id resolve, is already handled by the scan cache tier. Every scan, full or recently added, now derives ids from what Plex reports at that moment, so a corrected match in Plex is picked up on the next scan exactly as before.

In addition, node-cache was entirely dropped in favour of lru-cache.

- Fixes #3307

## How Has This Been Tested?

Tested on two instances: mine (with just my tv library enabled), and the reporter of #3307, where the crash in in that issue was reproducible nightly.

On my instance, heap snapshots over Chrome DevTools Protocol with the same library both pre- and post- fix:

| | before | after |
|---|---|---|
| idle | 110 MB | 107 MB |
| post-sync | 270 MB | 124 MB |
| +30 min, forced GC | 269 MB | 121 MB |

Cache counters were identical across both runs (33 hits, 482 misses, 482 keys), so both pre- and post- fix resolved the same set of shows. The scan tier drained from 482 keys to 11 over the 30 minutes so TTL is also working as intended.

The reporter ran the same sequence from a fresh container on each pre- and post- fix, `used_heap_size` after two forced collections:

| | develop | this branch |
|---|---|---|
| idle after restart | 99.68 MiB | 98.83 MiB |
| after availability sync | 451.22 MiB | 128.47 MiB |
| +30 min, forced GC | 452.63 MiB | 116.01 MiB |
| Sonarr scan from that state | died at 2m28s | completed in 4m46s |

The sync took 810s on develop and 840s on the branch, so both pre- and post- fix did the same work. Afterwards the `tmdb` cache on develop held 175.03 MiB of JSON against 7.29 MiB in `tmdbscan` on the branch, and the scan tier had drained to zero keys by the 30 minute probe.

They have since been running it without restarts and without heap aborts, sitting at 129.89 MiB after an unattended job block, on the same box that aborted five nights running.

## Checklist:

## Performance

* Improved TV metadata lookup speed with bounded caching and automatic expiration.
* Reduced unnecessary metadata processing during library scans.

## Bug Fixes

* Improved TV show identification across Jellyfin, Plex, and Sonarr scans.
* Added more reliable TVDB, IMDb, and AniDB resolution during scanning.
* Improved matching accuracy when resolving external identifiers.
* Improved handling of missing or incomplete TV metadata during scans.

### DIFF
--- a/server/api/themoviedb/index.ts
+++ b/server/api/themoviedb/index.ts
@@ -963,12 +963,14 @@ class TheMovieDb extends ExternalAPI implements TvShowProvider {
   public async getByExternalIdForScan(
     lookup: ExternalIdLookup
   ): Promise<TmdbExternalIdResponse> {
-    return this.findByExternalId(lookup, this.scanCache);
+    return this.findByExternalId(lookup, this.scanCache, SCAN_TTL);
   }
 
+  // Omitting ttl here silently falls back to DEFAULT_TTL, regardless of tier.
   private async findByExternalId(
     { externalId, type, language = this.locale }: ExternalIdLookup,
-    cache?: CacheStore
+    cache?: CacheStore,
+    ttl?: number
   ): Promise<TmdbExternalIdResponse> {
     try {
       return await this.get<TmdbExternalIdResponse>(
@@ -979,7 +981,7 @@ class TheMovieDb extends ExternalAPI implements TvShowProvider {
             language,
           },
         },
-        undefined,
+        ttl,
         { cache }
       );
     } catch (e) {
@@ -998,7 +1000,8 @@ class TheMovieDb extends ExternalAPI implements TvShowProvider {
     try {
       const extResponse = await this.findByExternalId(
         { externalId: imdbId, type: 'imdb' },
-        this.scanCache
+        this.scanCache,
+        SCAN_TTL
       );
 
       if (extResponse.movie_results[0]) {
@@ -1037,11 +1040,13 @@ class TheMovieDb extends ExternalAPI implements TvShowProvider {
 
   private async resolveTvdbId(
     tvdbId: number,
-    cache?: CacheStore
+    cache?: CacheStore,
+    ttl?: number
   ): Promise<number> {
     const extResponse = await this.findByExternalId(
       { externalId: tvdbId, type: 'tvdb' },
-      cache
+      cache,
+      ttl
     );
 
     if (!extResponse.tv_results[0]) {
@@ -1080,7 +1085,7 @@ class TheMovieDb extends ExternalAPI implements TvShowProvider {
   }): Promise<TmdbTvScanDetails> {
     try {
       return await this.getTvShowForScan({
-        tvId: await this.resolveTvdbId(tvdbId, this.scanCache),
+        tvId: await this.resolveTvdbId(tvdbId, this.scanCache, SCAN_TTL),
         language,
       });
     } catch (e) {
--- a/server/api/themoviedb/index.ts
+++ b/server/api/themoviedb/index.ts
@@ -128,13 +128,25 @@ interface DiscoverTvOptions {
 
 const TV_DETAILS_TTL = 43200;
 
-// Scan lookups read each show once and never again.
-const TV_SCAN_TTL = 900;
+// Scan lookups read an entry once and never again.
+const SCAN_TTL = 900;
 
 const TV_DETAILS_APPEND_TO_RESPONSE =
   'aggregate_credits,credits,external_ids,keywords,videos,content_ratings,watch/providers';
 
 const TV_SCAN_APPEND_TO_RESPONSE = 'keywords,external_ids';
+
+type ExternalIdLookup =
+  | {
+      externalId: string;
+      type: 'imdb';
+      language?: string;
+    }
+  | {
+      externalId: number;
+      type: 'tvdb';
+      language?: string;
+    };
 
 // Nothing anywhere reads aggregate_credits.crew or credits.cast.
 const stripUnreadTvCredits = <T>(data: T): T => {
@@ -468,7 +480,7 @@ class TheMovieDb extends ExternalAPI implements TvShowProvider {
         tvId,
         language,
         appendToResponse: TV_SCAN_APPEND_TO_RESPONSE,
-        ttl: TV_SCAN_TTL,
+        ttl: SCAN_TTL,
         cache: this.scanCache,
       });
     } catch (e) {
@@ -942,33 +954,34 @@ class TheMovieDb extends ExternalAPI implements TvShowProvider {
     }
   };
 
-  public async getByExternalId({
-    externalId,
-    type,
-    language = this.locale,
-  }:
-    | {
-        externalId: string;
-        type: 'imdb';
-        language?: string;
-      }
-    | {
-        externalId: number;
-        type: 'tvdb';
-        language?: string;
-      }): Promise<TmdbExternalIdResponse> {
+  public async getByExternalId(
+    lookup: ExternalIdLookup
+  ): Promise<TmdbExternalIdResponse> {
+    return this.findByExternalId(lookup);
+  }
+
+  public async getByExternalIdForScan(
+    lookup: ExternalIdLookup
+  ): Promise<TmdbExternalIdResponse> {
+    return this.findByExternalId(lookup, this.scanCache);
+  }
+
+  private async findByExternalId(
+    { externalId, type, language = this.locale }: ExternalIdLookup,
+    cache?: CacheStore
+  ): Promise<TmdbExternalIdResponse> {
     try {
-      const data = await this.get<TmdbExternalIdResponse>(
+      return await this.get<TmdbExternalIdResponse>(
         `/find/${externalId}`,
         {
           params: {
             external_source: type === 'imdb' ? 'imdb_id' : 'tvdb_id',
             language,
           },
-        }
+        },
+        undefined,
+        { cache }
       );
-
-      return data;
     } catch (e) {
       throw new Error(`[TMDB] Failed to find by external ID: ${e.message}`, {
         cause: e,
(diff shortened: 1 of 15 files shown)


fallenbagel:
Ignore the failing cypress tests. That's unrelated to this PR and not relevant to users and is fixed by
https://github.com/seerr-team/seerr/pull/3368 anyways.

server/api/themoviedb/index.ts
```diff
@@ -871,69 +954,65 @@ class TheMovieDb extends ExternalAPI implements TvShowProvider {
...
+      return await this.get<TmdbExternalIdResponse>(
         `/find/${externalId}`,
         {
           params: {
             external_source: type === 'imdb' ? 'imdb_id' : 'tvdb_id',
             language,
           },
-        }
+        },
+        undefined,
```
0xSysR3ll: Shouldn't we pass `SCAN_TTL` when  `cache === scanCache` instead ?
Right now `undefined` falls through to the 300s default, so `/find/{external_id}` entries on the scan tier expire way earlier than the rest of `tmdbscan`.

fallenbagel: yup. that was an oversight. `findByExternalId` is shared between the interactive and scan paths, and we currently have  the interactive `/find` call never passing an explicit ttl and relying on the 300s default. When I split it out to add the cache param glossed over it unlike the threading I did for `getTvShowDetails`. Should be fixed now.

0xSysR3ll (dismissed):
LGTM

server/lib/cache.ts
```diff
@@ -15,22 +17,175 @@ export type AvailableCacheIds =
...
+
+    const stdTtl = options.stdTtl ?? DEFAULT_TTL;
+
+    this.data = options.max
+      ? new LruCacheStore({ max: options.max, stdTtl })
+      : new NodeCache({
+          stdTTL: stdTtl,
+          checkperiod: options.checkPeriod ?? DEFAULT_CHECK_PERIOD,
+        });
   }
```
gauthier-th, line 207: I'm not a big fan of having 2 different cache systems being silently used depending only on a `max` parameter. Changing the core caching system should be explicit instead of implicitly deducted from this unrelated parameter.

Couldn't you add something like a `type` enum for the cache with the required parameters depending on the type?
Something like this:
```ts
type CacheOptions = {
  type: "node";
  stdTtl?: number;
  checkPeriod?: number;
} | {
  type: "lru";
  stdTtl?: number;
  max: number;
};
```

gauthier-th (changes requested):

server/lib/cache.ts
fallenbagel: Or how about we replace node cache entirely with lru?

gauthier-th: Yes that would be even better. `lru-cache` can do everything `node-cache` does, and it seems that `node-cache` is not maintained anymore (and less used than `lru-cache`).

fallenbagel: > Yes that would be even better. `lru-cache` can do everything `node-cache` does, and it seems that `node-cache` is not maintained anymore (and less used than `lru-cache`).

Yeah let me replace that right now and push

fallenbagel:
Ignore the failing cypress test. It will be fixed on:
https://github.com/seerr-team/seerr/pull/3432

gauthier-th (approved):
LGTM

0xSysR3ll (approved):

PR #3375, opened 2026-08-12
fix(datasource): register entities and subscribers explicitly

fallenbagel:
## Description

TypeORM entities and subscribers load from directory globs over every `.ts` file under `server/entity` and `server/subscriber`. While it is **harmless** today since neither folder has a test file, but drop one in and requiring it runs the module body, which calls `setupTestDb()` again. That registers a second root `before` and `beforeEach` after the runner already built its tree, so two schema rebuilds collide on the single in-memory sqlite connection: "cannot start a transaction within a transaction," every test cancelled, nothing in the failure pointing at the actual file or the actual error.

**Production is worse.** `build:server` compiles every `.ts` under `server`, test files included, and production loads `dist/entity/**/*.js`, so a compiled test file there runs its module body at boot. `seedTestDb` calls `dropDatabase()` on the shared `dataSource` with no check on which database it's pointing at, then rebuilds the schema and seeds `admin@seerr.dev` with `Permission.ADMIN` and a hardcoded bcrypt hash. I confirmed this against a throwaway sqlite file under `NODE_ENV=production` and it dropped a table I'd put there first, rebuilt 15 tables, and left those two accounts behind.

> [!important]
> **No released version is affected.** No tag ships a test file under `dist/entity` or `dist/subscriber`, so there's nothing to fix or disclose.

Now, `seedTestDb` and `resetTestDb` refuse to run unless `NODE_ENV` is test. `cypress:prepare`, now opts in via `seedTestDb({ allowOutsideTest: true })`. This deny-unless-permitted was added because a plain node process with no `NODE_ENV` set resolves to `devConfig` against `config/db/db.sqlite3`, someone's real database if they're running from source.

In addition, `build:server` points at a new `server/tsconfig.build.json` that excludes `*.test.ts`, so `dist` stops shipping the compiled test files it currently produces. Config was separated because because `typecheck:server` is the only thing type checking those files, and the root `tsconfig` covers `src` only and excluding at the base would've silently dropped the files from CI.

Lastly, typeORM gets the `entity` and `subscriber` classes **directly instead of globs**, which also drops the duplicated `ts`/`js` patterns across the five config variants. Narrowing the globs to skip `*.test.ts` would achieve the same thing provided the files named are named that way in directories and a differently named test file, a helper, or a newly globbed directory brings the whole thing back. Classes mean nothing is matched by pattern, at the cost of a list to keep current, forget to add an `entity` and it throws` EntityMetadataNotFoundError` the first time its repository is used anyways. Much more safer this way. Migrations stay globbed.

## How Has This Been Tested?

- Checked the dist if the test files were carried after build
- Checked if the tests still run
- Checked if entities and subscribers were properly loaded.

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database configuration reliability across supported environments.
  * Added configurable PostgreSQL connection timeouts, defaulting to 30 seconds.
  * Prevented accidental test database seeding or resets outside test environments.

* **Build**
  * Updated server builds to exclude test files from production compilation.

* **Tests**
  * Added coverage verifying safeguards against unsafe test database operations.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "preinstall": "npx only-allow pnpm",
     "postinstall": "next telemetry disable",
     "dev": "nodemon -e ts,json,yml --watch server --watch seerr-api.yml --exec 'ts-node -r tsconfig-paths/register --files --project server/tsconfig.json server/index.ts'",
-    "build:server": "tsc --project server/tsconfig.json && copyfiles -u 2 server/templates/**/*.{html,pug} dist/templates && copyfiles -u 2 \"server/i18n/locale/*.json\" dist/i18n && tsc-alias -p server/tsconfig.json",
+    "build:server": "tsc --project server/tsconfig.build.json && copyfiles -u 2 server/templates/**/*.{html,pug} dist/templates && copyfiles -u 2 \"server/i18n/locale/*.json\" dist/i18n && tsc-alias -p server/tsconfig.build.json",
     "build:next": "next build",
     "build": "pnpm build:next && pnpm build:server",
     "lint": "eslint \"./server/**/*.{ts,tsx}\" \"./src/**/*.{ts,tsx}\" --cache",
--- a/server/datasource.ts
+++ b/server/datasource.ts
@@ -1,9 +1,51 @@
+import { Blocklist } from '@server/entity/Blocklist';
+import DiscoverSlider from '@server/entity/DiscoverSlider';
+import Issue from '@server/entity/Issue';
+import IssueComment from '@server/entity/IssueComment';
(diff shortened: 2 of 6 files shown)


src/components/Settings/SettingsNetwork/index.tsx
fallenbagel, line 1: This was supposed to be in:
#3368

I accidentally pushed to this PR and not to #3368...
I have been wondering where this fix went as #3367 's cypress test fail because this is missing lol

gauthier-th (approved):

M0NsTeRRR (approved):

PR #3379, opened 2026-08-12
fix(requests): stop editing a request from re-requesting covered seasons

fallenbagel:
## Description

Creating a request skips seasons the media already has, but editing one only looked at seasons other requests hold. An edit could therefore add a season that was already available and charge the owner's quota for it.

The availability check applies only to seasons being added, not to the ones the request already holds. This matters because a request's own seasons stop being unknown as soon as it is approved, so treating them the same way would drop them from the request the moment they started processing.

This sits on the quota check PR below it and updates its arithmetic, which until now could assume the edit ends up with every season it asked for. That is no longer true once covered seasons are dropped, so the quota is charged against the seasons the edit actually ends up with.

## How Has This Been Tested?

- Via the attached unit test only

## Checklist:

## Summary by CodeRabbit

- **Improvements**
  - TV season requests now skip seasons already available or covered by existing requests.
  - Existing selected seasons are preserved when adding new seasons.
  - Season updates retain selected seasons and add newly requested seasons as pending.
  - Availability and quota calculations account only for genuinely new seasons, including edits that add and remove seasons.
  - Requests with no new seasons return successfully without unnecessary processing.
  - Movie ownership changes now respect quota limits.
  - Request status changes enforce valid transitions and prevent updates to pending requests.
  - Concurrent request operations are handled safely to prevent conflicting updates.


0xSysR3ll (dismissed):

M0NsTeRRR (dismissed):

0xSysR3ll (dismissed):

M0NsTeRRR (dismissed):

0xSysR3ll (dismissed):

gauthier-th (dismissed):
LGTM

M0NsTeRRR (approved):

gauthier-th (approved):

Issue #579, opened 2023-11-28
Filter Release Date Tied To Specific Type of Release


halexh:
### Description

Within the Filter section, there is a Release Date range in order to filter Movies that have been released within a certain date range. It would be nice to further specify whether this release date should apply to theatrical release or digital release.

### Desired Behavior

1. Select Movies
2. Select Filters
3. Select Date Range
4. Select some type of checkbox implying this range is for a digital release
5. Filter movies that have had a digital release within the supplied date range

### Code of Conduct

- [X] I agree to follow Jellyseerr's Code of Conduct

PierrePetit:
First of all, thanks for this amazing tool !

To add on what @halexh said :
This would be great to help people find movies that are available straight away. Some of my users are confused as for the availability of the movies they ask for.

A great first step would be to add the filter mentioned. I think also adding a category for "New Releases (At Home)" (and maybe "New Releases (Theaters)") in the "Discover" section would be a great UX improvement, to quickly find something to watch.

Furthermore, adding the different release dates (theatrical, premiere, physical, digital...) on a movie's page would be fantastic. These data are available on TMDB (even though, I have to admit, it's not the most regularly updated field).
There currently is a "status" field in jellyseerr's movies, maybe it would be possible to specify "released (Theaters)" or "released (at Home)" with subcategories as per TMDB's fields :

![image](https://github.com/Fallenbagel/jellyseerr/assets/4937633/8fbbf05f-2eb7-42b7-9e2e-a39ca764e212)

In my view, it would be the most useful update, with the OIDC one, but probably a bit simpler to make 😄 
Thanks for reading.

Bretterteig:
Love @PierrePetit recommendation. I'd add that it would be great to have this be configurable as server default similar to region and language.

Fallenbagel:
> I just saw that some movies had a digital release date in jellyseerr (ex: the three musketeers : D'Artagnan (2023)). The digital release date is an information available on rotten tomatoes. Maybe, adding TMDB's data on top of rottentomatoe's ones could be better ? Rottentomatoes really rarely has this information.

For your information, we don't use rotten tomatoes to get the release dates. Rotten tomatoes are only used to get the ratings. And that's it.

Everything else is tmdb. Including the digital release date of your mentioned movie which also comes from tmdb.

[...]

PR #3384, opened 2026-08-12
feat(discover): filter movie discovery by minimum release

demigodmode:
## Description

Adds a "Minimum Release" filter to the movie discover filters (part of #579). You pick a release stage from the dropdown and it shows movies that have reached at least that stage. Picking Digital also covers Physical and TV, since those are all effectively "available at home."

The fiddly bit is that TMDB's `with_release_type` does nothing on its own. It only filters when it's paired with a `release_date` range, and it works off `release_date` rather than the `primary_release_date` field the existing Release Date filter uses. So when a Minimum Release is selected, the query switches its date range over to `release_date` and defaults the "To" date to today if you haven't set one. With nothing selected, the query is unchanged and the normal Release Date filter behaves exactly as before.

Keeping this to movie discover for a first pass. Related to #579 but not closing it, since that issue also covers things like release-status badges on movie pages.

- refs #579 

AI disclosure: I originally wrote the first version of this filter myself. For this iteration I had some AI assistance to help rework it, mainly the cumulative release-type logic and to code-review the TMDB query behavior and the diff. The review caught a timezone bug I missed and an unrelated translation change, both fixed. I've reviewed and tested all the changes myself and understand how they work.

## How Has This Been Tested?

Ran it locally against a Jellyfin instance:
- The dropdown shows up in the movie filters, and the "To" date auto-fills to today when a level is picked.
- Results change by level. Checked against TMDB's discover endpoint directly: digital-and-up (`with_release_type=4|5|6` + `release_date.lte=<today>`) returns ~314k vs ~1.16M unfiltered, and narrows as the floor goes up (theatrical ~1.05M, TV ~105k). No region needed.
- With no Minimum Release selected, the existing Release Date filter still uses `primary_release_date` and behaves the same.
- `tsc` (client + server) and `eslint` pass.

Screenshot 2026-08-12 152506

Screenshot 2026-08-12 152354

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Added release-type filtering to movie discovery.
  - Filter movies by digital, physical, limited theatrical, theatrical, or TV release.
  - Selected release types include later release types, with an automatic end date when needed.
  - Added a movie-only release-type selector to discovery filters.
- **Documentation**
  - Documented supported release types and filtering behavior.
- **Localization**
  - Added English labels for release-type options.
- **Tests**
  - Added coverage for filtering movies by digital release.


server/api/themoviedb/index.ts
```diff
@@ -89,6 +89,7 @@ interface DiscoverMovieOptions {
   keywords?: string;
   excludeKeywords?: string;
   sortBy?: SortOptions;
+  releaseType?: string;
```
SoulofAkuma, line 92:
```suggestion
  minimumReleaseType?: string;
```

server/api/themoviedb/index.ts
```diff
@@ -601,6 +602,7 @@ class TheMovieDb extends ExternalAPI implements TvShowProvider {
     voteAverageLte,
     voteCountGte,
     voteCountLte,
+    releaseType,
```
SoulofAkuma, line 605:
```suggestion
    minimumReleaseType,
```

server/api/themoviedb/index.ts
```diff
@@ -619,6 +621,21 @@ class TheMovieDb extends ExternalAPI implements TvShowProvider {
...
+      // of primary_release_date (the field the plain Release Date filter uses).
+      const releaseTypeFilter = releaseType
+        ? Array.from(
+            { length: 6 - Number(releaseType) + 1 },
+            (_, i) => Number(releaseType) + i
+          ).join('|')
+        : undefined;
+      const releaseDateField = releaseType
+        ? 'release_date'
+        : 'primary_release_date';
```
SoulofAkuma, line 637:
```suggestion
      const releaseTypeFilter = minimumReleaseType
        ? Array.from(
            { length: 6 - Number(minimumReleaseType) + 1 },
            (_, i) => Number(minimumReleaseType) + i
          ).join('|')
        : undefined;
      const releaseDateField = minimumReleaseType
        ? 'release_date'
        : 'primary_release_date';
```

seerr-api.yml
```diff
@@ -5595,6 +5595,19 @@ paths:
           schema:
             type: string
             example: 2023-01-01
+        - in: query
+          name: releaseType
```
SoulofAkuma, line 5599:
```suggestion
          name: minimumReleaseType
```

src/components/Discover/FilterSlideover/index.tsx
```diff
@@ -133,6 +140,49 @@ const FilterSlideover = ({
...
+                {intl.formatMessage(messages.limitedTheatrical)}
+              </option>
+              <option value="3">
+                {intl.formatMessage(messages.theatrical)}
+              </option>
+              <option value="4">{intl.formatMessage(messages.digital)}</option>
+              <option value="5">{intl.formatMessage(messages.physical)}</option>
+              <option value="6">
+                {intl.formatMessage(messages.tvRelease)}
+              </option>
```
SoulofAkuma, line 184: The `Premiere` release type is missing. Is that intentional?

server/routes/discover.ts
```diff
@@ -65,6 +65,7 @@ const QueryFilterOptions = z.object({
   sortBy: z.coerce.string().optional(),
   primaryReleaseDateGte: z.coerce.string().optional(),
   primaryReleaseDateLte: z.coerce.string().optional(),
+  releaseType: z.enum(['2', '3', '4', '5', '6']).optional(),
```
SoulofAkuma, line 68:
```suggestion
  minimumReleaseType: z.enum(['2', '3', '4', '5', '6']).optional(),
```

server/routes/discover.ts
```diff
@@ -65,6 +65,7 @@ const QueryFilterOptions = z.object({
   sortBy: z.coerce.string().optional(),
   primaryReleaseDateGte: z.coerce.string().optional(),
   primaryReleaseDateLte: z.coerce.string().optional(),
```
SoulofAkuma, line 67: Might be worth renaming these to `releaseDateGte` and `releaseDateLte` now that it is not necessarily the primary one.

server/routes/discover.ts
```diff
@@ -115,6 +116,7 @@ discoverRoutes.get('/movies', async (req, res, next) => {
       primaryReleaseDateGte: query.primaryReleaseDateGte
         ? new Date(query.primaryReleaseDateGte).toISOString().split('T')[0]
         : undefined,
+      releaseType: query.releaseType,
```
SoulofAkuma, line 119:
```suggestion
      minimumReleaseType: query.minimumReleaseType,
```

src/components/Discover/FilterSlideover/index.tsx
SoulofAkuma, line 184: The mapping between Release Message Name (e.g. `pysical`) - Release Type Value (e.g. `5`) should probably be stored somewhere in a centralized place and not in the HTML.

I would also add a link to the TMDB source https://developer.themoviedb.org/reference/movie-release-dates

src/components/Discover/FilterSlideover/index.tsx
```diff
@@ -133,6 +140,49 @@ const FilterSlideover = ({
...
+                // with_release_type only filters when paired with a date range,
+                // so default the "To" date to today if the user hasn't set one.
+                if (value && !currentFilters[dateLte]) {
+                  const now = new Date();
+                  const today = `${now.getFullYear()}-${String(
+                    now.getMonth() + 1
+                  ).padStart(2, '0')}-${String(now.getDate()).padStart(
+                    2,
+                    '0'
+                  )}`;
```
SoulofAkuma, line 161:
```suggestion
                  // Get today's date in ISO format YYYY-mm-dd
                  const today = new Date().toISOString().substr(0,10);
```

src/components/Discover/FilterSlideover/index.tsx
```diff
@@ -133,6 +140,49 @@ const FilterSlideover = ({
         </div>
         {type === 'movie' && (
           <>
+            <span className="text-lg font-semibold">
+              {intl.formatMessage(messages.minimumRelease)}
+            </span>
+            <select
+              id="releaseType"
+              name="releaseType"
+              value={currentFilters.releaseType ?? ''}
```
SoulofAkuma, line 149:
```suggestion
              id="minimumReleaseType"
              name="minimumReleaseType"
              value={currentFilters.minimumReleaseType ?? ''}
```

src/components/Discover/FilterSlideover/index.tsx
```diff
@@ -133,6 +140,49 @@ const FilterSlideover = ({
         </div>
         {type === 'movie' && (
           <>
+            <span className="text-lg font-semibold">
+              {intl.formatMessage(messages.minimumRelease)}
+            </span>
```
SoulofAkuma, line 145:
```suggestion
            <label 
                className="text-lg font-semibold" 
                htmlFor="minimumReleaseType"
             >
              {intl.formatMessage(messages.minimumRelease)}
            </label>
```

src/components/Discover/FilterSlideover/index.tsx
```diff
@@ -133,6 +140,49 @@ const FilterSlideover = ({
...
+                if (value && !currentFilters[dateLte]) {
+                  const now = new Date();
+                  const today = `${now.getFullYear()}-${String(
+                    now.getMonth() + 1
+                  ).padStart(2, '0')}-${String(now.getDate()).padStart(
+                    2,
+                    '0'
+                  )}`;
+                  batchUpdateQueryParams({
+                    releaseType: value,
```
SoulofAkuma, line 163:
```suggestion
                    minimumReleaseType: value,
```

src/components/Discover/FilterSlideover/index.tsx
```diff
@@ -133,6 +140,49 @@ const FilterSlideover = ({
...
+                    [dateLte]: today,
+                  });
+                } else {
+                  updateQueryParams('releaseType', value);
```
SoulofAkuma, line 167:
```suggestion
                  updateQueryParams('minimumReleaseType', value);
```

src/components/Discover/FilterSlideover/index.tsx
SoulofAkuma, line 77: Not sure what would be the right place to change the `primaryReleaseDate` to `releaseDate`. Here or in the function that does the TMDB API call.

SoulofAkuma, line 139: I would suggest adding a warning message under the date picker here if the `type` is a movie, a `minimumReleaseType` other than `Any` has been selected and both the `to` and `from` dates are empty. That warning message should indicate that this will cause the selected minimum release to have no effect.

src/components/Discover/constants.ts
```diff
@@ -94,6 +94,7 @@ export const QueryFilterOptions = z.object({
   sortBy: z.string().optional(),
   primaryReleaseDateGte: z.string().optional(),
   primaryReleaseDateLte: z.string().optional(),
+  releaseType: z.enum(['2', '3', '4', '5', '6']).optional(),
```
SoulofAkuma, line 97:
```suggestion
  minimumReleaseType: z.enum(['2', '3', '4', '5', '6']).optional(),
```

src/components/Discover/constants.ts
```diff
@@ -138,6 +139,10 @@ export const prepareFilterValues = (
     filterValues.primaryReleaseDateLte = values.primaryReleaseDateLte;
   }
 
+  if (values.releaseType) {
+    filterValues.releaseType = values.releaseType;
+  }
```
SoulofAkuma, line 144:
```suggestion
  if (values.minimumReleaseType) {
    filterValues.minimumReleaseType = values.minimumReleaseType;
  }
```

SoulofAkuma (commented):
Thanks for the PR, just my 2 cents. I am not a code owner, so we probably need an actual code owner to review this as well.

SoulofAkuma:
It'd probably be great if this can also have some testing. There is already a discover cypress suite where you might be able to add a test case or two (at least validating that selecting a certain value leads to the correct TMDB query being made)

demigodmode:
@SoulofAkuma good idea. added a case to the discover cypress suite that opens the filter, picks Digital, and checks the request goes out with releaseType plus the date bound. while i was at it i moved the date default into the server too, so it's not only the UI setting it.

seerr-api.yml
```diff
@@ -5595,6 +5595,19 @@ paths:
...
+            type: string
+            enum: ['2', '3', '4', '5', '6']
+            example: '4'
+          description: >-
+            Minimum release availability for movie discovery. The selected type
+            and everything above it are matched (e.g. 4 = digital also matches 5
+            = physical and 6 = TV, sent to TMDB as with_release_type=4|5|6).
+            When set, results are filtered on release_date instead of
+            primary_release_date. 2 = limited theatrical, 3 = theatrical,
+            4 = digital, 5 = physical, 6 = TV
```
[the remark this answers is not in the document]
demigodmode: done. the server now defaults the release_date upper bound to today whenever a release type is set, so it filters even if the caller doesn't pass primaryReleaseDateLte. updated the api description to match.

src/components/Discover/FilterSlideover/index.tsx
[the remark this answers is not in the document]
demigodmode: i kept this as a span on purpose. every other filter label in that slideover is a span too, so switching just this one would make it the odd one out. i can do a separate PR converting all of them to label elements if that's wanted.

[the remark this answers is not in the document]
demigodmode: fair point, changed it to a label with htmlFor. left the other filters alone for now, they can go in a separate pass.

cypress/e2e/discover.cy.ts
```diff
@@ -176,6 +176,27 @@ describe('Discover', () => {
...
+
+    // open the filters slideover and pick Digital (release type 4)
+    cy.contains('button', 'Active Filter').click();
+    cy.get('#releaseType').select('4');
+
+    // since the "To" date was empty it should default to today, otherwise
+    // with_release_type wouldn't actually filter anything
+    cy.wait('@getFiltered')
+      .its('request.url')
+      .should('include', 'primaryReleaseDateLte=');
```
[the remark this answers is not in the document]
demigodmode: good catch. it now builds today's date the same way the component does and asserts primaryReleaseDateLte equals it, so a wrong date won't slip through. skipped cy.clock since freezing timers tends to make these specs flaky.

demigodmode:
on the releaseType/tv scoping note: this one i'd leave as is. prepareFilterValues doesn't take a media type at all, so nothing is scoped that way. primaryReleaseDateGte/Lte and studio are movie only too, and they get preserved on tv exactly the same, with the tv route dropping all of them before it calls getDiscoverTv. so releaseType is behaving like the existing movie filters rather than doing something new. scoping only this one would mean changing prepareFilterValues' signature, so if you want them all scoped that feels like its own cleanup.

PR #3378, opened 2026-08-12
fix(requests): enforce the quota when editing a request

fallenbagel:
## Description

Editing a request never checked the requester's quota at all. A user could add seasons beyond their limit on a request that already existed, which is #633, and an admin could move a request onto a user with no room for it.

An edit is charged for every season `getQuota` did not already count against the owner. Reassignment pays in full because none of the request's seasons count towards the new owner yet, a declined request pays in full because declined requests are excluded from the count, and a request older than the quota window pays in full because the window filters on when the request was created. An ordinary edit inside the window pays only the difference. Requests created with the quota bypass keep it, and users whose quota is unlimited are unaffected.

The handler serializes on the request id from the url so both parties to a reassignment queue behind each other, then takes the owner lock from the PR below around the quota check and save. The prefix on that first key is load bearing, since a request id and a user id would otherwise stringify to the same key on the same lock instance.

- Fixes #633

## How Has This Been Tested?

- Via the attached unit test only

## Checklist:

- [X] I have read and followed the contribution [guidelines](https://github.com/seerr-team/seerr/blob/develop/CONTRIBUTING.md).
- [ ] Disclosed any use of AI (see our [policy](https://github.com/seerr-team/seerr/blob/develop/CONTRIBUTING.md#ai-assistance-notice))
- [ ] I have updated the documentation accordingly.
- [X] All new and existing tests passed.
- [X] Successful build `pnpm build`
- [ ] Translation keys `pnpm i18n:extract`
- [ ] Database migration (if required)

## Bug Fixes

- Prevented conflicting simultaneous request updates, deletions, approvals, and retries.
- Enforced movie and TV quotas when requests are reassigned.
- Recalculated TV season usage accurately when changing requested seasons or ownership.
- Preserved support for unlimited quotas and authorized quota bypasses.
- Prevented declined or reassigned requests from incorrectly affecting quota usage.
- Improved request status transitions and retry error handling.
- Prevented duplicate media-server actions during concurrent retries.
- Kept request details and selected seasons accurate after updates.
- Restored orphaned media and TV season statuses when requests are deleted.

PR #3385, opened 2026-08-13
fix(requests): enforce pending and failed states on request routes

fallenbagel:
## Description

A request is only editable while it is pending, and every state other than pending and failed are terminal. The frontend enforces that everywhere: the edit modal only opens from pending-gated triggers in `RequestItem`, `RequestCard`, `RequestBlock` and `RequestButton`, approve and decline only render while pending, and retry only renders on a failed request. 

While working on #3378 I noticed that the API enforces none of it, so PUT will happily rewrite an approved request, approve and decline will move a completed or declined one, and retry sets APPROVED regardless of what the request was before.

This PR adds the three guards the product already implies. PUT and the approve and decline route reject anything that is not currently pending, and retry rejects anything that has not failed, all with 409 and a message naming the state that was expected.

Nothing in the frontend changes. I checked every path that can reach these routes, including the edit then approve flow in the request modals, which is safe because both modals send the PUT before posting the approve, so the request is still pending at each step. The two `modifyRequest` helpers are typed to approve and decline only, so moving a request back to pending is not something the UI can ask for at all.

Worth nothing though that blocking transitions out of declined means an accidental decline cannot be undone through the API, since approve on a declined request now returns 409. That is already the case in the UI anyways, where a declined request offers no buttons, so this hardens a gap that already existed.

## How Has This Been Tested?

- Tested using unit tests

## Checklist:

## Bug Fixes

* Prevented updates to requests that are no longer pending.
* Restricted retries to failed requests and approvals or declines to pending requests.
* Rejected unsupported status values with clear error responses.
* Returned conflict responses when operations are invalid for the request’s current state.
* Preserved request data and status when conflicting operations are rejected.

### DIFF
--- a/server/routes/request.test.ts
+++ b/server/routes/request.test.ts
@@ -211,6 +211,27 @@ describe('PUT /request/:requestId (movie)', () => {
     assert.strictEqual(saved.profileId, 7);
     assert.strictEqual(saved.rootFolder, '/updated/movies');
   });
+
+  it('refuses to modify a request that is no longer pending', async () => {
+    const requestRepo = getRepository(MediaRequest);
+    const mediaRequest = await seedRequest(MediaRequestStatus.APPROVED);
+
+    const agent = await loginAs('admin@seerr.dev', 'test1234');
+    const res = await agent.put(`/request/${mediaRequest.id}`).send({
+      mediaType: MediaType.MOVIE,
+      serverId: 3,
+      rootFolder: '/updated/movies',
+    });
+
+    assert.strictEqual(res.status, 409);
+    assert.match(res.body.message, /pending/i);
+
+    const saved = await requestRepo.findOneOrFail({
+      where: { id: mediaRequest.id },
+    });
+    assert.strictEqual(saved.serverId, null);
+    assert.strictEqual(saved.rootFolder, null);
+  });
 });
 
 describe('POST /request/:requestId/:status', () => {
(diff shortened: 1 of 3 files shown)


server/routes/request.ts
```diff
@@ -675,18 +689,27 @@ requestRoutes.post<{
         relations: { requestedBy: true, modifiedBy: true },
       });
 
+      if (request.status !== MediaRequestStatus.PENDING) {
+        return next({
+          status: 409,
+          message: 'Only pending requests can be approved or declined.',
+        });
+      }
```
0xSysR3ll: I think we should move this check after the switch so that we can send the 400 error regardless of whether the request is `PENDING` or not. And it would be consistent with the tests you made.

server/routes/request.test.ts
```diff
@@ -241,6 +262,38 @@ describe('POST /request/:requestId/:status', () => {
...
+    assert.match(res.body.message, /approve/i);
+
+    const persisted = await repo.findOneOrFail({
+      where: { id: pending.id },
+      relations: { modifiedBy: true },
+    });
+    assert.strictEqual(persisted.status, MediaRequestStatus.PENDING);
+    assert.ok(!persisted.modifiedBy);
+  });
+
```
0xSysR3ll, line 283: Peetty nit but maybe worth a test that `POST /request/:id/pending` on a declined request is rejected ?

0xSysR3ll (commented):
Since these 409 errors are new, it would be nice to document them in the API specs.

0xSysR3ll (approved):

gauthier-th (commented):
LGTM

fallenbagel:
> LGTM

But no approval? 🤨

gauthier-th (approved):

PR #3399, opened 2026-08-15
fix(scanner): confirm orphan candidates against the servers before declining

fallenbagel:
## Description

The Radarr and Sonarr scanners take a snapshot of the library at the start of a run and, at the end, decline any in-progress request that isn't in it. Anything requested while the scan is still running was never in that snapshot, so it gets declined a few minutes after Radarr or Sonarr already accepted it. On the default schedule, since both scan jobs fall on a watchlist sync tick and nothing stops the two running at the same time either thus this happening.

Before declining anything, the scanners now check the candidate against the servers directly  instead of trusting the snapshot. If that check can't be completed the request is left alone, so a server being briefly unreachable can't decline someone's request.

## How Has This Been Tested?

- [X] Tested via unit tests
- [x] Tested by reporter (available to test via `preview-scanner-decline-race-fix`)

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Radarr and Sonarr library synchronization during scans.
  * Prevents media added after a scan begins from being incorrectly marked as missing.
  * Preserves processing status when servers are unreachable or verification fails.
  * Correctly resets status only when a confirmed library lookup finds no matching movie or series.

### DIFF
--- a/server/api/servarr/radarr.ts
+++ b/server/api/servarr/radarr.ts
@@ -81,6 +81,23 @@ class RadarrAPI extends ServarrBase<{ movieId: number }> {
     }
   };
 
+  public getLibraryMoviesByTmdbId = async (
+    tmdbId: number
+  ): Promise<RadarrMovie[]> => {
+    try {
+      const response = await this.axios.get<RadarrMovie[]>('/movie', {
+        params: { tmdbId },
+      });
+
+      return response.data;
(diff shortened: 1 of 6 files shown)


0xSysR3ll (approved):
Success

gauthier-th (approved):

PR #3402, opened 2026-08-15
feat: track episode availability from media servers scanners

0xSysR3ll:
## Description

Part 3 of #1671, stacked on #3355.

Wires episode availability into the media server scanners so media-server-only users get per-episode badges without Sonarr.
Episode numbers come from the library agent, so Seerr's metadata provider must match that agent (TMDB or TVDB).

Sonarr tracking stays TVDB-only.

## How Has This Been Tested?

Only unit tests for now.

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Added episode-level availability tracking across Plex, Jellyfin, and Sonarr libraries.
  - Added separate standard and 4K availability status for supported libraries.
  - Combined Jellyfin episode ranges are now tracked individually.

- **Bug Fixes**
  - Missing or fileless episodes are now removed or marked unavailable more accurately.
  - Prevented existing available episodes from being incorrectly downgraded during scans.

- **Documentation**
  - Clarified that Seerr’s metadata provider must match the library agent for episode availability tracking.


server/lib/availabilitySync.ts
```diff
@@ -1207,15 +1252,55 @@ class AvailabilitySync {
...
+              (mediaItem) => mediaItem.videoResolution === '4k'
             );
           } else if (this.enable4kShow) {
             seasonExistsInPlex = episodeVersions.some(
-              (mediaItem) => (mediaItem.width ?? 0) < 2000
+              (mediaItem) => mediaItem.videoResolution !== '4k'
             );
           } else {
             seasonExistsInPlex = episodeVersions.length > 0;
           }
```
fallenbagel, line 1361: This changes when a 4K season counts as present, independent of episode tracking. Same switch at line 1134-1138, while the movie path at 1097 and 1116 stays on width. This should be split out so it lands with the movie path.

server/lib/availabilitySync.ts
```diff
@@ -1352,12 +1445,113 @@ class AvailabilitySync {
...
+                  episodeHasFileByNumber,
+                  existingEpisodes
+                );
+              }
+            } else {
+              const episodes = await this.jellyfinClient.getEpisodes(
+                seriesId,
+                seasonMeta.Id,
+                { includeMediaInfo: true }
+              );
```
fallenbagel, line 1585: This runs for every 4K instance whether or not episode tracking is on. The branch splits on `enable4kShow`, not `shouldTrackEpisodes`, so the 4K pass stops treating a season as present just because episodes exist and starts requiring a stream over 2000 wide. That's the right behavior and it lines up with what the scanner already does, but it changes 4K season removal for anyone with a 4K Sonarr configured whether or not they turned tracking on. Gate it on `shouldTrackEpisodes` and let the existing behavior stand for everyone else.

server/lib/availabilitySync.ts
```diff
@@ -1342,7 +1435,7 @@ class AvailabilitySync {
       const seriesId = is4k ? ratingKey4k : ratingKey;
 
       if (seriesId) {
-        const cacheKey = `${seriesId}-${seasonMeta.Id}`;
+        const cacheKey = `${is4k ? '4k' : 'std'}-${seriesId}-${seasonMeta.Id}`;
```
fallenbagel: This is a good catch since the Plex cache already scopes by 4K and this one didn't. But imo, this should go to develop on its own rather than sit behind this stack because its a live bug in develop.

server/lib/scanners/plex/index.ts
```diff
@@ -340,25 +341,48 @@ class PlexScanner
...
+          const has4k =
+            this.enable4kShow &&
+            versions.some((media) => media.videoResolution === '4k');
+
+          if (hasStandard) {
+            totalStandard += 1;
+          }
+          if (has4k) {
+            total4k += 1;
+          }
```
fallenbagel, line 365: With 4K off, an episode Plex returns without a `Media` array which used to count toward `totalStandard` and now doesn't. Probably fine since it is typed non-pption but `hasStandard = this.enable4kShow ? versions.some(...) : true` keeps the count identical and the `?? []` still does its job in the 4K branch.

server/lib/scanners/plex/index.ts
```diff
@@ -340,25 +341,48 @@ class PlexScanner
...
-          ? episodes.filter((episode) =>
-              episode.Media.some((media) => media.videoResolution === '4k')
-            ).length
-          : 0;
+        let totalStandard = 0;
+        let total4k = 0;
+        const episodeDetails: ProcessableEpisode[] | undefined = settings.main
+          .enableEpisodeAvailability
+          ? []
+          : undefined;
```
fallenbagel, line 349: I think we need to make this very clear. Either in settings or settings that leads to docs or whatever, Plex agents are per-library, so an install with one TMDB-agent library and one TVDB-agent library can't satisfy the provider-matching requirement for both. Same with jellyfin

server/lib/scanners/baseScanner.ts
```diff
@@ -716,6 +717,10 @@ class BaseScanner<T> {
               MediaStatus.AVAILABLE;
             toSave.push(existingEpisode);
           }
+          if (this.enable4kShow && episodeDetail.hasFile4k) {
+            existingEpisode.status4k = MediaStatus.AVAILABLE;
+            toSave.push(existingEpisode);
+          }
```
fallenbagel, line 746: Same entity gets pushed twice when both flags are set.

server/lib/scanners/jellyfin/index.ts
```diff
@@ -400,6 +420,7 @@ class JellyfinScanner
               totalEpisodes: season.episode_count,
               episodes: totalStandard,
               episodes4k: total4k,
+              episodeDetails,
```
fallenbagel: The AniDB path sums counts across entries because each entry restarts at episode 1 and there's no offset to map with, `animelist.ts` only reads `defaulttvdbseason`. `episodeDetails` carries raw `IndexNumber` through anyway, so cour 2's episode 1 writes a row for episode 1 of the TMDB season and collides with cour 1 on the unique constraint. We would have to skip `episodeDetails` when `tvdbSeasonFromAnidb` is set.

server/lib/availabilitySync.ts
```diff
@@ -1207,15 +1252,55 @@ class AvailabilitySync {
...
+              }
+            }
+
+            await this.unmarkMissingEpisodes(
+              media,
+              season,
+              is4k,
+              episodeHasFileByNumber,
+              existingEpisodes
+            );
```
fallenbagel: Every source demotes on its own. `episodeHasFileByNumber` starts with every known row seeded false, so anything this pass doesn't see gets marked DELETED regardless of what the other sources know. The Sonarr pass clears an episode Plex has, the next Plex scan promotes it back, and the badge flips every cycle. Plex makes it worse. A combined S01E01-E02 file comes back as one item with index 1, since `PlexMetadata` has no `IndexNumberEnd`, so this pass deletes episode 2 that Sonarr correctly wrote as available. Seasons and media already resolve exists-in-Plex-or-Sonarr before demoting. Episodes need the same union.

fallenbagel (changes requested):
Two changes in here aren't about episode badges. The Jellyfin cache key fix should go straight to develop on its own as it's fixing a bug live today. The Plex 4K detection switch should either come out or land as its own PR covering the movie path too, since right now the file is half on `videoResolution` and half on `width` and there is no explanation as to why. The Jellyfin season-existence change is fine but keys off `enable4kShow` rather than the tracking setting, so it needs a gate or this will change behaviour for people who do not use this setting. The worst blocker is the per-source demotion since, as written, each source deletes what it can't see and the sources disagree.

0xSysR3ll:
> Two changes in here aren't about episode badges. The Jellyfin cache key fix should go straight to develop on its own as it's fixing a bug live today. The Plex 4K detection switch should either come out or land as its own PR covering the movie path too, since right now the file is half on `videoResolution` and half on `width` and there is no explanation as to why. The Jellyfin season-existence change is fine but keys off `enable4kShow` rather than the tracking setting, so it needs a gate or this will change behaviour for people who do not use this setting. The worst blocker is the per-source demotion since, as written, each source deletes what it can't see and the sources disagree.

Yep, since I already had the fixes I wanted to include them directly here but I'll revert and keep that for another PR then.

fallenbagel:
> > Two changes in here aren't about episode badges. The Jellyfin cache key fix should go straight to develop on its own as it's fixing a bug live today. The Plex 4K detection switch should either come out or land as its own PR covering the movie path too, since right now the file is half on `videoResolution` and half on `width` and there is no explanation as to why. The Jellyfin season-existence change is fine but keys off `enable4kShow` rather than the tracking setting, so it needs a gate or this will change behaviour for people who do not use this setting. The worst blocker is the per-source demotion since, as written, each source deletes what it can't see and the sources disagree.
> 
> Yep, since I already had the fixes I wanted to include them directly here but I'll revert and keep that for another PR then. 

We can merge those before this stack so they will get added here anyways. Like the cache one is a live bug already present in develop so

PR #3405, opened 2026-08-16
feat: add runtime base path support

### LINKED ISSUE #97: Need baseURL for reverse proxy support
### Description

We need a baseURL setting (like Radarr and Sonarr have).

### Desired Behavior

Allow Jellyseerr to run on `https://my.url.com/jellyseerr/`.

### Additional Context

_No response_

### Code of Conduct

DoubleThePsycho:
## Disclosure regarding AI work

<details>
<summary>AI disclosure</summary>

**Full disclosure:** I do not know TypeScript, or really JavaScript in general. I have tried learning it before, but there appears to be some fundamental compatibility issue between JavaScript and my brain, so we have mostly agreed to keep our distance.

I can usually follow the broader intent of code while reading it, but actually writing and debugging JS/TS myself turns into frustration pretty quickly.

Because of that, I am not going to pretend this was a normal contribution where I personally sat down and authored the implementation.

This feature was developed with substantial assistance from ChatGPT (GPT-5.6 Sol, High reasoning effort) over a long, iterative debugging and testing session.

ChatGPT inspected the Seerr codebase, proposed and modified the implementation, helped diagnose build/runtime failures, and later audited the result for additional places that were still assuming Seerr lived at `/`.

My part was mainly to:

- define the behavior I needed from an actual reverse-proxy deployment
- run the commands, builds, and tests
- inspect the results and report failures
- test each iteration against a real Seerr installation
- reproduce routing and hydration problems in the browser
- verify root-path behavior was not broken
- repeatedly rebuild and retest fixes as new issues were found

The implementation itself was therefore primarily produced with AI assistance rather than manually written by me.

This was not a one-shot generated patch. It went through a fairly ridiculous number of typecheck, lint, test, build, Docker, browser, reverse-proxy, routing, hydration, dynamic-route, and hard-coded-path debugging cycles before reaching the current version.

I also want to be clear that I understand the overall architecture and the problems we ran into, but I am not going to claim I could have written all 600+ lines of TypeScript from scratch myself.

This PR description has also been AI-assisted. I wrote the original version and then used ChatGPT to help trim it, correct factual mistakes, and make it less of a wall of text.

I am aware that the extent of AI involvement here may put this contribution outside the project's AI-assisted contribution policy. I would rather say that plainly than try to wordsmith around it.

I am submitting it because base-path support has been requested for years, the implementation is working on my actual homelab, and it has been tested far beyond just "it compiled on my machine."

If the maintainers decide it is better treated as a reference implementation rather than something mergeable as-is, that is fair.

</details>

## Description

Adds **runtime base path support** so Seerr can be hosted under a URL path such as `/seerr` without requiring the reverse proxy to strip the prefix, rewrite responses, or use a separately built image for each path.

The runtime base path can be configured using:

- `SEERR_BASE_PATH`
- Settings → Network → URL Base

`SEERR_BASE_PATH` takes precedence over the persisted setting when both are configured.

The same Docker image can therefore run at `/`, `/seerr`, or another supported path without being rebuilt.

The implementation makes the parts of Seerr that previously assumed `/` aware of the configured runtime path, including:

- Express/API/proxy routing
- browser API requests
- Next.js assets
- client-side links
- imperative `next/router` navigation
- dynamic routes
- public assets
- image/avatar proxies
- login/setup redirects
- Plex-related URLs
- service worker/PWA paths
- persisted Network settings
- server-side validation of base paths

Next.js' native `basePath` setting is not used because it is compiled into the frontend at build time.

Instead, Seerr sets its asset prefix at runtime and keeps Next.js' internal routes root-relative while exposing the configured prefix in browser-visible URLs.

The production Next.js build currently uses webpack instead of Turbopack for this because webpack allows the chunk public path to be changed at runtime. During testing, Turbopack could successfully serve JavaScript from the prefixed URL while still failing hydration because its chunk identity retained the build-time path.

The reverse proxy must preserve the prefix when forwarding requests to Seerr.

- Fixes #97

## How Has This Been Tested?

Testing was performed locally against the feature branch and the resulting image is also currently running as my actual Seerr instance.

I plan to keep using it there unless I run into something that requires switching back.

### Static validation

The following all passed:

- `pnpm typecheck`
- `pnpm format:check`
- `pnpm lint`
- `pnpm test`
- `pnpm build`
- `git diff --check`

Lint completed with:

- 0 errors
- 19 warnings from existing code

Tests completed with:

- 153 tests
- 153 passed
- 0 failed
- 0 skipped

The branch was rebased onto the latest `develop` immediately before submitting this PR and the full validation chain above was run again afterward.

### Translation validation

`pnpm i18n:extract` was run during development and the generated frontend locale changes were checked.

The translation keys for the URL Base setting are included in this PR.

### Docker / runtime testing

A generic production Docker image was built without compiling a base path into it.

The same image was tested at:

- `/`
- `/seerr`
- `/foobar`
- `/poop_aids_final_final`

Yes, the last test path was deliberately stupid. The point was to make sure this actually worked with an arbitrary path and not just `/seerr`.

No rebuild was required when changing between them.

Runtime testing included:

- initial page load and hydration
- prefixed Next.js JavaScript/assets
- API routing
- image/avatar proxy routing
- public assets
- service worker registration and scope
- web manifest
- offline behavior
- setup flow
- Plex configuration
- search navigation
- movie and TV detail navigation
- browser Back/Forward navigation
- dynamic Next.js routes
- persisted URL Base setting
- restarting Seerr after changing the setting
- `SEERR_BASE_PATH` override behavior

Client-side navigation was specifically checked using both normal links and imperative router navigation.

Dynamic movie/TV routes were also checked to make sure browser URLs contained the actual IDs rather than escaped route placeholders such as `[movieId]` or `[tvId]`.

### Server-side validation

The Network settings API was tested directly with an invalid base path:

```json
{
  "basePath": "/this is invalid"
}
```

The API returned HTTP 400:

```json
{
  "message": "Invalid URL base path."
}
```

The stored base path remained unchanged after the rejected request.

### Reverse proxy deployment

The feature image is currently running as my actual Seerr instance behind Caddy under:

```text
/seerr
```

The proxy preserves the prefix when forwarding requests to Seerr.

There is no response rewriting and the prefix is not stripped.

URL Base setting under Settings → Network

The visible UI addition is the **URL Base** setting under Settings → Network. Most of the rest of this change affects routing and deployment behavior rather than the UI.

## Checklist:

- [ ] I have read and followed the contribution [guidelines](https://github.com/seerr-team/seerr/blob/develop/CONTRIBUTING.md).

## Summary by CodeRabbit

* **New Features**
  * Added support for hosting the application under a configurable URL base path.
  * Added Network Settings controls to view, validate, and update the base path.
  * Updated navigation, API requests, images, PWA assets, authentication, and offline functionality to respect the configured path.
  * Added restart-required messaging when the base path changes.

* **Bug Fixes**
  * Improved asset and notification URL handling when running from a subdirectory.
  * Added validation and normalization for configured base paths.


fallenbagel:
Closing this. Per our [AI Assistance Notice](https://github.com/seerr-team/seerr/blob/4d17e08b91c8e1de41fd75a750c11b635d046434/CONTRIBUTING.md#ai-assistance-notice), this is AI-driven development, not AI-assisted, and disclosing that plainly does not change which bucket it falls into. And you said it yourself, you don't know TypeScript or JavaScript, you didn't write this, and you can't write the 600 lines you're asking us to review.

Running the build and clicking around your own homelab instance is not the same as reviewing the code. It means you verified behavior, not that you understood or can stand behind the implementation, which is what the policy actually requires.

The implementation itself reinforces that as well. This entire implementation isn't clean at all, and it's a stack of hacky workarounds that were just band-aided on because you couldn't get it to work.

For example:
- `next/link` and `next/router` are globally monkey-patched by aliasing them to your own modules through webpack and turbopack `resolveAlias`. That silently rewrites every import of Next's own routing primitives across the entire codebase, and not only just the parts touching base paths. This hacky workaround is not documented or a supported extension point, it relies on internal module resolution behavior, and it will break in ways that are painful to debug the moment Next changes anything about how those internals resolve.
- Production builds are switched from Turbopack to webpack for the sole reason that webpack lets you override the chunk public path at runtime. That's just hacking around a limitation by downgrading the build tool for the entire project, affecting everyone.
- `images.unoptimized: true` is set globally. That silently disables Next Image Optimization for every user of the project, rather than being limited to users with a base path. It's also not mentioned anywhere as an intentional tradeoff, and seems to be a side effect of the hacky workaround that was shipped with it. From your own admission, I don't think you can even defend a decision like this, which affects everyone.

**The entire implementation is a monkey-patched, hacky bandaid**. It's making global changes to work around limitations, and that's the kind of thing that can destabilize the codebase, introduce hard-to-debug breaking bugs, and become a maintenance nightmare down the line.

None of this is something a maintainer can reasonably sign off on, and by your own admission, it's not something you can defend if asked about it. This gets closed under the policy regardless of how much testing was done.

**If someone is implementing a feature or fix for the project, they need to understand and be able to own the code they're putting in front of us.** We can't reasonably review and maintain code that the author can't explain or defend themselves, regardless of how thoroughly they tested the resulting behavior. Relaying hundreds of debugging iterations from ChatGPT and calling the result “tested” doesn't change that. **It's disrespectful to the maintainers who are expected to spend their time reviewing, debugging, and maintaining code they didn't write and the author can't explain.**

PR #3428, opened 2026-08-24
fix(override-rules): match default *arr server by id

aussierk:
## Description

 Override rules are matched against a server's persistent id, but the default-server lookup used findIndex when determining if the request matched the rule. The use of findIndex led to Override rules comparing that id against the server's position in the settings array instead of its id. If a server had been deleted or if an instance had multiple servers configured, the array index and id would diverge so a matching override rule would silently never apply even when all the conditions were met.

This change makes the default-server lookup match by id instead of by findIndex to ensure that the match is done correctly and override rules are applied when the conditions are met. I added regression coverage in request.test.ts for both Radarr and Sonarr to cover the scenario where the findIndex array position and the persistent id do not match and a test where the findIndex array position and the persistent id do match. 

AI Disclosure: I diagnosed the bug myself and came up with the fix but had Claude help write the regression tests which I reviewed before completing testing. 

- Fixes issue reported in the discord by staky where override rule was not applying. 

## How Has This Been Tested?

 I ran the unit test and will have the original reporter test the fix. 

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected override-rule selection for default Radarr and Sonarr services, including configurations where service IDs differ from their list positions.
  * Prevented invalid override-rule lookups when no default media service is configured.

* **Tests**
  * Added coverage for movie and TV requests using matching and non-matching service IDs.
  * Added test scenarios for override rules and configurable media service settings.

### DIFF
--- a/server/entity/MediaRequest.ts
+++ b/server/entity/MediaRequest.ts
@@ -252,13 +252,21 @@ export class MediaRequest {
         ? settings.sonarr.find((s) => s.is4k && s.isDefault)?.id
         : settings.sonarr.find((s) => !s.is4k && s.isDefault)?.id;
 
+      const defaultServiceId =
+        requestBody.mediaType === MediaType.MOVIE
+          ? defaultRadarrId
+          : defaultSonarrId;
+
       const overrideRuleRepository = getRepository(OverrideRule);
-      const overrideRules = await overrideRuleRepository.find({
-        where:
-          requestBody.mediaType === MediaType.MOVIE
-            ? { radarrServiceId: defaultRadarrId }
-            : { sonarrServiceId: defaultSonarrId },
-      });
+      const overrideRules =
+        defaultServiceId === undefined
+          ? []
+          : await overrideRuleRepository.find({
+              where:
+                requestBody.mediaType === MediaType.MOVIE
+                  ? { radarrServiceId: defaultServiceId }
+                  : { sonarrServiceId: defaultServiceId },
+            });
 
       const appliedOverrideRules = overrideRules.filter((rule) => {
         const hasAnimeKeyword =
--- a/server/entity/MediaRequest.ts
+++ b/server/entity/MediaRequest.ts
@@ -246,11 +246,11 @@ export class MediaRequest {
 
     if (useOverrides) {
       const defaultRadarrId = requestBody.is4k
-        ? settings.radarr.findIndex((r) => r.is4k && r.isDefault)
-        : settings.radarr.findIndex((r) => !r.is4k && r.isDefault);
+        ? settings.radarr.find((r) => r.is4k && r.isDefault)?.id
+        : settings.radarr.find((r) => !r.is4k && r.isDefault)?.id;
       const defaultSonarrId = requestBody.is4k
-        ? settings.sonarr.findIndex((s) => s.is4k && s.isDefault)
-        : settings.sonarr.findIndex((s) => !s.is4k && s.isDefault);
+        ? settings.sonarr.find((s) => s.is4k && s.isDefault)?.id
+        : settings.sonarr.find((s) => !s.is4k && s.isDefault)?.id;
 
       const overrideRuleRepository = getRepository(OverrideRule);
       const overrideRules = await overrideRuleRepository.find({
--- a/server/entity/MediaRequest.ts
+++ b/server/entity/MediaRequest.ts
@@ -252,20 +252,17 @@ export class MediaRequest {
         ? settings.sonarr.find((s) => s.is4k && s.isDefault)?.id
         : settings.sonarr.find((s) => !s.is4k && s.isDefault)?.id;
 
-      const defaultServiceId =
+      const [defaultServiceIdField, defaultServiceId] =
         requestBody.mediaType === MediaType.MOVIE
-          ? defaultRadarrId
-          : defaultSonarrId;
+          ? (['radarrServiceId', defaultRadarrId] as const)
+          : (['sonarrServiceId', defaultSonarrId] as const);
 
       const overrideRuleRepository = getRepository(OverrideRule);
       const overrideRules =
         defaultServiceId === undefined
           ? []
           : await overrideRuleRepository.find({
-              where:
-                requestBody.mediaType === MediaType.MOVIE
-                  ? { radarrServiceId: defaultServiceId }
-                  : { sonarrServiceId: defaultServiceId },
+              where: { [defaultServiceIdField]: defaultServiceId },
             });
 
       const appliedOverrideRules = overrideRules.filter((rule) => {
--- a/server/routes/request.test.ts
+++ b/server/routes/request.test.ts
@@ -33,12 +33,6 @@ const sendNotificationMock = mock.method(
   async () => undefined
 ).mock;
 
-// getMovie/getTvShow are class-field arrow functions (own instance
-// properties), so they can't be mocked via a plain prototype assignment —
-// intercept them with a prototype accessor whose setter is a no-op, so the
-// constructor's `this.getMovie = ...` hits the no-op setter instead of
-// shadowing the prototype getter. (Pattern copied from
-// availabilitySync.test.ts.)
 const getMovieImpl: (args: {
   movieId: number;
   language?: string;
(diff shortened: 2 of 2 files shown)


server/routes/request.test.ts
```diff
@@ -26,6 +33,114 @@ const sendNotificationMock = mock.method(
   async () => undefined
 ).mock;
 
+// getMovie/getTvShow are class-field arrow functions (own instance
+// properties), so they can't be mocked via a plain prototype assignment —
+// intercept them with a prototype accessor whose setter is a no-op, so the
+// constructor's `this.getMovie = ...` hits the no-op setter instead of
+// shadowing the prototype getter. (Pattern copied from
+// availabilitySync.test.ts.)
```
fallenbagel: Remove this. This is just noise

server/routes/request.test.ts
```diff
@@ -26,6 +33,114 @@ const sendNotificationMock = mock.method(
...
+    original_language: 'en',
+    keywords: { results: [] },
+    external_ids: {},
+  } as unknown as TmdbTvDetails;
+}
+
+// Copied from server/lib/scanners/radarr/radarr.test.ts's configureRadarr —
+// `id` defaults to array index i unless overridden, so an explicit `id`
+// that differs from array position simulates a server that was deleted and
+// re-added (the scenario that exposes the override-rule matching bug).
```
fallenbagel: This is just noise.

server/routes/request.test.ts
```diff
@@ -26,6 +33,114 @@ const sendNotificationMock = mock.method(
...
+    isDefault: i === 0,
+    syncEnabled: true,
+    preventSearch: false,
+    externalUrl: '',
+    ...o,
+  })) as RadarrSettings[];
+}
+
+// Copied from server/lib/scanners/sonarr/sonarr.test.ts's configureSonarr,
+// same reasoning as configureRadarr above.
```
fallenbagel: Same noise.

server/routes/request.test.ts
```diff
@@ -535,3 +650,123 @@ describe('DELETE /request/:requestId, deleted media status restoration', () => {
     assert.strictEqual(updated.status, MediaStatus.PARTIALLY_AVAILABLE);
   });
 });
+
+describe('POST /request (movie), override rules', () => {
+  it('applies an override rule when the default Radarr server id differs from its array index', async () => {
+    // Simulates a Radarr server that was deleted and re-added: it sits at
+    // array index 0, but its persistent id is 5 (id != index).
```
fallenbagel: Noise. Literally the test scenario already explains this.

fallenbagel (changes requested):

server/routes/request.test.ts
aussierk: Okay, Will remove all the noise.

aussierk:
Removed the comments so that the test file is clean

server/entity/MediaRequest.ts
```diff
@@ -246,11 +246,11 @@ export class MediaRequest {
 
     if (useOverrides) {
       const defaultRadarrId = requestBody.is4k
-        ? settings.radarr.findIndex((r) => r.is4k && r.isDefault)
-        : settings.radarr.findIndex((r) => !r.is4k && r.isDefault);
+        ? settings.radarr.find((r) => r.is4k && r.isDefault)?.id
+        : settings.radarr.find((r) => !r.is4k && r.isDefault)?.id;
```
0xSysR3ll, line 250: A bit nit but I would have added this : 
```suggestion
        ? settings.radarr.find((r) => r.is4k && r.isDefault)?.id ?? -1
        : settings.radarr.find((r) => !r.is4k && r.isDefault)?.id ?? -1;
```
because if there’s no default server, `.find()?.id` is `undefined`. Which means you'd load every override rule instead of `none`.
`-1` is what `findIndex` used to return on a miss, so nothing matches.

@fallenbagel wdyt ?

server/entity/MediaRequest.ts
```diff
@@ -246,11 +246,11 @@ export class MediaRequest {
...
       const defaultSonarrId = requestBody.is4k
-        ? settings.sonarr.findIndex((s) => s.is4k && s.isDefault)
-        : settings.sonarr.findIndex((s) => !s.is4k && s.isDefault);
+        ? settings.sonarr.find((s) => s.is4k && s.isDefault)?.id
+        : settings.sonarr.find((s) => !s.is4k && s.isDefault)?.id;
```
0xSysR3ll, line 253:
```suggestion
        ? settings.sonarr.find((s) => s.is4k && s.isDefault)?.id ?? -1
        : settings.sonarr.find((s) => !s.is4k && s.isDefault)?.id ?? -1;
```
Same here.

server/entity/MediaRequest.ts
aussierk: I started looking into this when I saw the comments in code rabbit. An idea I had was to define defaultServiceId once for the relevant media type and then short-circuiting the override-rules when its undefined which avoids all the rules being applied if its undefined for some reason.
```suggestion
 const defaultServiceId =
  requestBody.mediaType === MediaType.MOVIE
    ? defaultRadarrId
    : defaultSonarrId;
    
const overrideRuleRepository = getRepository(OverrideRule);
const overrideRules =
  defaultServiceId === undefined
    ? []
    : await overrideRuleRepository.find({
        where:
          requestBody.mediaType === MediaType.MOVIE
            ? { radarrServiceId: defaultServiceId }
            : { sonarrServiceId: defaultServiceId },
      });
```

aussierk: This change makes sense, I was looking at this after the CodeRabbit review and had come up with a slight alternative that is a bit more code but more explicit in the bypass, avoiding the magic number fix but I am happy to apply either.

```suggestion
     const defaultServiceId =
        requestBody.mediaType === MediaType.MOVIE
          ? defaultRadarrId
          : defaultSonarrId;

      const overrideRuleRepository = getRepository(OverrideRule);
      const overrideRules =
        defaultServiceId === undefined
          ? []
          : await overrideRuleRepository.find({
              where:
                requestBody.mediaType === MediaType.MOVIE
                  ? { radarrServiceId: defaultServiceId }
                  : { sonarrServiceId: defaultServiceId },
            });
```

This suggestion spans both of your suggestions, removes lines 248 & 251 and modifies lines 256-261

0xSysR3ll: I guess it's slightly better like this yes.

aussierk: Okay, I just pushed that fix then.

server/entity/MediaRequest.ts
```diff
@@ -246,19 +246,27 @@ export class MediaRequest {
...
-      });
+      const overrideRules =
+        defaultServiceId === undefined
+          ? []
+          : await overrideRuleRepository.find({
+              where:
+                requestBody.mediaType === MediaType.MOVIE
+                  ? { radarrServiceId: defaultServiceId }
+                  : { sonarrServiceId: defaultServiceId },
+            });
```
fallenbagel: Nit: `requestBody.mediaType === MediaType.MOVIE` is checked twice here, once for `defaultServiceId` and again for the where shape. Would be cleaner to build the column/value pair once off `defaultServiceId` and skip the second ternary

server/routes/request.test.ts
```diff
@@ -535,3 +638,121 @@ describe('DELETE /request/:requestId, deleted media status restoration', () => {
...
+    const res = await agent.post('/request').send({
+      mediaType: MediaType.TV,
+      mediaId: 88004,
+      seasons: [1],
+    });
+
+    assert.strictEqual(res.status, 201);
+    assert.strictEqual(res.body.rootFolder, '/overridden/tv');
+  });
+});
```
fallenbagel, line 815: I think we should cover a default Radarr/Sonarr configured for the relevant tier case test scenario too. Add one case per media type where `isDefault` is never set (or the array is empty) and an unrelated override rule exists, asserting it does NOT apply

fallenbagel (changes requested):

fallenbagel (commented):
Just two nits. After which this can be approved

server/routes/request.test.ts
aussierk: Okay, will add those in a sec.

aussierk:
Nits should be fixed

server/routes/request.test.ts
aussierk: Should be good

fallenbagel (approved):

0xSysR3ll (approved):

Issue #1420, opened 2025-03-03
A page dedicated to collections


kavinask007:
### Description

Currently AFAIK there is no page to see collections . If I need to download the james bond collection for example I need to find a james bond move then check the collections.

### Desired Behavior

 It would be nice to have a page dedicated to popular collections  ability to search collections

### Code of Conduct

- [x] I agree to follow Jellyseerr's Code of Conduct

fallenbagel:
> ability to search collections

You can already search for collections using search but yes we dont yet have a collections dedicated page

Mansarde:
> > ability to search collections
> 
> You can already search for collections using search but yes we dont yet have a collections dedicated page

Is this still correct?
I'm using the latest release (`v3.2.0`), but when I search for any collection (e.g. "James Bond Collection") it returns no results.
And if I search through only for "James Bond" then the collection is not among the results either.

I've tried this with multiple collections and none appear.
Did maybe something change on the TMDB side that breaks collection search? 🤔

PR #3430, opened 2026-08-25
feat(search): restore collection results in main search

0xSysR3ll:
## Description

Searching for something like "Star Wars Collection" no longer returned movie collections since TMDB's `/search/multi` route stopped including them. See https://developer.themoviedb.org/reference/search-multi
We weren't calling `/search/collection` neither, so the only way collections showed up was through the collection banner on a movie page.

- Related to #1420

## How Has This Been Tested?

Searched for "Star Wars Collection".

image

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Search results now include collections alongside movies, TV shows, and people.
  * Collection results display titles, artwork, overviews, language, and adult-content status.
  * Collection searches support pagination, language preferences, and adult-content filtering.
  * Collections are combined with other search results for a unified experience.

* **Bug Fixes**
  * Improved search pagination and end-of-results detection for more reliable loading.
  * Prevented collection searches from requesting unavailable pages.

### DIFF
--- a/seerr-api.yml
+++ b/seerr-api.yml
@@ -829,6 +829,38 @@ components:
             oneOf:
               - $ref: '#/components/schemas/MovieResult'
               - $ref: '#/components/schemas/TvResult'
+    CollectionResult:
+      type: object
+      required:
+        - id
+        - mediaType
+        - title
+      properties:
+        id:
+          type: number
+          example: 10
+        mediaType:
+          type: string
+          default: 'collection'
+        title:
+          type: string
+          example: Star Wars Collection
+        originalTitle:
+          type: string
+          example: Star Wars Collection
+        adult:
+          type: boolean
+          example: false
+        posterPath:
+          type: string
+        backdropPath:
+          type: string
+        overview:
+          type: string
+          example: Overview of the collection
+        originalLanguage:
+          type: string
+          example: 'en'
     Genre:
       type: object
       properties:
@@ -5491,8 +5523,8 @@ paths:
                     type: number
   /search:
     get:
-      summary: Search for movies, TV shows, or people
-      description: Returns a list of movies, TV shows, or people a JSON object.
+      summary: Search for movies, TV shows, people, or collections
+      description: Returns a list of movies, TV shows, people, or collections as a JSON object.
       tags:
         - search
       parameters:
@@ -5537,6 +5569,7 @@ paths:
                         - $ref: '#/components/schemas/MovieResult'
                         - $ref: '#/components/schemas/TvResult'
                         - $ref: '#/components/schemas/PersonResult'
+                        - $ref: '#/components/schemas/CollectionResult'
   /search/keyword:
     get:
       summary: Search for keywords
--- a/server/api/themoviedb/index.ts
+++ b/server/api/themoviedb/index.ts
@@ -19,6 +19,7 @@ import type {
   TmdbPersonDetails,
   TmdbProductionCompany,
   TmdbRegion,
+  TmdbSearchCollectionResponse,
   TmdbSearchMovieResponse,
   TmdbSearchMultiResponse,
   TmdbSearchTvResponse,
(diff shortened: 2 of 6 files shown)


fallenbagel (commented):
Isn't this calling `tmdb.searchCollections()` on every page regardless of whether collections have already run out. Once `page > collections.total_pages` from an earlier response, that call will just return an empty result and it's an extra TMDB request for nothing on every subsequent page. Might be worth caching/skipping once we know collections are exhausted for a given query?

Wdyt @gauthier-th ?

0xSysR3ll:
> Isn't this calling `tmdb.searchCollections()` on every page regardless of whether collections have already run out. Once `page > collections.total_pages` from an earlier response, that call will just return an empty result and it's an extra TMDB request for nothing on every subsequent page. Might be worth caching/skipping once we know collections are exhausted for a given query?
> 
> Wdyt @gauthier-th ?

Yeah I see what you mean. How do you see this though ?

Only hit collections on `page === 1`? They’re almost always a single page anyway.
Or skip once we know they’re exhausted? And use a short-lived server cache keyed by query/language, or the client passing something back.

gauthier-th:
>  Or skip once we know they’re exhausted? And use a short-lived server cache keyed by query/language, or the client passing something back.

Maybe not the cleanest way to do this, but it could be done client-side (e.g. if client doesn't receive anymore collection, ask to stop fetching them)

fallenbagel:
Dug into this a bit. Since the tmdb cache key includes `page`, it doesn't dedupe here, and the frontend's infinite scroll keeps incrementing `page` on every scroll tick. So once collections run out, every subsequent scroll fires a real (empty) TMDB request.

I think the cleanest fix would be to cache `collections.total_pages` per `query+language` in the tmdb cache, and skip `searchCollections` once `page` exceeds it.

Maybe something like:
```ts
const tmdbCache = cacheManager.getCache('tmdb');
const pagesKey = `search-collections-pages:${language}:${queryString}`;
const knownTotalPages = tmdbCache.get<number>(pagesKey);

let collections: TmdbSearchCollectionResponse;
if (knownTotalPages !== undefined && page > knownTotalPages) {
  collections = { page, results: [], total_pages: knownTotalPages, total_results: 0 };
} else {
  collections = await tmdb.searchCollections({ query: queryString, page, language });
  tmdbCache.set(pagesKey, collections.total_pages, 300);
}
```
That would make sude there isnt a need for a client contract change and should be good even for queries with more fhan one oage of collection matches.

Cc: @gauthier-th

0xSysR3ll:
> Dug into this a bit. Since the tmdb cache key includes `page`, it doesn't dedupe here, and the frontend's infinite scroll keeps incrementing `page` on every scroll tick. So once collections run out, every subsequent scroll fires a real (empty) TMDB request.
> 
> I think the cleanest fix would be to cache `collections.total_pages` per `query+language` in the tmdb cache, and skip `searchCollections` once `page` exceeds it.
> 
> Maybe something like:
> 
> ```ts
> const tmdbCache = cacheManager.getCache('tmdb');
> const pagesKey = `search-collections-pages:${language}:${queryString}`;
> const knownTotalPages = tmdbCache.get<number>(pagesKey);
> 
> let collections: TmdbSearchCollectionResponse;
> if (knownTotalPages !== undefined && page > knownTotalPages) {
>   collections = { page, results: [], total_pages: knownTotalPages, total_results: 0 };
> } else {
>   collections = await tmdb.searchCollections({ query: queryString, page, language });
>   tmdbCache.set(pagesKey, collections.total_pages, 300);
> }
> ```
> 
> That would make sude there isnt a need for a client contract change and should be good even for queries with more fhan one oage of collection matches.
> 
> Cc: @gauthier-th

Yeah I think I'll implement something like this.

gauthier-th:
> Dug into this a bit. Since the tmdb cache key includes `page`, it doesn't dedupe here, and the frontend's infinite scroll keeps incrementing `page` on every scroll tick. So once collections run out, every subsequent scroll fires a real (empty) TMDB request.
> 
> I think the cleanest fix would be to cache `collections.total_pages` per `query+language` in the tmdb cache, and skip `searchCollections` once `page` exceeds it.
> 
> Maybe something like:
> 
> ```ts
> const tmdbCache = cacheManager.getCache('tmdb');
> const pagesKey = `search-collections-pages:${language}:${queryString}`;
> const knownTotalPages = tmdbCache.get<number>(pagesKey);
> 
> let collections: TmdbSearchCollectionResponse;
> if (knownTotalPages !== undefined && page > knownTotalPages) {
>   collections = { page, results: [], total_pages: knownTotalPages, total_results: 0 };
> } else {
>   collections = await tmdb.searchCollections({ query: queryString, page, language });
>   tmdbCache.set(pagesKey, collections.total_pages, 300);
> }
> ```
> 
> That would make sude there isnt a need for a client contract change and should be good even for queries with more fhan one oage of collection matches.
> 
> Cc: @gauthier-th

Yes, makes sense

PR #3444, opened 2026-08-30
fix(security): pin handlebars to 4.7.9 (CVE-2026-33937)

nateships:
## Description

`email-templates` → `@ladjs/consolidate` pulls in `handlebars@4.7.8`, which is vulnerable to **CVE-2026-33937**: remote code execution via a crafted AST object passed to `Handlebars.compile()`. The fix is `handlebars@4.7.9`.

This adds a pnpm override pinning `handlebars` to `4.7.9`, a semver-compatible patch within the existing `^4.7.6` range, and regenerates the lockfile with the pinned `pnpm@10.24.0` on Node 22.

Trivy flags this CVE as CRITICAL on the published container image, so downstream image scanners raise it on every deploy.

Note: pnpm 10.24 prints a deprecation warning that `pnpm.overrides` in `package.json` is no longer the preferred location (new home: `pnpm-workspace.yaml`). It still works today (this PR and the existing overrides rely on it), but the migration may be worth tracking.

## How Has This Been Tested?

- `pnpm install` with the pinned `pnpm@10.24.0` on Node 22.19: the lockfile regenerates with `handlebars@4.7.9` resolved and the override recorded in the lockfile `overrides:` block.
- `pnpm build` succeeds on this branch (Node 22.19, pnpm 10.24.0).
- Verified `handlebars` is consumed only via `email-templates`/`@ladjs/consolidate` with a `^4.7.6` range, so `4.7.9` is a compatible patch; no other resolution changes in the lockfile.

Trivy on `ghcr.io/seerr-team/seerr:v3.4.1`:

```
handlebars (package.json)  CVE-2026-33937  CRITICAL  fixed  4.7.8 → 4.7.9
```

## Checklist:

- [x] Disclosed any use of AI (see our [policy](https://github.com/seerr-team/seerr/blob/develop/CONTRIBUTING.md#ai-assistance-notice)): this PR was prepared with AI assistance (Claude); the change and verification steps were reviewed by the author.
- [ ] I have updated the documentation accordingly. (N/A: dependency pin only)
- [ ] All new and existing tests passed. (no code paths changed; deferring to CI)
- [ ] Translation keys `pnpm i18n:extract` (N/A: no UI strings changed)
- [ ] Database migration (if required) (N/A)

### DIFF
--- a/package.json
+++ b/package.json
@@ -214,7 +214,8 @@
       "sqlite3>node-gyp": "8.4.1",
       "@types/express-session": "1.18.2",
       "@types/react": "19.2.14",
-      "@types/react-dom": "19.2.3"
+      "@types/react-dom": "19.2.3",
+      "handlebars": "4.7.9"
     }
   },
   "scarfSettings": {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -9,6 +9,7 @@ overrides:
   '@types/express-session': 1.18.2
   '@types/react': 19.2.14
   '@types/react-dom': 19.2.3
+  handlebars: 4.7.9
 
 importers:
 
@@ -82,7 +83,7 @@ importers:
         version: 0.2.9
       email-templates:
         specifier: 13.0.1
-        version: 13.0.1(@babel/core@7.29.0)(handlebars@4.7.8)(mustache@4.2.0)(pug@3.0.4)(react-dom@19.2.6(react@19.2.6))(react@19.2.6)(underscore@1.13.7)
+        version: 13.0.1(@babel/core@7.29.0)(mustache@4.2.0)(pug@3.0.4)(react-dom@19.2.6(react@19.2.6))(react@19.2.6)(underscore@1.13.7)
       express:
         specifier: 5.2.1
         version: 5.2.1
@@ -1668,7 +1669,7 @@ packages:
       haml-coffee: ^1.14.1
       hamlet: ^0.3.3
       hamljs: ^0.6.2
-      handlebars: ^4.7.6
+      handlebars: 4.7.9
       hogan.js: ^3.0.2
       htmling: ^0.0.8
       jazz: ^0.0.18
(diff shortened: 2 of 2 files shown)


M0NsTeRRR:
1) This is not the proper way to report a security vulnerability.
2) Please read the Contributing and Security guidelines.
3) Just because Trivy reports it doesn’t mean we are affected.

> pnpm 10.24 prints a deprecation warning that pnpm.overrides in package.json is no longer the preferred location (new home: pnpm-workspace.yaml). It still works today (this PR and the existing overrides rely on it), but the migration may be worth tracking.

We are aware

fallenbagel:
This is not a Seerr vulnerability. `handlebars` is not a direct dependency. It is an unused optional peer dependency that `@ladjs/consolidate` (via "email-templates") resolves in the dependency tree, but never actually loads. All of our email templates are `.pug`, and `consolidate` selects the render engine based on the file extension. That means `Handlebars.compile()`, which is the actual RCE sink in CVE-2026-33937, is never reached. **There is simply no code path here that would allow this to be exploited.**

Because of that, labeling this as "fix(security)" is **inaccurate**, and I am not going to merge a security fix for a vulnerability that does not exist in the application.

This has come up a few times and I have already explained this same reasoning (if you searched you might be able to find a duplicate pr for exactly this). The answer is the same from me. **If a dependency is not reachable from our code, it is not a Seerr vulnerability. Full stop.**

If the actual goal here is to reduce Trivy noise on the published image, that's fine, but it should be handled as such. We should not be adding `pnpm.overrides` entries for packages that have no attack surface in Seerr just because a scanner happened to flag them. Every override in this repo should have a real reason for being pinned. Otherwise, `package.json` just becomes a dumping ground for every CVE reported against unused transitive dependencies, which makes the overrides that actually matter harder to understand.

Scanner noise should be handled at the scanner level, not by changing the dependency tree. Add a `.trivyignore` entry for `CVE-2026-33937` with a comment explaining why it does not apply here. That works with the Trivy Container Vulnerability Scan workflows to ignore the noise for vulnerabilities that are not affecting the application. 

That's the PR I would accept in place of this one. So I'll close this.

CC: @seerr-team/seerr-core 

EDIT: @M0NsTeRRR didn't realise you already closed while I was typing this 😄
But my point still stands. This is not an exploitable vulnerability present on seerr anyways.

PR #3461, opened 2026-08-31
fix(requests): mark request FAILED when request *arr server is  deleted

### LINKED ISSUE #3460: Approved requests can get permanently stuck
### Description

 When the subscriber that dispatches a request to Radarr/Sonarr can't figure out which server to use because: 
1. The request's stored `serverId` points at a server that was deleted OR
2. There is simply no default server for that 4K/non-4K tier

Then it logs a warning and returns allowing the request to stay `APPROVED` and then the media proceeds to `PROCESSING` but nothing is ever sent to either Sonarr/Radarr and is not recoverable because `POST /request/:id/retry` only accepts request that are marked `FAILED`. The only recourse is to delete the request and make a new one. 

### Version

3.4.1

### Steps to Reproduce

1. Make a request without auto-approve on. 
2. As admin delete the server that the request was intended for. 
3. Approve the request. 

### Screenshots

_No response_

### Logs

```shell

```

### Platform

desktop

### Database

SQLite (default)

### Device

PC

### Operating System

Windows 11

### Browser

Brave

### Additional Context

_No response_

### Search Existing Issues

### Code of Conduct

aussierk:
## Description

 When MediaRequestSubscriber cannot resolve the server for a request due to the server being deleted or no default for that category being set, it now sets the request to `FAILED` instead of `PROCESSING`. Previously, the request would be marked as `PROCESSING` and fail silently. Admins were not alerted and the request would become permanently stuck. 

 

- Fixes #3460 

**AI Disclosure**: Claude Code assisted in writing the tests which I reviewed and verified. 
## How Has This Been Tested?

I ran the existing and new unit tests. 

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Requests with an unavailable configured media server now transition to **Failed**.
  * Clear failure notifications are sent when specifically configured media servers cannot be found.
  * Requests without a default media server remain **Approved** instead of being incorrectly marked as failed.
  * Removed requests no longer leave affected seasons stuck in pending or processing states.
  * Updating a TV request no longer adds seasons already held by another request.


fallenbagel (commented):
This changes behaviour for a very specific workflow. An admin who deliberately leaves no default Radarr/Sonarr server, and gates the server picker behind `REQUEST_ADVANCED` so only advanced users choose a server explicitly, will now see every plain (non-advanced) request fail. Track normal user requests without sending to arr.

A non-advanced user's request never touches the server field, so `entity.serverId` stays `null`. With no default configured, `radarrSettings`/`sonarrSettings` resolves to `undefined`, the override branch is skipped since `serverId` is `null`, and we land in `failRequestForUnresolvedServer`. Before this PR that branch just logged and left the request APPROVED. Now it flips to FAILED and fires a `MEDIA_FAILED` notification.

While that's the intended fix for the "forgot to set a default" case, but it also catches setups where no default was ever meant to exist for that tier of user. There's also no clean way back right now: `/retry` needs `MANAGE_REQUESTS` and doesn't accept a `serverId`, and the serverId-edit route only works while `PENDING`.

We already compute `hasExplicitServer` here for the message wording. Could we use it to gate behavior too? Only mark FAILED when there was an explicit `serverId` that no longer resolves (the deleted-override case), and keep the old silent-skip for "no default, no override" so we don't regress the advanced-only-routing setup. That narrows the fix to the case in the PR title l/linked issue as well.

CC: @seerr-team/seerr-core wdyt?

fallenbagel:
Also maybe make the pr title shorter haha 😅

aussierk:
> This changes behaviour for a very specific workflow. An admin who deliberately leaves no default Radarr/Sonarr server, and gates the server picker behind `REQUEST_ADVANCED` so only advanced users choose a server explicitly, will now see every plain (non-advanced) request fail. Track normal user requests without sending to arr.
> 
> A non-advanced user's request never touches the server field, so `entity.serverId` stays `null`. With no default configured, `radarrSettings`/`sonarrSettings` resolves to `undefined`, the override branch is skipped since `serverId` is `null`, and we land in `failRequestForUnresolvedServer`. Before this PR that branch just logged and left the request APPROVED. Now it flips to FAILED and fires a `MEDIA_FAILED` notification.
> 
> While that's the intended fix for the "forgot to set a default" case, but it also catches setups where no default was ever meant to exist for that tier of user. There's also no clean way back right now: `/retry` needs `MANAGE_REQUESTS` and doesn't accept a `serverId`, and the serverId-edit route only works while `PENDING`.
> 

So the next PR was going to address this part where I was going to stop the 200 -> refail loop with a 409 error then make FAILED requests editable by lifting the `PENDING` - only guard to include `FAILED` so that the admin can pick a valid server then Retry work normally. 

> We already compute `hasExplicitServer` here for the message wording. Could we use it to gate behavior too? Only mark FAILED when there was an explicit `serverId` that no longer resolves (the deleted-override case), and keep the old silent-skip for "no default, no override" so we don't regress the advanced-only-routing setup. That narrows the fix to the case in the PR title l/linked issue as well.
> 
> CC: @seerr-team/seerr-core wdyt?

I think that this would be a happy medium so that the two setups remain functional.

aussierk:
@fallenbagel I have incorporated your suggestion into this fix to separate the two workflows. A stacked PR with the retry fix to allow editing the request is coming soon. 

Am I correct in my understanding that the empty-array branches of the guard are for this same advanced setup use-case?

PR #3469, opened 2026-09-02
feat: filter by original languages (#3469)

developerwill:
## Description

Prevents unwanted requests by allowing the owner to restrict discovery results to a specific original language.

The **Original Language** filter is disabled whenever **Discover Language** is set to a value other than an empty string, since the two filters should not be used together.

## How Has This Been Tested?

Tested locally in the development environment.

* Verified that the **Original Language** filter is available on discovery pages.
* Verified that the filter can be used when **Discover Language** is empty.
* Verified that the **Original Language** filter is disabled when **Discover Language** has a value selected.
* Verified that the discovery results respect the selected original language.
* Ran the existing test suite and confirmed that all tests pass.
* Verified that the project builds successfully with `pnpm build`.

The **Original Language** filter field on discovery pages is disabled if **Discover Language** has any other value than an empty string.

image

image
I haven't used AI for any of my work.

## Checklist:

## New Features

- Added configurable original-language filtering across search, discovery, trending, upcoming, recommendations, similar titles, and media sliders.
- Added per-section language-filter toggles in Settings.
- Results are automatically filtered by supported original languages for applicable movie and TV content.
- Added server settings for configuring filter behavior and supported original languages.
- Media sliders now apply language, availability, and blocklist filters consistently.

## Updates

- Language selection is disabled when a server-level original language is configured.


fallenbagel (changes requested):
Please use the discover language multi select instead of the toggles.

developerwill:
> Please use the discover language multi select instead of the toggles.

Thanks, going to look at that. I get it now.

fallenbagel:
Also I'm not sure I understand this PR. Discover Language is a discover setting. It already sets `with_original_language` on `/discover/movie` and `/discover/tv`, and that's the whole contract. Applying it to search, recommendations, similar, trending and sliders changes what the setting means for every admin who already has it configured. Someone who set it to get Japanese rows on discover for example did not ask for search to stop returning results.

If the goal is restricting what users can request, that is a separate feature with its own setting, and shouldn't be a reinterpretation of this one.

Two other things:
- The filtering happens client side after fetch, so a page of 20 can render as 3 and pagination and counts stop lining up. A filter that isn't enforced at the request layer will always do this.
- The premise that Discover Language and the slideover's Original Language conflict isn't right. The slideover value already overrides the server default:

https://github.com/seerr-team/seerr/blob/970bb545716505e3d3d2fa7072a7aabbc5712c05/server/api/themoviedb/index.ts#L725-L730

That override is deliberate. **Disabling the selector removes it.**

developerwill:
> Also I'm not sure I understand this PR. Discover Language is a discover setting. It already sets `with_original_language` on `/discover/movie` and `/discover/tv`, and that's the whole contract. Applying it to search, recommendations, similar, trending and sliders changes what the setting means for every admin who already has it configured. Someone who set it to get Japanese rows on discover for example did not ask for search to stop returning results.
> 
> If the goal is restricting what users can request, that is a separate feature with its own setting, and shouldn't be a reinterpretation of this one.
> 
> Two other things:
> 
> * The filtering happens client side after fetch, so a page of 20 can render as 3 and pagination and counts stop lining up. A filter that isn't enforced at the request layer will always do this.
> * The premise that Discover Language and the slideover's Original Language conflict isn't right. The slideover value already overrides the server default:
> 
> https://github.com/seerr-team/seerr/blob/970bb545716505e3d3d2fa7072a7aabbc5712c05/server/api/themoviedb/index.ts#L725-L730
> 
> That override is deliberate. **Disabling the selector removes it.**

My ideia is to take advantage of the value already set for Discover Language since this option is to filter  content by original language. The issue with this option is that users can still search for content for all languages no matter what this input states.

For example, if I want my users to only be able to find content with Original Language [English, Japanese, Portuguese] currently there's no way to achieve that.

My PR gives the ability to filter out results based on the Discover Language input value with the flexibility of choosing where this filter can be applied (preferably everywhere).

I'll look into the pagination issue, that would intruduce a bug for sure.

developerwill (commented):
File changes review for the Filter by original language feature

<sub>Commented in [CodeRabbit Change Stack](https://app.coderabbit.ai/change-stack/seerr-team/seerr/pull/3469)</sub>

fallenbagel (changes requested):
I didn't do a thorough review but here are somethings I noticed while I skimmed through:

The MediaSlider fix looks right, but it’s the only place that actually got fixed. Every other scroll-driven view still takes `isEmpty` and `isReachingEnd` directly from `useDiscover`, which calculates them from the unfiltered titles:

https://github.com/seerr-team/seerr/blob/1efaf70e0588333a1a5ba45f92fc18660cfa7aae/src/hooks/useDiscover.ts#L145-L150

If page 1 returns 20 results and the filter removes all of them, `isEmpty` is still false, so the list renders nothing. `fetchMore` is also only triggered by `onScrollBottom`, but there’s nothing to scroll, so it never fires. It just gets stuck instead of fetching the next page.

That affects both Discover pages, both keyword pages, Upcoming, Trending, DiscoverTvUpcoming, and Search. Search also still uses `titles?.length` in its `isLoading` check.

MediaSlider itself still has the `size < 5` cap too, so a narrow allowlist could leave a slider with three items and then stop fetching.

The slideover issue I mentioned earlier is also still unaddressed:

https://github.com/seerr-team/seerr/blob/1efaf70e0588333a1a5ba45f92fc18660cfa7aae/src/components/Discover/FilterSlideover/index.tsx#L208

If Discover Language becomes an allowlist of [English, Japanese, Portuguese], the selector should be limited to those three rather than disabled. As it is now, a user who is allowed all three can’t narrow Discover down to just Japanese. So the restriction ends up removing the filtering it was supposed to preserve.

As for the overall goal, I still think this should be its own setting. The toggles can remain their own setting, but **the language allowlist shouldn’t be tied to Discover Language.**

Right now the hook just reads `originalLanguage` from public settings, so the allowlist effectively becomes Discover Language. That means you can’t have Japanese-only Discover rows while still allowing English, Japanese, and Portuguese elsewhere, such as Search.

That’s also why you had to disable the selector, since one value is effectively being used for two different purposes, so whatever you set for Discover gets interpreted as a restriction too. **That also removes functionality for existing users who don’t care about this feature.**

I’d give the allowlist its own languages field. That solves both issues, and you wouldn’t need the 13 flags either.

_As it stands, this can’t be merged regardless of the rest. Anyone who upgrades with Discover Language already set will lose the language filter in the slideover without opting into any of these changes._ That effectively removes existing functionality in the name of adding a new feature (in other words I consider this a regression).

Issue #800, opened 2024-06-07
Set partially requested Series to 'Available' when download completes


POD319:
### Description

When you request a partial series, ie -- only Season 1 of a 4 Season show -- when the download completes, it does not change from 'Requested'. This makes it hard to know when your request is actually complete. The Series will never set to 'Available' unless the entire Series is downloaded.

### Desired Behavior

Jellyseerr should mark partial series requests as 'available' when the request has downloaded.
Requested should only be shown when a series is not yet downloaded as according to the request, and not the entirety of the series.

### Code of Conduct

- [X] I agree to follow Jellyseerr's Code of Conduct

Fallenbagel:
It should add a badge as partially available. If it's not changing from requested, then it's an improperly configured jellyseerr, and I have explained this on discord as well.

![Screenshot_20240607_113137_Chrome Dev.jpg](https://github.com/Fallenbagel/jellyseerr/assets/98979876/8947ed5e-46fb-4a48-ad5b-26878c4781be)

Issue #710, opened 2024-04-09
Jellyseerr forces full series search instead of just the selected season


Atakku:
### Description

When I try to request one more season to a partially available series jellyseerr seems to request sonarr to perform a full series search instead of just the new season(s) i requested, which can take quite a while with large shows, and lead to issues

### Version

1.7.0 (docker.io/drcarrot/jellyseerr-postgres)

### Steps to Reproduce

1. Request another season on partially available series
2. Watch as sonarr sends requests to indexers looking for S1, S2, S3, and finally S4

### Platform

desktop

### Device

PC

### Operating System

Windows 11

### Browser

Firefox

### Code of Conduct

- [X] I agree to follow Jellyseerr's Code of Conduct

Fallenbagel:
Could you hop onto discord. This is a common configuration issue and we do troubleshooting there.

gauthier-th:
Can this be closed?

PCJones:
> Can this be closed?

No. It's still a problem

Fallenbagel:
> > Can this be closed?
> 
> No. It's still a problem 

Can you detail how to reproduce this then? Because none of us can.

PCJones:
> > > Can this be closed?
> > 
> > 
> > No. It's still a problem
> 
> Can you detail how to reproduce this then? Because none of us can.

The issue has been explained a bit more here:
https://github.com/Fallenbagel/jellyseerr/pull/711

You might need to set "Upgrade Until Custom Format Score" in your Sonarr quality profile to a really large number so that sonarr will (in theory) always look for upgrades and is never satisfied (as is recommended by TRaSH Guides). If you still can't reproduce it that way I will try to create a step by step guide with a clean sonarr instance to see what exactly are needed to reproduce this

Fallenbagel:
> You might need to set "Upgrade Until Custom Format Score" in your Sonarr quality profile to a really large number so that sonarr will (in theory) always look for upgrades and is never satisfied (as is recommended by TRaSH Guides).

I already have that enabled. I just requested just season 2 of a series and i got just season 2 only

PCJones:
Alright, I'll look into a step by step guide. It'll take a while as I'm currently on vacation

gauthier-th:
I can reproduce this

PR #3487, opened 2026-09-08
fix(requests): scope request availability badge to requested seasons

SanJJ1:
## What this changes

A request for only some seasons of a multi-season show currently shows **Partially Available** and never flips to **Available**, even after every season you actually requested is available. The badge reads the show's overall status, not the seasons the request was for.

This scopes the request's availability badge to the **specific seasons that were requested**:

- Request S8 only, S8 available → **Available**
- Request S1–S3, some available → **Partially Available**
- Request S1–S3, none available yet → falls back to the current behavior

Movie requests and the show-level media badge are untouched.

## Where

- `src/utils/requestStatus.ts` — new `getRequestedSeasonsStatus` helper: rolls up the requested seasons' own statuses (`media.seasons[].status` / `status4k`).
- `src/components/RequestList/RequestItem/index.tsx` and `src/components/RequestCard/index.tsx` — the two request views now pass that helper's result to `StatusBadge` instead of `media.status`.

## Why it needs no backend change

`MediaRequest.media` and `Media.seasons` are both `eager`, so `/api/v1/request/:id` (which every request row fetches) already returns `media.seasons[].status`. The helper falls back to the media status when seasons aren't present (e.g. the request-list payload before the per-row fetch resolves), so there's no regression during that window.

## Scope

Only the badge shown on a request is changed. The show-level "Available/Partially Available" badge and the request-list server-side status filter are intentionally left alone.

## Related

- seerr #800, #710
- Original report: sct/overseerr #3796

### DIFF
--- a/src/components/RequestCard/index.tsx
+++ b/src/components/RequestCard/index.tsx
@@ -14,6 +14,7 @@ import {
   getRequestDownloadStatus,
   refreshIntervalHelper,
 } from '@app/utils/refreshIntervalHelper';
+import getRequestedSeasonsStatus from '@app/utils/requestStatus';
 import { withProperties } from '@app/utils/typeHelpers';
 import {
   ArrowPathIcon,
@@ -156,11 +157,7 @@ const RequestCardError = ({ requestData }: RequestCardErrorProps) => {
                     </Badge>
                   ) : (
                     <StatusBadge
-                      status={
-                        requestData.media[
-                          requestData.is4k ? 'status4k' : 'status'
-                        ]
-                      }
+                      status={getRequestedSeasonsStatus(requestData)}
                       downloadItem={requestDownloadStatus}
                       title={intl.formatMessage(messages.unknowntitle)}
                       inProgress={requestDownloadStatus.length > 0}
@@ -459,9 +456,7 @@ const RequestCard = ({ request, onTitleData }: RequestCardProps) => {
               </Badge>
             ) : (
               <StatusBadge
-                status={
-                  requestData.media[requestData.is4k ? 'status4k' : 'status']
-                }
+                status={getRequestedSeasonsStatus(requestData)}
                 downloadItem={requestDownloadStatus}
                 title={isMovie(title) ? title.title : title.name}
                 inProgress={requestDownloadStatus.length > 0}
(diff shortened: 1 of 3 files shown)


fallenbagel:
Closing this.

The required template has been bypassed. Our own structure was replaced with a generated one and the checklist is gone, so neither the guidelines nor the AI disclosure box was checked. That alone is grounds to close: https://github.com/seerr-team/seerr/blob/develop/CONTRIBUTING.md#ai-assistance-notice

The premise is also wrong. The status field on a request row is the media availability badge, not the request's status. It reads `media.status` / `media.status4k` and it's what drives Play on Plex, Open in Sonarr and the download progress tooltip. DECLINED and FAILED are the only request statuses that take that slot, because neither has an availability equivalent.

https://github.com/seerr-team/seerr/blob/6f5a17735d383b110cca04326ecd536ad7675ed6/src/components/RequestList/RequestItem/index.tsx#L148-L153

Partial season requests already complete correctly. MediaSubscriber rolls up only the seasons the request asked for and marks the request COMPLETED once each of them is available or deleted.

https://github.com/seerr-team/seerr/blob/6f5a17735d383b110cca04326ecd536ad7675ed6/server/subscriber/MediaSubscriber.ts#L123-L129

image

So a single season request on a multi season series is already COMPLETED while the series itself is correctly Partially Available.. Nothing is stuck. What this patch does is make the same media read Available on the requests page and Partially Available on its detail page and title cards.

"Why it needs no backend change" is wrong too. `filter=available` on the request list is request COMPLETED AND `media.status = AVAILABLE`, so a row would render Available and then be excluded from the Available filter.

The helper has its own issues on top of that. DELETED counts as fulfilled server side but not in the `every()` check. Requested seasons with no matching row in `media.seasons` get filtered out of the rollup instead of counting as unavailable, so a partial payload yields a false Available. And `Math.min` over the status values assumes the enum ordinals are a severity scale, when BLOCKLISTED and DELETED both sit above AVAILABLE.

The bigger issue is that the PR confidently describes behavior that the code simply does not have. Combined with the generated template, removed checklist, and unchecked AI disclosure, this is not something we can accept as a valid contribution under our guidelines.

There is a real gap though. The requests page and the discover slider never surface COMPLETED at all, unlike the manage slideover. That's a separate change.

**Treat this as a warning.** We have closed other contributions on the same grounds and will keep applying it consistently. Further PRs that skip the template or contain undisclosed AI work may be closed without review.

PR #3501, opened 2026-09-11
fix(server): respond instead of hanging on proxy route errors

### LINKED ISSUE #3486: /avatarproxy hangs forever instead of responding when the user ID is invalid
### Description

`GET /avatarproxy/:jellyfinUserId` never sends a response when the path parameter
fails the 32-hex validation check. The request hangs until the client or an
upstream proxy times out.

In `server/routes/avatarproxy.ts`, the handler throws when the parameter does not
match `/^[a-f0-9]{32}$/`, and the `catch` block logs the error but never calls
`res`:

```js
} catch (e) {
  logger.error('Failed to proxy avatar image', {
    errorMessage: e.message,
  });
}
```

Since nothing is sent and `next(e)` is not called, Express keeps the request open
indefinitely. The route is mounted outside `/api/v1` (`server/index.ts`), so it is
not covered by the OpenAPI validator and requires no authentication — an
unauthenticated request can hold a server connection open with no rate limit.
Behind a reverse proxy this surfaces as a gateway timeout (Cloudflare returns 524
after 100s).

Still present on `develop` as of this report.

### Suggested fix

Send a response on the error path, e.g.:

```js
} catch (e) {
  logger.error('Failed to proxy avatar image', {
    errorMessage: e.message,
  });
  if (!res.headersSent) {
    res.status(400).send('Invalid avatar request');
  }
}
```

Returning 404 instead of 400 would be equally reasonable. Validating the parameter
before entering the `try` block would also work.

### Version

3.4.1

### Steps to Reproduce

1. Run any Seerr instance.
2. `curl -m 15 -o /dev/null -w '%{http_code} %{time_total}\n' http://<host>:5055/avatarproxy/1`
3. Observe no response — curl reports `000` at the timeout instead of a status code.
4. For contrast, `curl http://<host>:5055/avatarproxy/00000000000000000000000000000000`
   returns `200` with an image in well under a second.

Any value that is not exactly 32 lowercase hex characters reproduces it
(`/avatarproxy/abc`, `/avatarproxy/1`, etc.).

Note: this reproduces regardless of media server type. It was found on a
Plex-backed instance, where `/avatarproxy` is never used by the frontend at all —
those URLs are only generated for Jellyfin/Emby users — so the endpoint is
reachable and hangs even on installs that have no use for it.

### Screenshots

_No response_

### Logs

```shell
2026-09-08T01:32:32.635Z [error][API]: Failed to proxy avatar image {"errorMessage":"Provided URL is not an Emby avatar."}

$ curl -m 15 -o /dev/null -w 'HTTP %{http_code} after %{time_total}s\n' http://127.0.0.1:5055/avatarproxy/1
HTTP 000 after 15.001360s

$ curl -m 15 -o /dev/null -w 'HTTP %{http_code} after %{time_total}s\n' http://127.0.0.1:5055/avatarproxy/00000000000000000000000000000000
HTTP 200 after 0.009405s
```

### Platform

desktop

### Database

SQLite (default)

### Device

Desktop PC

### Operating System

Windows 11

### Browser

Chrome 152

### Additional Context

_No response_

### Search Existing Issues

### Code of Conduct

aussierk:
## Description

Prevents an infinite hang when the :jellyfinUserid fails the 32-hex validation check by providing an error response via next().  Also aligns `imageproxy.ts` to using the global error handler instead of its unique error handling.

**AI Disclosure**: no AI was used as part of this fix

- Fixes #3486 

## How Has This Been Tested?

- Ran all existing tests
- Replicated the reporting user's example calls and received a response

`$ curl -m 15 -o /dev/null -s -w 'HTTP %{http_code} after %{time_total}s\n' http://127.0.0.1:5055/avatarproxy/1
HTTP 400 after 0.008540s`

## Checklist:

## Summary by CodeRabbit

- **Bug Fixes**
  - Invalid avatar URLs now return clear 400 error responses.
  - Invalid or unsupported image proxy requests now return structured error responses.
  - Unexpected avatar and image proxy failures now return appropriate 500 responses when possible.
  - Existing error responses are preserved when a response has already started.

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -267,8 +267,11 @@ app
         res: Response,
         // We must provide a next function for the function signature here even though its not used
         // eslint-disable-next-line @typescript-eslint/no-unused-vars
-        _next: NextFunction
+        next: NextFunction
       ) => {
+        if (res.headersSent) {
+          return next(err);
+        }
         // format error
         res.status(err.status || 500).json({
           message: err.message,
--- a/server/index.ts
+++ b/server/index.ts
@@ -265,8 +265,6 @@ app
         err: { status: number; message: string; errors: string[] },
         _req: Request,
         res: Response,
-        // We must provide a next function for the function signature here even though its not used
-        // eslint-disable-next-line @typescript-eslint/no-unused-vars
         next: NextFunction
       ) => {
         if (res.headersSent) {
--- a/server/routes/avatarproxy.ts
+++ b/server/routes/avatarproxy.ts
@@ -115,19 +115,17 @@ export async function checkAvatarChanged(
   }
 }
 
-router.get('/:jellyfinUserId', async (req, res) => {
+router.get('/:jellyfinUserId', async (req, res, next) => {
+  if (!req.params.jellyfinUserId.match(/^[a-f0-9]{32}$/)) {
+    const mediaServerType = getSettings().main.mediaServerType;
+    return next({
+      status: 400,
+      message: `Provided URL is not ${
+        mediaServerType === MediaServerType.JELLYFIN ? 'a Jellyfin' : 'an Emby'
+      } avatar.`,
+    });
+  }
   try {
-    if (!req.params.jellyfinUserId.match(/^[a-f0-9]{32}$/)) {
-      const mediaServerType = getSettings().main.mediaServerType;
-      throw new Error(
-        `Provided URL is not ${
-          mediaServerType === MediaServerType.JELLYFIN
-            ? 'a Jellyfin'
-            : 'an Emby'
-        } avatar.`
-      );
-    }
-
     const avatarImageCache = await initAvatarImageProxy();
 
     const userEtag = req.headers['if-none-match'];
@@ -173,9 +171,14 @@ router.get('/:jellyfinUserId', async (req, res) => {
 
     res.end(imageData.imageBuffer);
   } catch (e) {
-    logger.error('Failed to proxy avatar image', {
-      errorMessage: e.message,
-    });
+    logger.error('Failed to proxy avatar image', { errorMessage: e.message });
+    if (!res.headersSent) {
+      return next({
+        status: 500,
+        message: 'Failed to proxy avatar image.',
+      });
+    }
+    next(e);
   }
 });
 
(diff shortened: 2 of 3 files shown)


server/index.ts
```diff
@@ -267,8 +267,11 @@ app
         res: Response,
         // We must provide a next function for the function signature here even though its not used
         // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
fallenbagel: `next` is used now, so both of these have to go. The comment says the opposite of what the code does, and `eslint.config.mts` has `reportUnusedDisableDirectives: true`, so the disable directive itself will start warning.

server/routes/avatarproxy.ts
```diff
@@ -173,9 +171,14 @@ router.get('/:jellyfinUserId', async (req, res) => {
...
-    });
+    logger.error('Failed to proxy avatar image', { errorMessage: e.message });
+    if (!res.headersSent) {
+      return next({
+        status: 500,
+        message: 'Failed to proxy avatar image.',
+      });
+    }
+    next(e);
   }
```
fallenbagel, line 182: `imageproxy.ts` handles the same cases inline with `res.status(403).send(...)` and `res.status(500).send()`, so the two proxy routes now answer differently for the same kind of failure, one plain text and one JSON through the global handler. Make them match

fallenbagel (changes requested):
Also retitle. We squash, so the PR title becomes the commit message, and this one is long enough that GitHub already truncates it. It also scopes everything to `avatarproxy` when half the change is the global error handler. Something like `fix(server): respond instead of hanging on invalid avatarproxy user ID` maybe? Or shorter?

server/routes/avatarproxy.ts
aussierk: Okay I will make them match

server/index.ts
aussierk: Okay

server/routes/avatarproxy.ts
aussierk: I am planning to move imageproxy tk the global handler based on gauthier's comments on using nsxt in the related issue.

aussierk:
> Also retitle. We squash, so the PR title becomes the commit message, and this one is long enough that GitHub already truncates it. It also scopes everything to `avatarproxy` when half the change is the global error handler. Something like `fix(server): respond instead of hanging on invalid avatarproxy user ID` maybe? Or shorter?

All changes made.

fallenbagel (approved):

0xSysR3ll (approved):

PR #3502, opened 2026-09-11
fix(jellyfin-api): update Authorization headers for Jellyfin

aussierk:
## Description

 As of Jellyfin 12.0 rejects deprecated Authorization Header `X-Emby-Authorization` when `EnableLegacyAuthorization` is `disabled` which is the new default . Emby still expects that header format so I branched by mediaServerType instead of hardcoding one header for both.

**AI Disclosure**: No AI was used in developing this fix

## How Has This Been Tested?

I ran the existing tests and then also ran it a local deployment of this branch to ensure that the api worked with Jellyfin 12.0.

## Checklist:

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved authentication compatibility for Jellyfin and Emby sign-in and sign-out requests.
  - Fixed avatar image retrieval to use the appropriate authorization format for the configured media server.

### DIFF
--- a/server/routes/auth.ts
+++ b/server/routes/auth.ts
@@ -897,7 +897,7 @@ authRoutes.post('/logout', async (req, res, next) => {
             await axios.delete(`${baseUrl}/Devices`, {
               params: { Id: user.jellyfinDeviceId },
               headers: {
-                'X-Emby-Authorization': `MediaBrowser Client="Seerr", Device="Seerr", DeviceId="seerr", Version="${
+                Authorization: `MediaBrowser Client="Seerr", Device="Seerr", DeviceId="seerr", Version="${
                   settings.main.mediaServerType === MediaServerType.EMBY
                     ? '1.0.0'
                     : getAppVersion()


fallenbagel (changes requested):
**Most of this isn't needed.** The API client moved to the modern `Authorization` header for both Jellyfin and Emby in #2211, and that constructor covers effectively all server to media-server traffic including `/Users/AuthenticateByName`:

https://github.com/seerr-team/seerr/blob/4fc265b60b67796b7b688571bf30f0a513b8bcb6/server/api/jellyfin.ts#L157-L173

So Jellyfin 12 already works. Emby also already works. What this PR does there is put Emby back on `X-Emby-Authorization`, which is **a change to a working path**, and you only tested Jellyfin. Drop it. Don't change behaviour that already works, keep it on `Authorization` for both.

Only two spots still hardcode the legacy header, and the avatar one makes literally no difference. Jellyfin serves `/UserImage` anonymously, so avatars load regardless of what we send. No change needed there either.

https://github.com/seerr-team/seerr/blob/4fc265b60b67796b7b688571bf30f0a513b8bcb6/server/routes/avatarproxy.ts#L28-L35

The logout one is the actual bug. `DELETE /Devices` needs auth, so on 12.0 it fails and the surrounding catch swallows it, which means device cleanup quietly stops happening.

https://github.com/seerr-team/seerr/blob/4fc265b60b67796b7b688571bf30f0a513b8bcb6/server/routes/auth.ts#L896-L905

**Also `src/utils/jellyfin.ts` is dead code.** (we dont use that file anymore). `JellyAPI` has no importers anywhere in the tree, and the `isEmby` param you added has no call site so it would always default to `false`. **That file should be deleted in a separate PR, not patched here.**

**So the fix is one line in `auth.ts`**: switch that header to `Authorization`. No mediaServerType branching, since Emby accepts the `MediaBrowser` scheme on `Authorization` too.

aussierk:
> **Most of this isn't needed.** The API client moved to the modern `Authorization` header for both Jellyfin and Emby in #2211, and that constructor covers effectively all server to media-server traffic including `/Users/AuthenticateByName`:
> 
> https://github.com/seerr-team/seerr/blob/4fc265b60b67796b7b688571bf30f0a513b8bcb6/server/api/jellyfin.ts#L157-L173
> 
> So Jellyfin 12 already works. Emby also already works. What this PR does there is put Emby back on `X-Emby-Authorization`, which is **a change to a working path**, and you only tested Jellyfin. Drop it. Don't change behaviour that already works, keep it on `Authorization` for both.
> 
> Only two spots still hardcode the legacy header, and the avatar one makes literally no difference. Jellyfin serves `/UserImage` anonymously, so avatars load regardless of what we send. No change needed there either.
> 
> https://github.com/seerr-team/seerr/blob/4fc265b60b67796b7b688571bf30f0a513b8bcb6/server/routes/avatarproxy.ts#L28-L35
> 
> The logout one is the actual bug. `DELETE /Devices` needs auth, so on 12.0 it fails and the surrounding catch swallows it, which means device cleanup quietly stops happening.
> 
> https://github.com/seerr-team/seerr/blob/4fc265b60b67796b7b688571bf30f0a513b8bcb6/server/routes/auth.ts#L896-L905
> 
> **Also `src/utils/jellyfin.ts` is dead code.** (we dont use that file anymore). `JellyAPI` has no importers anywhere in the tree, and the `isEmby` param you added has no call site so it would always default to `false`. **That file should be deleted in a separate PR, not patched here.**
> 
> **So the fix is one line in `auth.ts`**: switch that header to `Authorization`. No mediaServerType branching, since Emby accepts the `MediaBrowser` scheme on `Authorization` too.

I updated src/utils/jellyfin knowing it was dead-code in case it was ever used again but I won't touch it and minimize the change. I added the branching in case and probably should have marked as a draft until I checked Emby. I will modify and resubmit.

fallenbagel:
> I updated src/utils/jellyfin knowing it was dead-code in case it was ever used again but I won't touch it and minimize the change.

That file hasn't been used since the release of jellyseerr. I just forgot to remove it and it has just been sitting there since. Forgotten 😅. There is absolutely no need for it. We should remove it in a follow up pr tbh.

> I added the branching in case and probably should have marked as a draft until I checked Emby. I will modify and resubmit.

We have several emby users using the modern header since my pr last year that adds the modern authorization support to jellyfin (although i did miss the delete endpoint). It has always been supported by emby. So there's no point in changing it back.

aussierk:
> > I updated src/utils/jellyfin knowing it was dead-code in case it was ever used again but I won't touch it and minimize the change.
> 
> That file hasn't been used since the release of jellyseerr. I just forgot to remove it and it has just been sitting there since. Forgotten 😅. There is absolutely no need for it. We should remove it in a follow up pr tbh.
> 
> > I added the branching in case and probably should have marked as a draft until I checked Emby. I will modify and resubmit.
> 
> We have several emby users using the modern header since my pr last year that adds the modern authorization support to jellyfin (although i did miss the delete endpoint). It has always been supported by emby. So there's no point in changing it back.

Makes complete sense. I've updated the PR to just contain the one fix. Was doing the Avatar proxy fix and saw the hardcodes and grepped to see a few more instances which made me jump .

I can put together a dead code refactor PR soon.

fallenbagel (approved):
Lgtm!

0xSysR3ll (approved):

PR #3504, opened 2026-09-12
fix(auth):  refresh avatar on Quick Connect login

aussierk:
## Description

 Previously, Quick Connect sign-in never called `checkAvatarChanged`, unlike the
regular Jellyfin login route, so `avatarVersion/avatarETag` were never
populated for accounts that only ever signed in via Quick Connect. Therefore, `avatarproxy` then permanently served the Gravatar fallback instead of the real Jellyfin avatar.

This PR adds support for the `checkAvatarChanged` call in the Quick Connect sign in route by mirroring the existing call in the regular sign in route. 

## How Has This Been Tested?

 I logged into seer using Quick Connect and ensured the user image loaded. I then changed the user's image on Jellyfin, logged out of Seer and used Quick Connect again and the avatar updated. 

## Checklist:

## Summary by CodeRabbit

- **Bug Fixes**
  - Quick Connect authentication now refreshes a user’s avatar when it has changed on the Jellyfin server.
  - Avatar updates are saved automatically after successful login.
  - Avatar synchronization errors no longer interrupt the login process.

### DIFF
--- a/server/routes/auth.test.ts
+++ b/server/routes/auth.test.ts
@@ -1,4 +1,5 @@
 import assert from 'node:assert/strict';
+import { createHash } from 'node:crypto';
 import { before, beforeEach, describe, it, mock } from 'node:test';
 
 import JellyfinAPI from '@server/api/jellyfin';
@@ -8,10 +9,12 @@ import { UserType } from '@server/constants/user';
 import { getRepository } from '@server/datasource';
 import { User } from '@server/entity/User';
 import PreparedEmail from '@server/lib/email';
+import ImageProxy from '@server/lib/imageproxy';
 import { getSettings } from '@server/lib/settings';
 import { checkUser } from '@server/middleware/auth';
 import { setupTestDb } from '@server/test/db';
 import { ApiError } from '@server/types/error';
+import axios from 'axios';
 import type { Express } from 'express';
 import express from 'express';
 import session from 'express-session';
@@ -273,11 +276,44 @@ describe('GET /auth/jellyfin/quickconnect/check', () => {
 });
 
 describe('POST /auth/jellyfin/quickconnect/authenticate', () => {
+  const fakeAvatarBuffer = Buffer.from('fake-quickconnect-avatar-bytes');
+
+  const axiosHeadMock = mock.method(axios, 'head', async () => ({
+    status: 200,
+    headers: { 'last-modified': 'Wed, 01 Jan 2025 00:00:00 GMT' },
+  }));
+
+  const clearCachedImageMock = mock.method(
+    ImageProxy.prototype,
+    'clearCachedImage',
+    async () => undefined
+  );
+
+  const getImageMock = mock.method(
+    ImageProxy.prototype,
+    'getImage',
+    async () => ({
+      imageBuffer: fakeAvatarBuffer,
+      meta: {
+        revalidateAfter: 3600,
+        curRevalidate: 3600,
+        isStale: false,
+        etag: 'mock-meta-etag',
+        extension: 'jpg',
+        cacheKey: 'mock-cache-key',
+        cacheMiss: true,
+      },
+    })
+  );
+
   beforeEach(() => {
     authenticateQCMock.mock.resetCalls();
     authenticateQCMock.mock.mockImplementation(async () => ({
       ...defaultAuthenticateResponse,
     }));
+    axiosHeadMock.mock.resetCalls();
+    clearCachedImageMock.mock.resetCalls();
+    getImageMock.mock.resetCalls();
     configureJellyfin();
   });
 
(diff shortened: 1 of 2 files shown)


fallenbagel (approved):
~~LGTM~~ 
Edit: just a few small nitpicks

0xSysR3ll (dismissed):

server/routes/auth.ts
```diff
@@ -799,6 +799,26 @@
         await userRepository.save(user);
       }
 
+      if (user && user.jellyfinUserId) {
```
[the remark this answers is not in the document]
fallenbagel: @aussierk oh yeah drop the `user` here. It can't be null here. Every branch above either assigns it or returns, which is why the code below reaches `user.id` without a null check. Drop the `user &&`.

server/routes/auth.test.ts
```diff
@@ -273,11 +276,44 @@ describe('GET /auth/jellyfin/quickconnect/check', () => {
...
+        revalidateAfter: 3600,
+        curRevalidate: 3600,
+        isStale: false,
+        etag: 'mock-meta-etag',
+        extension: 'jpg',
+        cacheKey: 'mock-cache-key',
+        cacheMiss: true,
+      },
+    })
+  );
```
fallenbagel: These aren't scoped to the describe. They patch the prototypes as soon as the file loads and stay patched for every test below them, including the local login and reset-password blocks. Move them up with the other module level mocks so that's visible

server/routes/auth.test.ts
```diff
@@ -381,6 +417,48 @@ describe('POST /auth/jellyfin/quickconnect/authenticate', () => {
...
+      where: { jellyfinUserId: 'jf-qc-user-001' },
+    });
+    assert.strictEqual(updatedUser.avatarVersion, expectedRemoteVersion);
+    assert.strictEqual(updatedUser.avatarETag, expectedEtag);
+    assert.strictEqual(
+      updatedUser.avatar,
+      `/avatarproxy/jf-qc-user-001?v=${expectedRemoteVersion}`
+    );
+    assert.strictEqual(getImageMock.mock.callCount(), 1);
+    assert.strictEqual(clearCachedImageMock.mock.callCount(), 1);
```
fallenbagel: Drop these two too. `expectedEtag` is a sha256 of the fake buffer, so the etag assertion above already can't pass unless `getImage` ran, and the `clearCachedImage` count ties this test to how `checkAvatarChanged` invalidates its cache today. If that invalidation is worth covering at all, it would still belong in an avatarproxy test and not here

fallenbagel (changes requested):

aussierk:
Ill make the changes in the next hour or so.

fallenbagel:
> Ill make the changes in the next hour or so. 

No rush, take your time 😄

aussierk:
> > Ill make the changes in the next hour or so.
> 
> No rush, take your time 😄

Oh I know, but getting it addressed and merged means I got 1 less PR to worry about keeping up to date 😆

fallenbagel:
> > > Ill make the changes in the next hour or so.
> > 
> > No rush, take your time 😄
> 
> Oh I know, but getting it addressed and merged means I got 1 less PR to worry about keeping up to date 😆 

Tbh no need to keep them up to date unless there is a conflict (and even then you can take your time and resolve the conflicts when your pr starts getting actively reviewed) since it can still be merged even when not up-to-date.

server/routes/auth.test.ts
```diff
@@ -381,6 +420,46 @@ describe('POST /auth/jellyfin/quickconnect/authenticate', () => {
...
+
+    const updatedUser = await userRepo.findOneOrFail({
+      where: { jellyfinUserId: 'jf-qc-user-001' },
+    });
+    assert.strictEqual(updatedUser.avatarVersion, expectedRemoteVersion);
+    assert.strictEqual(updatedUser.avatarETag, expectedEtag);
+    assert.strictEqual(
+      updatedUser.avatar,
+      `/avatarproxy/jf-qc-user-001?v=${expectedRemoteVersion}`
+    );
```
fallenbagel: These assertions re-derive `checkAvatarChanged` and `getUserAvatarUrl` inside an auth test. The regression test we are doing here is that Quick Connect never refreshed the avatar, and asserting the values moved off `old-version`, `old-etag` and the stale URL catches that on its own. As written, changing the version scheme in avatarproxy breaks an auth test for reasons that have nothing to do with auth. Drop the recomputation and assert they changed.

aussierk:
> > > > Ill make the changes in the next hour or so.
> > > 
> > > 
> > > No rush, take your time 😄
> > 
> > 
> > Oh I know, but getting it addressed and merged means I got 1 less PR to worry about keeping up to date 😆
> 
> Tbh no need to keep them up to date unless there is a conflict (and even then you can take your time and resolve the conflicts when your pr starts getting actively reviewed) since it can still be merged even when not up-to-date.

Fair enough, I am will keep that in mind. Unless some quick easy bugs come up, my next two PRs will probably be a dead code clean and continue work on my draft multiple requests for the same media PR.

server/routes/auth.test.ts
```diff
@@ -68,6 +71,39 @@ const authenticateQCMock = mock.method(
   async () => ({ ...defaultAuthenticateResponse })
 );
 
+// Avatar-refresh mocks (checkAvatarChanged internals) — module level since
+// mock.method patches axios/ImageProxy.prototype globally regardless of
+// which describe block declares it.
```
fallenbagel: While making changes in this file, `mock.method` patching the prototype globally is just how the API works, so this doesn't really need three lines explaining. Cut it to one on the placement, or drop it

fallenbagel (changes requested):
Just teeny tiny two nitpicks

server/routes/auth.test.ts
aussierk: Yeah, I realized I didn't remove my comment on this part on the first push but cleaned it up already.

fallenbagel (approved):

0xSysR3ll (approved):
LGTM

PR #3511, opened 2026-09-12
test: block outbound HTTP in unit tests

fallenbagel:
## Description

Several unit test suites were silently calling the real TMDB API. TheMovieDb assigns most of its methods as class properties in the constructor rather than defining them on the prototype, so the suites that stub them use a getter/setter pair on `TheMovieDb.prototype`, where an empty setter swallows the constructor’s assignment and reads fall through to the getter. 

That only works for instances built after the stub is installed, and both availabilitySync and the scanner singletons build theirs at import time, so their real methods shadowed the stub permanently. The affected suites worked around this by importing the singleton further down the file, below the stubs, which holds only as long as nothing else pulls the module in first. `server/api/jellyfin.ts` imports availabilitySync, so that suite’s deferred import was already a cache hit by the time it ran and its stubs never bound at all. The ones in the scanner suites survived on the accident of their import graphs. The result was tests that made real network calls, passed anyway, and quietly asserted against live data. One availability test was only green because TMDB really does report more seasons than its own fixture declared, and three specials fixtures had never been applied to anything. 

This PR stubs the singleton’s own instance, which does not depend on import order at all, and removes the deferred imports entirely. To stop the class of problem coming back, the test setup now wraps request and get on `node:http` and `node:https` and refuses anything that is not loopback. Loopback stays open because supertest serves the app over a real socket. Blocking alone would not be enough, since the two leaks this found were both in code that catches and logs its errors, so the guard also records every blocked host and fails the suite from the existing after hook. It can be turned off with `ALLOW_NETWORK=true` if a test ever genuinely needs the network. 

## How Has This Been Tested?

The leaks were found by pointing axios at a dead proxy so that every outbound attempt had to surface, rather than succeeding quietly: 
```
env HTTPS_PROXY=http://127.0.0.1:1 HTTP_PROXY=http://127.0.0.1:1 \
NO_PROXY=localhost,127.0.0.1,::1 VERBOSE=true pnpm test
```

 That run recorded 17 real outbound requests, all external, and one failing test (should mark show as PARTIALLY_AVAILABLE when some seasons are available and some are unknown), which returned AVAILABLE once it could no longer reach TMDB. The same command now reports no outbound attempts, and that test passes on the four-season fixture its own `beforeEach` was already building. 

pnpm test passes. Also to confirm the guard actually engages, I tested with a throwaway test file calling new `TheMovieDb().getMovie({ movieId: 550 })` was run against it and failed on both paths, the request itself threw, and the after hook reported that test reached the network: https://api.themoviedb.org. 

## Checklist:

## Summary by CodeRabbit

* **Tests**
  * Improved test setup and mocking for TV show scanning, Jellyfin, Sonarr, watchlists, and issue notifications.
  * Added safeguards to prevent unintended external network requests during tests, while allowing explicitly enabled network access.
  * Tests now detect and report any unexpected outbound requests.

* **Chores**
  * Organized test imports for improved consistency.

### DIFF
--- a/server/lib/availabilitySync.test.ts
+++ b/server/lib/availabilitySync.test.ts
@@ -28,6 +28,7 @@ import Media from '@server/entity/Media';
 import MediaRequest from '@server/entity/MediaRequest';
 import Season from '@server/entity/Season';
 import { User } from '@server/entity/User';
+import availabilitySync from '@server/lib/availabilitySync';
 import type { RadarrSettings, SonarrSettings } from '@server/lib/settings';
 import { getSettings } from '@server/lib/settings';
 import { setupTestDb } from '@server/test/db';
@@ -170,6 +171,14 @@ Object.defineProperty(TheMovieDb.prototype, 'getShowByTvdbIdForScan', {
   configurable: true,
 });
 
+// getTvShowForScan is assigned in the constructor, so the prototype stub misses
+// the instance availabilitySync built when it was first imported
+Object.defineProperty(availabilitySync.tmdb, 'getTvShowForScan', {
+  value: async (args: { tvId: number; language?: string }) =>
+    getTvShowImpl(args),
+  configurable: true,
+});
+
 // --- Helpers ---
 
 function fakeTmdbShow(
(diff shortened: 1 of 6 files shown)


gauthier-th (approved):
LGTM

0xSysR3ll (approved):

Issue #3298, opened 2026-07-27
Outdated Openapi Yaml file ?


Billos:
### Description

Hello, 
I have been using the yaml file to generate a quick [Typescript SDK](https://www.npmjs.com/package/@billos/seerr-sdk).

I'm currently trying to use the POST `/api/v1/watchlist`, with the following parameters => `{ tmdbId: 1422 }`
I can use the following fields : 
```typescript
export type WatchlistWritable = {
    tmdbId?: number;
    ratingKey?: string;
    type?: string;
    title?: string;
    media?: MediaInfoWritable;
    requestedBy?: UserWritable;
};
```
This matches what the yaml file describe.

But the response I'm getting from the Seerr backend is : 
```
ℹ Error: [
  {
    "code": "invalid_value",
    "values": [
      "movie",
      "tv"
    ],
    "path": [
      "mediaType"
    ],
    "message": "Invalid option: expected one of \"movie\"|\"tv\""
  }
]
```

I think the Yaml file is simply outdated compared to the real APIs.

### Version

3.3.0

### Steps to Reproduce

Simple curl should do the trick

```bash
curl -L 'https://seerr.domain.net/api/v1/watchlist' -H 'Content-Type: application/json' -H 'Accept: application/json' -H 'X-Api-Key: xxxxxxxxxxxxx' -d '{ "tmdbId": 1422 }'
```

### Logs

```shell

```

### Platform

desktop

### Database

PostgreSQL

### Device

Desktop

### Operating System

Kubuntu 26.04

### Browser

Firefox

### Search Existing Issues

- [x] Yes, I have searched existing issues.

### Code of Conduct

- [x] I agree to follow Seerr's Code of Conduct

M0NsTeRRR:
Hello,
Yes some parts are outdated

gauthier-th:
Related to #2700

Gauvino:
Some context for where this comes from: I maintain Streamyfin, a Jellyfin client with a Seerr integration. It gets its types from a fork of the server pulled in as a git submodule, because that is the only source that matches what the API actually sends. I'm trying to move it onto the published spec, and that is how I ended up measuring all this.

The yaml drifted because nothing has been checking it since 2020.

[...]

PR #3522, opened 2026-09-18
fix(api): restore OpenAPI response validation behind a flag

Gauvino:
## Description

Streamyfin is a Jellyfin client with a Seerr integration. It takes its types from a fork of the server pulled in as a submodule, because the spec does not match the responses. Moving it onto the published spec is what turned this up.

`validateResponses: true` was dropped in `96f3861c`, the bump to express-openapi-validator v4.8.0, when the call moved from `new OpenApiValidator().install()` to `OpenApiValidator.middleware()`. Nothing has checked a response against `seerr-api.yml` since, which is how the spec got to where #3298 found it. The workaround comment for it is still under the validator, guarding nothing.

Switching it straight back on would be rough after six years of drift, so it only runs with `VALIDATE_API_RESPONSES=true`, and it logs through `onError` instead of throwing. The response is still served, with one exception that comes from the validator's own response wrapper: a `null` JSON body becomes a 204. Nothing in `server/routes` sends one today. The Cypress job sets it, so the drift shows up in CI without failing anything.

It won't catch a property that is served and not declared, since none of our schemas set `additionalProperties: false`. It catches missing required properties, wrong types and nullability.

Related to #3298.

**AI Disclosure**: AI-assisted throughout. I used Claude Code (Opus 5, max effort) to trace the regression through the git history, to measure a live server against the spec, to write the patch, and to run the build and the test suite. I reviewed all of it, and this description started as its draft and was edited by me.

## How Has This Been Tested?

Built in a `node:22` container, seeded with `pnpm cypress:prepare`, run with the flag. Signing in as the seeded admin catches one straight away:

```
[warn][OpenAPI]: Response does not match the API specification
{"method":"POST","path":"/api/v1/auth/local","errors":[{"path":"/response/email",
"message":"must have required property 'email'","errorCode":"required.openapi.validation"}]}
```

`POST /auth/local` sends the user through `filter()`, which strips `email`, against a schema that declares it required. Still answered 200.

Without the flag the same requests log nothing, so the default path is unchanged.

`pnpm test` is 189 passing before and after. `pnpm build` passes.

## Checklist:

- [X] Disclosed any use of AI (see our [policy](https://github.com/seerr-team/seerr/blob/develop/CONTRIBUTING.md#ai-assistance-notice))

## Summary by CodeRabbit

* **New Features**
  * Added optional API response validation against the published API specification.
  * Validation mismatches are logged with the request method and path while responses continue to be served.

* **Documentation**
  * Added contributor guidance for enabling response validation during development and testing.

* **Tests**
  * Cypress runs now enable response validation to report specification mismatches without failing the test run.


fallenbagel (commented):
I have not reviewed this but from a quick look at the PR, there are a few details that violate our contributing guidelines, and we haven't made exceptions for it in the past either, so just wanted to flag it. 

While we do appreciate the detail in the AI disclosure, since it's more specific than most we get here, read against our policy it describes Claude Code doing the actual implementation work, which is a violation of our contributing guidelines' AI policy. Claude Code traced the regression through git history, measured a live server against the spec, wrote the patch, and ran the build and tests, with your review coming afterward. 

Our [contributing guideline's AI assistance notice](https://github.com/seerr-team/seerr/blob/794743a45f17e3d6aba06d68e1716e8b15146673/CONTRIBUTING.md#ai-assistance-notice) asks for AI-assisted development, where AI is used as a tool to assist with development that you are actually doing, not AI-driven development where the model is doing the implementation and you are reviewing the result. Reviewing or testing an AI-generated implementation afterward doesn't move a contribution from one category to the other. How clean the resulting diff reads doesn't change that either.

The same notice also expects PR descriptions to be your own words. The disclosure says this description started as Claude's draft and was edited by you, which is the same issue. An edited AI draft isn't a description written by you, no matter how much editing happened. Writing the PR description yourself takes a few minutes, so please just write it yourself next time.

To keep this open, I'd like to understand your reasoning around a couple of things the PR text doesn't already spell out. If an endpoint started returning a genuinely empty body on purpose, would the 204 rewrite mask that being wrong, or would it still catch it? And why log at warn rather than error, given the response is served either way?

If the intention is to have Claude do the implementation and then have you review what it produced, though, that's not something we can accept as a contribution here. At that point the implementation is being outsourced to the model rather than being your work with AI assisting you. If that's the workflow you want to use, that's fine, but those contributions will be closed without review under our existing guidelines for this repository.

PR #3531, opened 2026-09-20
fix(requests): tolerate orphaned request owners

JavierQuinan:
## Description

I ran into a case where an older request can still exist even though its original requester is no longer available.

When Seerr checks existing auto-requests, it accesses `requestedBy.id` directly. If the requester relation is missing, that causes request creation to fail with:

`TypeError: Cannot read properties of null (reading 'id')`

The fix is intentionally small. The check now uses `r.requestedBy?.id`, so an orphaned historical request does not crash the request flow.

I also added a regression test that recreates this situation and verifies that a new request can still be created normally.

**AI Disclosure:** I used ChatGPT to assist with reviewing the repository state, contribution workflow, and validation process. I reviewed the final implementation myself, reproduced the issue, ran the tests, and understand the submitted change.

## How Has This Been Tested?

I first reproduced the problem with the original code.

With the direct `r.requestedBy.id` access, the regression test fails as expected with:

`TypeError: Cannot read properties of null (reading 'id')`

After changing it to `r.requestedBy?.id`, the same test passes.

Validation completed:

- Regression test: PASS, 1/1
- Related test suites: PASS, 12 suites / 36 tests
- Typecheck: PASS
- Lint: PASS, with 19 pre-existing warnings in unrelated files
- Format check: PASS
- Full `pnpm build`: PASS
- `git diff --check`: PASS

The build still reports the existing Next.js warning about Automatic Static Optimization caused by `_app.getInitialProps`. This is unrelated to this change.

Not applicable.

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where creating a media request could fail when an existing automatic request had missing requester information.
  * New requests are now created correctly in this situation.


fallenbagel:
**A missing `requestedBy` here is leftover corruption from old TypeORM/media availability sync job issues in earlier overseerr/jellyseerr versions, and not something current code can produce, and it shouldn't be silently tolerated where it's found.**

That relation is `onDelete: 'CASCADE'`, not nullable, foreign key enforcement is switched back on right after migrations run in production, and the user-delete route explicitly removes all of a user's requests before deleting the user. None of that leaves room for an orphaned request under current code. The rows that do exist like this are historical artifacts from a TypeORM bug/media availability sync job in older versions that could drop the `requestedById` reference, and the regression test here proves as much as well, as in order to reproduce this, it required running `PRAGMA foreign_keys = OFF`, hand-editing the row with a raw `UPDATE`, then turning foreign keys back on. That's manually reproducing old corruption, something the codebase itself cannot otherwise produce otherwise.

Understanding the actual root cause would therefore have led somewhere else entirely, given that, `?.` here is the wrong place to handle it. It just makes the duplicate-request check silently skip the corrupted row on every run, forever, with nothing logged and nothing cleaned up. Anyone still carrying one of these old rows keeps carrying it, invisibly, instead of it ever getting fixed. If this needs handling at all, it should be addressed as a one-time data cleanup, or at minimum cleaned up directly via SQL ran on the database itself, rather than hiding the corruption behind optional chaining.

Separately, your disclosure says:
> I used ChatGPT to assist with reviewing the repository state, contribution workflow, and validation process. I reviewed the final implementation myself, reproduced the issue, ran the tests, and understand the submitted change.

Just for next time, per our [Contributing Guide](https://github.com/seerr-team/seerr/blob/59d5947b4df8591882bda70ae199f3a708e2d02b/CONTRIBUTING.md#expectations-for-ai-assisted-contributions), reviewing and testing an AI-produced implementation afterward doesn't turn AI-driven development into AI-assisted development. The guidelines are explicit that later scrutiny doesn't change that classification and what matters is how the implementation itself was produced.

That said, I'm not treating this as an AI-driven violation here because the fix is wrong on its own merits regardless of how it was produced. Please read through the guidelines before opening a PR next time, though.

Closing this since patching around the symptom isn't the right fix. Thank you for the contribution, though! **For anyone who has these orphaned requests, I'd recommend cleaning them up directly in the database. If you need help with that, please create a support ticket and I'll be happy to guide you through it.**

